PR #23, opened 2025-01-07
chore: Pull upstream changes

DaughterOfMars:
# Description

This PR rebases changes from `develop` on top of a renamed/refactored version of upstream changes.

PR #26, opened 2025-02-17
chore: Pull Upstream Changes but keep `schemars`

DaughterOfMars:
## Description

Pulls upstream changes as in #23 but excludes the removal of schemars implementations.


Thoralf-M:
CI fails

DaughterOfMars:
> CI fails

I don't think it's ever succeeded?

Thoralf-M:
> I don't think it's ever succeeded?

It did at some point, but now it doesn't even compile, at least this should work
```
error: this file contains an unclosed delimiter
   --> crates/iota-sdk-types/src/hash.rs:379:3
    |
294 | impl crate::ObjectId {
    |                      - unclosed delimiter
...
301 |     pub fn derive_id(digest: crate::TransactionDigest, count: u64) -> Self {
    |                                                                            - this delimiter might not be properly closed...
...
336 | }
    | - ...as it matches this but it has different indentation
...
379 | }
    |  ^
    ```

PR #29, opened 2025-04-07
feat: include gas price feedback in the execution error for transactions cancelled due to congestion

roman1e2f5p8s:
This PR extends `ExecutionCancelledDueToSharedObjectCongestion` to include gas price feedback in the execution error for transactions cancelled due to shared object congestion. It is needed for https://github.com/iotaledger/iota/issues/6206.

> [!IMPORTANT]
> This PR should only be merged after https://github.com/iotaledger/iota/pull/6280 is accepted.

### DIFF
--- a/crates/iota-rust-sdk/src/types/execution_status.rs
+++ b/crates/iota-rust-sdk/src/types/execution_status.rs
@@ -158,6 +158,13 @@ pub enum ExecutionError {
     /// Certificate is cancelled due to congestion on shared objects
     ExecutionCancelledDueToSharedObjectCongestion { congested_objects: Vec<ObjectId> },
 
+    /// Certificate is cancelled due to congestion on shared objects;
+    /// suggested gas price can be used to give this certificate more priority.
+    ExecutionCancelledDueToSharedObjectCongestionV2 {
+        congested_objects: Vec<ObjectId>,
+        suggested_gas_price: u64,
+    },
+
     /// Address is denied for this coin type
     AddressDeniedForCoin { address: Address, coin_type: String },
 
@@ -167,13 +174,6 @@ pub enum ExecutionError {
     /// Certificate is cancelled because randomness could not be generated this
     /// epoch
     ExecutionCancelledDueToRandomnessUnavailable,
-
-    /// Certificate is cancelled due to congestion on shared objects;
-    /// suggested gas price can be used to give this certificate more priority.
-    ExecutionCancelledDueToSharedObjectCongestionV2 {
-        congested_objects: Vec<ObjectId>,
-        suggested_gas_price: u64,
-    },
 }
 
 #[derive(Eq, PartialEq, Clone, Debug)]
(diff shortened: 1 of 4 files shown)


Thoralf-M:
@roman1e2f5p8s just wondering, should this PR target develop or do you want to create a feature branch, since you want to do more changes to actually return this error?

roman1e2f5p8s:
> @roman1e2f5p8s just wondering, should this PR target develop or do you want to create a feature branch, since you want to do more changes to actually return this error?

I am not sure which way is better. Do you mean creating a feature branch only for this PR in this repo, or also try merging https://github.com/iotaledger/iota/pull/6280 into some feature branch first there?

Thoralf-M:
> I am not sure which way is better. Do you mean creating a feature branch only for this PR in this repo, or also try merging https://github.com/iotaledger/iota/pull/6280 into some feature branch first there?

Depends on how many changes you want to do, is it only these two PRs? Then it should be fine like they are, if there is more then it might be worth to have a feature branch
At least for this PR here alone it looks very incomplete, just adding unused code that's why I wondered

roman1e2f5p8s:
> > I am not sure which way is better. Do you mean creating a feature branch only for this PR in this repo, or also try merging [iotaledger/iota#6280](https://github.com/iotaledger/iota/pull/6280) into some feature branch first there?
> 
> Depends on how many changes you want to do, is it only these two PRs? Then it should be fine like they are, if there is more then it might be worth to have a feature branch At least for this PR here alone it looks very incomplete, just adding unused code that's why I wondered

Yes, currently, it is only these two PRs, no additional changes at the moment. I think this particular PR should only be merged after https://github.com/iotaledger/iota/pull/6280 is accepted; otherwise this PR does not make sense.

crates/iota-rust-sdk/src/types/execution_status.rs
```diff
@@ -167,6 +167,13 @@ pub enum ExecutionError {
     /// Certificate is cancelled because randomness could not be generated this
     /// epoch
     ExecutionCancelledDueToRandomnessUnavailable,
+
+    /// Certificate is cancelled due to congestion on shared objects;
+    /// suggested gas price can be used to give this certificate more priority.
+    ExecutionCancelledDueToSharedObjectCongestionV2 {
+        congested_objects: Vec<ObjectId>,
+        suggested_gas_price: u64,
+    },
```
DaughterOfMars: Should we put it beneath the V1?

roman1e2f5p8s: Right after `ExecutionCancelledDueToSharedObjectCongestion` variant? Yes, I think that would make sense.

roman1e2f5p8s: Done

DaughterOfMars (approved):

Thoralf-M (approved):

thibault-martinez (approved):

DaughterOfMars (approved):

PR #42, opened 2025-06-19
feat(bindings): Create `iota-sdk-ffi` crate for bindings

DaughterOfMars:
## Description

This PR adds a new crate as compilation target for bindings. It defines a graphql and faucet client wrapper which have a set of uniFFI-friendly methods, and reverts some of the changes made to the existing crates. The intention is to have only feature-gated derives in these crates, and not to define specific functionality needed for the bindings.


crates/iota-graphql-client/src/query_types/coin.rs
```diff
@@ -34,16 +34,22 @@ use crate::query_types::{BigInt, schema};
 #[cynic(schema = "rpc", graphql_type = "CoinMetadata")]
 pub struct CoinMetadata {
     /// The number of decimal places used to represent the token.
+    #[cfg_attr(feature = "uniffi", uniffi(default = None))]
```
Thoralf-M: Is this really needed for each field individually?

Thoralf-M: If uniffi doesn't provide something, is there not a good way to write an own macro that adds it for Option<> fields?

DaughterOfMars: We would need a proc macro afaik

Thoralf-M: Had a look, but seems to be rather complicated, probably not worth it

crates/iota-sdk-types/src/lib.rs
```diff
@@ -201,6 +202,50 @@ macro_rules! impl_uniffi_byte_vec_wrapper {
...
+                None
+            }
+        }
+        )*}
+    };
+    ($($variant:ident),* $(,)?) => {
+        crate::def_is_as_opt!{$($variant => $variant,)*}
+    };
+}
+
```
Thoralf-M: Are these macros now needed only for uniffi? Should be in the uniffi_helpers.rs maybe?

DaughterOfMars: No, they are just general helper methods so I added them outside of the feature gate

Thoralf-M:
Clippy you can address in this PR, not sure about the other stuff

Thoralf-M (approved):

crates/iota-sdk-types/Cargo.toml
```diff
@@ -45,6 +45,7 @@ base64ct = { version = "1.6.0", features = ["alloc"] }
 bnum = "0.12.0"
 bs58 = "0.5.1"
 hex = "0.4.3"
+paste = "1.0"
```
thibault-martinez, line 46: Only for uniffi?

crates/iota-graphql-client/src/query_types/packages.rs
```diff
@@ -55,7 +55,6 @@ pub struct PackageByNameArgs<'a> {
 }
 
 #[derive(cynic::QueryFragment, Debug)]
-#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
```
thibault-martinez, line 38: How come?

crates/iota-graphql-client/Cargo.toml
```diff
@@ -45,4 +37,4 @@ iota-graphql-client-build = { version = "0.0.1", path = "../iota-graphql-client-
 
 [features]
 default = []
-uniffi = ["dep:uniffi", "iota-types/uniffi"]
```
thibault-martinez, line 48: Why? The crate doesn't compile independently anymore if you remove it

crates/iota-sdk-types/src/crypto/signature.rs
```diff
@@ -164,6 +164,7 @@ impl super::PasskeyPublicKey {
 }
 
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+#[cfg_attr(feature = "uniffi", derive(uniffi::Object), uniffi::export(Display))]
```
thibault-martinez: What's the logic of when to put `Display` or not?

crates/iota-sdk-types/Cargo.toml
thibault-martinez: Ah, I see this is a similar question as Thoralf

crates/iota-graphql-client/Cargo.toml
Thoralf-M: What do you mean? `cargo build` works fine in iota-graphql-client and iota-sdk-types

thibault-martinez: `cargo build` wouldn't use the uniffi feature so it works yes, `cargo build --all-features` does not

Thoralf-M: Ah, yeah `"iota-types/uniffi"` should probably stay

crates/iota-graphql-client/src/query_types/packages.rs
DaughterOfMars: Wasn't needed after all

crates/iota-sdk-types/src/crypto/signature.rs
DaughterOfMars: Errors got display, the rest did not

crates/iota-graphql-client/Cargo.toml
DaughterOfMars: fixed

Thoralf-M (approved):

Alex6323 (approved):

thibault-martinez (dismissed):
I would like to review this

crates/iota-sdk-ffi/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+path = "uniffi_bindgen.rs"
+
+[lib]
+name = "iota_sdk_ffi"
+crate-type = ["lib", "cdylib"]
+
+[dependencies]
+derive_more = { version = "2.0", features = ["from", "deref"] }
+rand = "0.8"
+serde_json = { version = "1.0.95" }
```
thibault-martinez: You're using both with "version" and without, can you harmonise please

crates/iota-graphql-client/src/query_types/active_validators.rs
```diff
@@ -107,7 +106,7 @@ pub struct Validator {
     /// The epoch at which this pool became active.
     pub staking_pool_activation_epoch: Option<u64>,
     /// The ID of this validator's `0x3::staking_pool::StakingPool`.
-    pub staking_pool_id: Address,
+    pub staking_pool_id: ObjectId,
```
thibault-martinez, line 109: Where does this change come from?

DaughterOfMars: It just makes more sense for the API since those aren't actually addresses

thibault-martinez: Ok but I mean, it has nothing to do with uniffi, right?

crates/iota-sdk-types/Makefile
```diff
@@ -15,10 +15,10 @@ test:
 	cargo nextest run --all-features
 	cargo test --all-features --doc
 
-# --all-features except `uniffi`
+# --all-features
 .PHONY: wasm
 wasm:
-	CC=clang wasm-pack test --node -F serde,schemars,rand,hash,proptest
+	RUSTFLAGS='--cfg getrandom_backend="wasm_js"' CC=clang wasm-pack test --node --all-features
```
thibault-martinez: Is this correct? You've removed it from another Makefile in another PR

crates/iota-sdk-ffi/README.md
```diff
@@ -0,0 +1,21 @@
...
+This crate can generate bindings for various languages (Go, Kotlin, Python, etc.) using [UniFFI](https://github.com/mozilla/uniffi-rs).
+
+Start by building the library to generate the appropriate dylib files.
+
+```sh
+cargo build -p iota-sdk-ffi --lib --release
+```
+
+Next, run the binary to generate the bindings for the desired language.
+
```
thibault-martinez, line 12: Please add a python section name so we can later add other languages clearly

crates/iota-sdk-ffi/README.md
```diff
@@ -0,0 +1,21 @@
+# IOTA SDK Bindings
```
thibault-martinez:
```suggestion
# IOTA SDK FFI
```

crates/iota-sdk-ffi/src/lib.rs
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Core type definitions for the IOTA blockchain.
```
thibault-martinez, line 4: Are docs exported to the bindings?

crates/iota-sdk-ffi/src/types/address.rs
```diff
@@ -0,0 +1,87 @@
...
+        Ok(Self(iota_types::Address::from_hex(hex)?))
+    }
+
+    #[uniffi::constructor]
+    pub fn generate() -> Self {
+        let mut rng = rand::thread_rng();
+        Self(iota_types::Address::generate(&mut rng))
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
```
thibault-martinez, line 80: So on regular methods (not constructor), we don't need any attributes for them to be exported?

crates/iota-sdk-ffi/src/types/gas.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use iota_types::GasCostSummary;
+
+#[uniffi::remote(Record)]
+pub struct GasCostSummary {
```
thibault-martinez, line 7: What's happening here? Why do we have two symbols with the same name? Why aren't we wrapping?

crates/iota-graphql-client/src/pagination.rs
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::query_types::PageInfo;
+
+#[derive(Clone, Debug)]
+/// A page of items returned by the GraphQL server.
```
thibault-martinez, line 6: Please put the doc before the attributes

crates/iota-graphql-client/src/query_types/dry_run.rs
```diff
@@ -38,19 +38,18 @@ pub struct TransactionMetadata {
     pub sender: Option<Address>,
 }
 
-#[derive(cynic::InputObject, Debug)]
+#[derive(Clone, cynic::InputObject, Debug)]
 #[cynic(schema = "rpc", graphql_type = "ObjectRef")]
 pub struct ObjectRef {
-    pub address: Address,
+    pub address: ObjectId,
```
thibault-martinez, line 44: Idk tbh, Address type for an address field seemed alright...

crates/iota-graphql-client/src/query_types/epoch.rs
```diff
@@ -129,5 +127,5 @@ pub struct ValidatorSet {
     /// Size of the validator candidates `Table`.
     pub validator_candidates_size: Option<i32>,
     /// Object ID of the `Table` storing the validator candidates.
-    pub validator_candidates_id: Option<Address>,
+    pub validator_candidates_id: Option<ObjectId>,
```
thibault-martinez, line 130: I still feel like address was better for things like this. @Thoralf-M @Alex6323 opinions?

crates/iota-sdk-ffi/README.md
Thoralf-M: I would just put the language specific things in their own folder, if I want to use the python binding I look in the python folder how to build it, not in the iota-sdk-ffi folder

crates/iota-graphql-client/src/query_types/object.rs
```diff
@@ -2,6 +2,8 @@
 // Modifications Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use iota_types::ObjectId;
```
thibault-martinez: Export it via `query_types` like you did for the other ones?

crates/iota-graphql-client/src/query_types/epoch.rs
Thoralf-M: For this case I prefer ObjectId as this table id shouldn't be used as address in the sense that you send something to it
But generally ofc `Address` is more generic, every ObjectId is also an Address, but not every Address is an ObjectId

crates/iota-graphql-client/src/query_types/active_validators.rs
DaughterOfMars: Yeah

crates/iota-sdk-types/Makefile
DaughterOfMars: Yeah, we do it in the workflow instead

DaughterOfMars: not sure how this got here

crates/iota-graphql-client/src/query_types/epoch.rs
thibault-martinez: Even if eventually it's the same thing, I still think we shouldn't change things like this because the schema says Address (meaning API docs will say Address) but the code will say ObjectId which is confusing.

crates/iota-sdk-ffi/README.md
DaughterOfMars: Let's do this in a later PR please

thibault-martinez: Then please open an issue for it

DaughterOfMars: https://github.com/iotaledger/iota-rust-sdk/issues/75

crates/iota-sdk-ffi/src/lib.rs
DaughterOfMars: Yes

DaughterOfMars: Actually, I don't know about the lib level docs, maybe not. For python they don't seem to be, but maybe other languages do?

crates/iota-sdk-ffi/src/types/address.rs
DaughterOfMars: Yes, only the `uniffi::export` at the impl block level

crates/iota-sdk-ffi/src/types/gas.rs
DaughterOfMars: From the user guide:

```
// Type aliases can be used to give remote types nicer names when exposed in the UniFFI api.
type LogLevel = log::Level;

// Write a definition that mirrors the definition from the remote crate and wrap it with `[uniffi::remote(<kind>)]`.
//
// - UniFFI will generate the FFI scaffolding code for the item, but will not output the item itself
//   (since the real item is defined in the remote crate).
// - `<kind>` can be any parameter that's valid for `uniffi::derive()`.
#[uniffi::remote(Enum)]
enum LogLevel {
    Error,
    Warn,
    Info,
    Debug,
    Trace,
}
```

crates/iota-graphql-client/src/query_types/dry_run.rs
DaughterOfMars: Yes, but it's not an address. The field in graphql is named incorrectly, but I don't want to touch the schema since it is supposed to match the core one.

crates/iota-graphql-client/src/query_types/object.rs
```diff
@@ -28,7 +28,7 @@ pub struct ObjectsQuery {
 
 #[derive(cynic::QueryVariables, Debug)]
 pub struct ObjectQueryArgs {
-    pub address: Address,
+    pub object_id: ObjectId,
```
thibault-martinez, line 31: Doesn't cynic care about field names?

crates/iota-sdk-types/src/lib.rs
```diff
@@ -176,30 +177,62 @@ pub use transaction::{
...
+        pub fn [< into_ $variant:snake >](self) -> $inner {
+            let Self::$variant(inner) = self else {
+                panic!("not a {}", stringify!($variant));
+            };
+            inner
+        }
+        )*}
+    };
+    ($($variant:ident),* $(,)?) => {
+        $crate::def_is_as_opt!{$($variant => $variant,)*}
```
thibault-martinez: What's happening here? If there is no given inner type it considers the variant name as inner type as well?

crates/iota-sdk-types/src/lib.rs
```diff
@@ -176,30 +177,62 @@ pub use transaction::{
...
+        pub fn [< as_ $variant:snake _opt >](&self) -> Option<&$inner> {
+            if let Self::$variant(inner) = self {
+                Some(inner)
+            } else {
+                None
+            }
+        }
+
+        #[inline]
+        pub fn [< into_ $variant:snake >](self) -> $inner {
```
thibault-martinez, line 231: Maybe this should be mentioned in the macro name

crates/iota-sdk-types/src/object.rs
```diff
@@ -323,14 +285,14 @@ pub struct UpgradeInfo {
...
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
-#[cfg_attr(feature = "uniffi", derive(uniffi::Record))]
 pub struct MoveStruct {
     /// The type of this object
     #[cfg_attr(
         feature = "serde",
         serde(with = "::serde_with::As::<serialization::BinaryMoveStructType>")
     )]
     pub type_: StructTag,
+
```
thibault-martinez: why?

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -207,6 +201,20 @@ pub enum TransactionKind {
...
 
+impl TransactionKind {
+    crate::def_is_as_opt! {
+        ProgrammableTransaction,
+        ConsensusCommitPrologueV1,
+        AuthenticatorStateUpdateV1,
+        RandomnessStateUpdate,
+    }
+
+    crate::def_is_as_opt! {
```
thibault-martinez: Just curious, couldn't it have been a single `def_is_as_opt`?

crates/iota-graphql-client/src/lib.rs
```diff
@@ -507,7 +416,7 @@ impl Client {
         &self,
         address: Address,
         coin_type: Option<String>,
-    ) -> Result<Option<u128>> {
+    ) -> Result<Option<u64>> {
```
thibault-martinez, line 419: How come?

crates/iota-graphql-client/src/query_types/object.rs
DaughterOfMars: so in this case they have the name correct in the schema 😆 

```
input ObjectKey {
	objectId: IotaAddress!
	version: UInt53!
}
```
I double checked the core repo and it's the same

crates/iota-graphql-client/src/lib.rs
DaughterOfMars: Well, we can't actually have larger values than u64, can we?

crates/iota-graphql-client/src/query_types/dry_run.rs
thibault-martinez: I don't think it's necessarily incorrect in this context

crates/iota-sdk-types/src/transaction/mod.rs
DaughterOfMars: No, because I wrote it such that it only maps one pattern. It's possible I could fix this but I didn't have the time and felt it wasn't worth the effort

crates/iota-sdk-types/src/object.rs
DaughterOfMars: looks like an accident

crates/iota-sdk-ffi/src/error.rs
```diff
@@ -0,0 +1,38 @@
...
+use std::fmt;
+
+use uniffi::Error;
+
+pub type Result<T, E = BindingsSdkError> = std::result::Result<T, E>;
+
+#[derive(Debug, Error)]
+#[uniffi(flat_error)]
+pub enum BindingsSdkError {
+    Generic(String),
```
thibault-martinez, line 13: Will we make no effort for having a more granular error? Is there a technical limitation due to uniffi?

crates/iota-graphql-client/src/query_types/dry_run.rs
DaughterOfMars: I mean this is the core type def:
```
pub type ObjectRef = (ObjectID, SequenceNumber, ObjectDigest);
```

crates/iota-sdk-ffi/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+//! ; --- ULEB128 definition ---
+//! uleb128         = 1*5uleb128-byte       ; Variable-length ULEB128 encoding
+//! uleb128-byte    = %x00-7F / %x80-FF     ; ULEB128 continuation rules
+//! uleb128-index   = uleb128               ; ULEB128-encoded variant index
+//! ```
+//!
+//! [BCS]: https://docs.rs/bcs
+//! [RFC-5234]: https://datatracker.ietf.org/doc/html/rfc5234
+
+#![allow(unused)]
```
thibault-martinez: expect so we don't forget to remove?

crates/iota-sdk-types/src/lib.rs
DaughterOfMars: Yes, basically if the inner type matches the variant name then there's no need to specify both

crates/iota-sdk-types/src/lib.rs
DaughterOfMars: I thought about it but does it really matter?

crates/iota-sdk-ffi/src/error.rs
DaughterOfMars: The error exists only because of the format that uniffi requires, i.e. it must be an enum. However, there's really no reason in my view to expand it

crates/iota-graphql-client/src/lib.rs
thibault-martinez: I don't know, just wondering if there was a reason it was 128 in the first place

crates/iota-sdk-types/src/lib.rs
thibault-martinez: Just thought it was weird and arbitrary that one method is not mentioned

crates/iota-graphql-client/src/lib.rs
DaughterOfMars: I also wondered the same

crates/iota-sdk-types/Makefile
```diff
@@ -15,10 +15,10 @@ test:
 	cargo nextest run --all-features
 	cargo test --all-features --doc
 
-# --all-features except `uniffi`
+# --all-features
 .PHONY: wasm
 wasm:
-	CC=clang wasm-pack test --node -F serde,schemars,rand,hash,proptest
+	CC=clang wasm-pack test --node --all-features
```
thibault-martinez:
```suggestion
	CC=clang wasm-pack test -r --node --all-features
```
Just in case it needs to be the same as crypto

crates/iota-sdk-types/src/type_tag/mod.rs
```diff
@@ -59,23 +58,83 @@ pub enum TypeTag {
     Struct(Box<StructTag>),
 }
 
-#[cfg(feature = "uniffi")]
-pub type BoxedTypeTag = Box<TypeTag>;
+impl TypeTag {
+    pub fn vector_type_opt(&self) -> Option<&TypeTag> {
```
thibault-martinez: I guess you did want to do a ref version of the macro?

crates/iota-sdk-types/src/lib.rs
DaughterOfMars: alright I changed it

crates/iota-sdk-types/src/lib.rs
```diff
@@ -196,7 +196,7 @@ macro_rules! def_is {
 }
 
 #[macro_export]
-macro_rules! def_is_as_opt {
+macro_rules! def_is_as_into_opt {
```
thibault-martinez, line 199: Sorry but it should be `def_is_as_opt_into`, opt separated from as doesn't make sense

crates/iota-sdk-types/src/type_tag/mod.rs
DaughterOfMars: It's just the box which is annoying here

DaughterOfMars: also the name conflicts in the case of fn struct :c

crates/iota-sdk-ffi/src/error.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use std::fmt;
+
+use uniffi::Error;
+
+pub type Result<T, E = BindingsSdkError> = std::result::Result<T, E>;
+
+#[derive(Debug, Error)]
+#[uniffi(flat_error)]
+pub enum BindingsSdkError {
```
thibault-martinez: Can you rename it to `SdkFfiError` please?

crates/iota-sdk-ffi/src/error.rs
thibault-martinez: Alright

crates/iota-sdk-ffi/src/graphql.rs
```diff
@@ -0,0 +1,861 @@
...
+    },
+    uniffi_helpers::{
+        CheckpointSummaryPage, CoinPage, DynamicFieldOutputPage, EpochPage, EventPage,
+        MovePackagePage, ObjectPage, SignedTransactionPage, TransactionDataEffectsPage,
+        TransactionEffectsPage, ValidatorPage,
+    },
+};
+
+#[derive(uniffi::Object)]
+pub struct GraphQLClient(RwLock<iota_graphql_client::Client>);
```
thibault-martinez, line 33: I think it should be `GraphQlClient` instead of `GraphQLClient`

crates/iota-sdk-ffi/src/types/graphql.rs
```diff
@@ -0,0 +1,319 @@
...
+#[derive(Clone, Debug, derive_more::From, uniffi::Object)]
+pub struct Event(pub iota_types::Event);
+
+#[derive(Clone, Debug, derive_more::From, uniffi::Object)]
+pub struct ObjectRef(pub iota_graphql_client::query_types::ObjectRef);
+
+#[derive(Clone, Debug, derive_more::From, uniffi::Object)]
+pub struct Epoch(pub iota_graphql_client::query_types::Epoch);
+
+#[derive(Clone, Debug, derive_more::From, uniffi::Record)]
```
thibault-martinez: What does this `derive_more::From` do?

crates/iota-sdk-types/src/type_tag/mod.rs
thibault-martinez: I see

crates/iota-sdk-types/src/lib.rs
DaughterOfMars: I added `into_opt` so now what

crates/iota-sdk-ffi/src/graphql.rs
DaughterOfMars: I disagree

crates/iota-sdk-types/src/lib.rs
thibault-martinez: Touché

crates/iota-sdk-ffi/src/types/graphql.rs
DaughterOfMars: accidentally left it after converting the wrapper impl

thibault-martinez: There are much more `derive_more::From` like this that don't make sense if you want to remove them as well...

DaughterOfMars: whoops, done

crates/iota-graphql-client/src/query_types/epoch.rs
Alex6323: I agree with the change tbh, although I see your point Thibault. It's also confusing as it was before, because the docs for this field clearly say it's expecting an Object ID.

thibault-martinez: For the record, I am against this change unless it is reflected in the schema (and I don't think infra would have an issue with that). This is introducing inconsistency between the API references and the code which is going to be confusing for ours users. I am not going to prevent merging this PR for this though.

crates/iota-sdk-ffi/src/graphql.rs
thibault-martinez: This is not very rusty but alright

crates/iota-graphql-client/src/query_types/epoch.rs
DaughterOfMars: I can create an issue on the core repo to update the schema

DaughterOfMars: https://github.com/iotaledger/iota/issues/8133

thibault-martinez: Thanks

thibault-martinez (approved):

Issue #61, opened 2025-07-30
ProtocolConfigs as record


Thoralf-M:

Thoralf-M:
The linked PR didn't update ProtocolConfigs https://github.com/iotaledger/iota-rust-sdk/blob/de39a56094e7bb1638782ccb3f1d8feb8827b56a/crates/iota-sdk-ffi/src/types/graphql.rs#L736-L737

Thoralf-M:
Fixed by https://github.com/iotaledger/iota-rust-sdk/pull/95

PR #63, opened 2025-07-30
feat(bindings): Convert some types to remote records and add docs

### LINKED ISSUE #58: PaginationFilter
accessors

DaughterOfMars:
## Description

Some types can still be Records, but they have to be redefined in the bindings crates. I added `From` impls to convert between them (with the added benefit of verifying them at compile time). 

In one case, `GasCostSummary`, there are only fields that are basic uniffi compatable types so it could be implemented with the remote attribute.

Overall this method provides a clean API for users in the case where we do not need associated methods, which is true for many of our structs.

Closes #58 #61 #62

### DIFF
--- a/crates/iota-sdk-ffi/src/types/digest.rs
+++ b/crates/iota-sdk-ffi/src/types/digest.rs
@@ -4,7 +4,8 @@
 use crate::error::Result;
 
 macro_rules! impl_digest_wrapper {
-    ($t:ident) => {
+    ($(#[$meta:meta])* $t:ident) => {
+        $(#[$meta])*
         #[derive(
             Copy,
             Clone,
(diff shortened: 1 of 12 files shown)


crates/iota-sdk-ffi/src/types/digest.rs
```diff
@@ -46,10 +47,32 @@ macro_rules! impl_digest_wrapper {
...
+    ///
+    /// ```text
+    /// digest = %x20 32OCTET
+    /// ```
+    ///
+    /// Due to historical reasons, even though a `Digest` has a fixed-length of 32,
+    /// IOTA's binary representation of a `Digest` is prefixed with its length
+    /// meaning its serialized binary form (in bcs) is 33 bytes long vs a more
+    /// compact 32 bytes.
+    Digest
```
Thoralf-M, line 74: The documentation works like this?

DaughterOfMars: I added the meta tag to the macro :)

crates/iota-sdk-ffi/src/types/graphql.rs
```diff
@@ -9,44 +9,45 @@ use crate::types::{
...
-        Self(inner)
-    }
-
-    pub fn gas_price(&self, gas_price: u64) -> Self {
-        let Self(mut inner) = self.clone();
-        inner.gas_price = Some(gas_price);
-        Self(inner)
-    }
+#[derive(Clone, Debug, derive_more::From, uniffi::Record)]
+pub struct TransactionMetadata {
```
Thoralf-M, line 13: Also want to put the docs here?

Thoralf-M (approved):

crates/iota-sdk-ffi/src/types/graphql.rs
DaughterOfMars: There are none on the core type actually

Alex6323 (approved):

PR #66, opened 2025-07-30
docs(bindings): Add Linux instructions to `iota-sdk-ffi::README.md`

Alex6323:


crates/iota-sdk-ffi/README.md
```diff
@@ -2,20 +2,42 @@
 
 This crate can generate bindings for various languages (Go, Kotlin, Python, etc.) using [UniFFI](https://github.com/mozilla/uniffi-rs).
 
-Start by building the library to generate the appropriate dylib files.
+## 1. Build the Rust FFI library
+
+Start by building the library to generate the appropriate `.dylib` (Mac) or `.so` (Linux) files.
```
Thoralf-M: For Windows one would need .dll, also instead of having one command for each OS, one could just run a script like this one:
```
#!/bin/bash
LANGUAGE="python"  # Change this to the desired language (e.g., go, kotlin, python)
# Start by building the library to generate the appropriate dylib files
cargo build -p iota-sdk-ffi --lib --release
# Determine the library extension based on the platform
case "$(uname -s)" in
  Darwin)   LIB_EXT=".dylib" ;;
  Linux)    LIB_EXT=".so" ;;
  MINGW*|MSYS*|CYGWIN*|Windows_NT) LIB_EXT=".dll" ;;
  *)        echo "Unsupported platform"; exit 1 ;;
esac
cargo run --bin iota_sdk_bindings -- generate --library "target/release/libiota_sdk_ffi${LIB_EXT}" --language $LANGUAGE --out-dir bindings/$LANGUAGE/lib --no-format
#Finally, copy the library file to the output directory.
cp target/release/libiota_sdk_ffi${LIB_EXT} bindings/$LANGUAGE/lib/
```

Alex6323: While it's possible to do that, here I would argue against a script tbh, because the target platform is always the same, so why let a script find out the platform everytime? Also you need to make it executable first as well. 

One could also do something like "target/relase/libiota_sdk_ffi.{dylib|so|dll} do reduce repetition, but then it's less copyable :shrug:

Thoralf-M: > Also you need to make it executable first as well.

No, I just execute this in my terminal
Adding an own version for each target makes it much more annoying to read, if you really want to have an own command for each, then I think it would be better to have an own section for each OS with all the steps for it without the commands for other targets in between
But then again, for each language you also have to change the commands

Alex6323: I don't think it's so annoying to read tbh. And having own separate sections, just because the file extension is different per target platform feels a bit too much, since with separate sections I would then also have to repeat the accompanying text.

Also ... once it's in the shell history it doesn't have to be copied again.

I am against the script, because that is actually hard to read with the comments in between, and also people shouldn't run random scripts they don't immediatedly understand. 


Personally I also wrote a script to make this all one command, but I wouldn't put this in the README.

Thoralf-M: Could have the individual steps and everything together, not having it just makes it harder for other people
But I guess I'm fine without it here, because the language specific thing should probably be only in the folder for this language

Thoralf-M (approved):

PR #74, opened 2025-08-06
feat(bindings): Add `UserSignature` and dependent type impls

### LINKED ISSUE #53: UserSignature
constructor + getter

DaughterOfMars:
## Description

Adds the `UserSignature` type impls and all types and impls that it depends on.

Closes #53


crates/iota-sdk-ffi/src/types/crypto/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+/// # BCS
+///
+/// The BCS serialized form for this type is defined by the following ABNF:
+///
+/// ```text
+/// validator-committee-member = bls-public-key
+///                              u64 ; stake
+/// ```
+#[derive(Clone, Debug, uniffi::Record)]
+pub struct ValidatorCommitteeMember {
```
thibault-martinez: This feels slightly out of place tbh

DaughterOfMars: How so?

DaughterOfMars: Oh, you mean the file it's in?

bindings/python/test.py
```diff
@@ -19,9 +19,16 @@ async def main():
...
 
-    filter=EventFilter(sender=my_address)
+    filter = EventFilter(sender=my_address)
+
+    txn = await client.transaction(TransactionDigest.from_base58("HT1wvebXV4LUisLa4aJQEyoxt1HEMqznjf5UtStmdxyM"))
+
+    if txn is not None:
+        for sig in txn.signatures():
+            print("Scheme: ", sig.scheme(), sig.scheme().value)
+        print(txn.transaction().sender().to_hex())
```
thibault-martinez:
```suggestion
        print("Sender: ", txn.transaction().sender().to_hex())
```

crates/iota-sdk-ffi/src/types/crypto/mod.rs
thibault-martinez: I mean sure it uses a crypto primitive but this type itself has nothing to do with crypto

crates/iota-sdk-ffi/src/types/crypto/validator.rs
thibault-martinez, line 1: I still think this does not belong here

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
```diff
@@ -0,0 +1,360 @@
...
+/// secp256k1-multisig-member-signature = %x01 secp256k1-signature
+/// secp256r1-multisig-member-signature = %x02 secp256r1-signature
+/// zklogin-multisig-member-signature   = %x03 zklogin-authenticator
+/// ```
+#[derive(Clone, Debug, derive_more::From, uniffi::Object)]
+pub struct MultisigMemberSignature(pub iota_types::MultisigMemberSignature);
+
+#[uniffi::export]
+impl MultisigMemberSignature {
+    pub fn is_ed25519(&self) -> bool {
```
thibault-martinez, line 36: Can we not use the macro from the main crate?

crates/iota-sdk-ffi/src/types/crypto/zklogin.rs
```diff
@@ -0,0 +1,334 @@
...
+/// ```
+///
+/// [`Address`]: crate::Address
+#[derive(Clone, Debug, derive_more::From, uniffi::Object)]
+pub struct ZkLoginPublicIdentifier(pub iota_types::ZkLoginPublicIdentifier);
+
+#[uniffi::export]
+impl ZkLoginPublicIdentifier {
+    #[uniffi::constructor]
+    pub fn new(iss: String, address_seed: &Bn254FieldElement) -> Result<Arc<Self>> {
```
thibault-martinez: Do we really need the Arc here?

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
thibault-martinez: Ah, uses self.0 I guess that's annoying

crates/iota-sdk-ffi/src/types/crypto/validator.rs
DaughterOfMars: That's how it is in the types crate

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
DaughterOfMars: yep

DaughterOfMars: also we can't use macros in general because of a limitation in the `#[uniffi::export]` macro

crates/iota-sdk-ffi/src/types/crypto/zklogin.rs
DaughterOfMars: Yep

crates/iota-sdk-ffi/src/types/crypto/validator.rs
thibault-martinez: Doesn't mean it's right and we can't change it

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
```diff
@@ -0,0 +1,360 @@
...
+        matches!(self.0, iota_types::MultisigMemberSignature::Ed25519(_))
```
thibault-martinez, line 37: I would prefer if we just call the inner functions instead of reimplementing them tbh

crates/iota-sdk-ffi/src/types/crypto/zklogin.rs
thibault-martinez: Why? How is that different from `ZkLoginInputs::new` that don't need it? Also, I removed it locally and it's compiling, what am I missing?

crates/iota-sdk-ffi/src/types/crypto/validator.rs
DaughterOfMars: yes, I'm just saying maybe it's outside the scope of this PR

crates/iota-sdk-ffi/src/types/crypto/zklogin.rs
DaughterOfMars: Oh, guess I'm wrong 😆

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
DaughterOfMars: What inner functions?

thibault-martinez: Just call `self.0.is_ed25519()`

crates/iota-sdk-ffi/src/types/crypto/validator.rs
thibault-martinez: I'm fine with that, please open an issue

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
thibault-martinez: I think in general we should just forward everything, it's a bit inconsistent atm, you do it for a lot of methods already

DaughterOfMars: That fn doesn't exist

crates/iota-sdk-ffi/src/types/crypto/validator.rs
DaughterOfMars: https://github.com/iotaledger/iota-rust-sdk/issues/84

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
thibault-martinez: Ah I see, very odd then. I think they should exist. Why are we providing more methods in the bindings than we have in the SDK?

DaughterOfMars: it's a transparent type in the SDK so it's not needed

thibault-martinez:
Conflicts

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
thibault-martinez: Sorry I don't get it, can you elaborate please? Why don't we want/need these convenience methods in the SDK?

DaughterOfMars: It's easy to match the type so you don't need to call all of this stuff 🤷‍♀️ but maybe we still provide it, I dunno

thibault-martinez: You introduced the is_as_opt etc macros, following your logic this wouldn't have been necessary then. I'd just do it everywhere if we end up having these methods in the bindings anyway.

DaughterOfMars: https://github.com/iotaledger/iota-rust-sdk/issues/86

crates/iota-sdk-ffi/src/types/signature.rs
```diff
@@ -1,5 +1,268 @@
...
+/// as `bytes` meaning it has a length prefix that defines the length of
+/// the completely serialized signature.
 #[derive(Clone, Debug, derive_more::From, uniffi::Object)]
 pub struct UserSignature(pub iota_types::UserSignature);
+
+#[uniffi::export]
+impl UserSignature {
+    /// Return the flag for this signature scheme
+    pub fn scheme(&self) -> SignatureScheme {
+        match &self.0 {
```
thibault-martinez: Just call `self.0.scheme()` ?

crates/iota-sdk-ffi/src/types/signature.rs
```diff
@@ -1,5 +1,268 @@
...
+
+    fn to_bytes(&self) -> Vec<u8> {
+        self.0.to_bytes()
+    }
+
+    pub fn is_ed25519(&self) -> bool {
+        matches!(self.0, iota_types::SimpleSignature::Ed25519 { .. })
+    }
+
+    pub fn ed25519_sig_opt(&self) -> Option<Arc<Ed25519Signature>> {
```
thibault-martinez, line 179: Name not consistent with the rest, missing the `as_` prefix. Same for following methods.

DaughterOfMars: Yes, but this is because this enum is different to others. It doesn't wrap single types we can return, so these are more like getters.

crates/iota-sdk-ffi/src/types/signature.rs
DaughterOfMars: done

thibault-martinez (approved):

crates/iota-sdk-ffi/src/types/signature.rs
thibault-martinez: Fair enough

Thoralf-M (approved):

PR #77, opened 2025-08-07
fix(graphql-client): Fix an issue that can cause the `test_total_transaction_blocks` to fail randomly

DaughterOfMars:
## Description

This test can now fail randomly if the queries are not fast enough and return different checkpoints. This fixes the problem by changing the assert.

### DIFF
--- a/crates/iota-graphql-client/src/lib.rs
+++ b/crates/iota-graphql-client/src/lib.rs
@@ -2319,9 +2319,9 @@ mod tests {
             .await
             .unwrap()
             .unwrap();
-        assert_eq!(
-            total_transaction_blocks,
-            total_transaction_blocks_by_seq_num
+        assert!(
+            total_transaction_blocks_by_seq_num > total_transaction_blocks,
+            "expected at least {total_transaction_blocks} transaction blocks, found {total_transaction_blocks_by_seq_num}"
         );
 
         let chckp = client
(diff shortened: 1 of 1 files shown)


crates/iota-graphql-client/src/lib.rs
```diff
@@ -2319,9 +2319,9 @@ mod tests {
             .await
             .unwrap()
             .unwrap();
-        assert_eq!(
-            total_transaction_blocks,
-            total_transaction_blocks_by_seq_num
+        assert!(
+            total_transaction_blocks_by_seq_num >= total_transaction_blocks,
+            "expected at least {total_transaction_blocks} transaction blocks, found {total_transaction_blocks_by_seq_num}"
```
Alex6323, line 2324: this is essentially a race condition, no? Wondering whether we should just change the assert to also just enforce `> 0` :thinking:

Alex6323: my reasoning is that the assert! should proof something about the internal logic, right? But here ... if you just change the order of calls in this test, the assert will sometimes fail and the test be flaky again.

DaughterOfMars: don't change the order :)

Thoralf-M (approved):

thibault-martinez (approved):

PR #91, opened 2025-08-12
feat(ffi): add MoveModule, MoveFunction and their inner types as records

### LINKED ISSUE #59: MoveModule MoveFunction
constructor + getter

Thoralf-M:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/59

Had to rename graphql MoveStruct to MoveStructQuery, because it was conflicting with the other one in iota-sdk-types:
`error: symbol `UNIFFI_META_IOTA_SDK_FFI_RECORD_MOVESTRUCT` is already defined`


thibault-martinez (approved):

DaughterOfMars (approved):

thibault-martinez (approved):

thibault-martinez:
Conflicts

crates/iota-sdk-ffi/src/types/graphql.rs
```diff
@@ -739,11 +744,209 @@ pub struct ProtocolConfigs(pub iota_graphql_client::query_types::ProtocolConfigs
 #[derive(Clone, Debug, derive_more::From, uniffi::Object)]
 pub struct CoinMetadata(pub iota_graphql_client::query_types::CoinMetadata);
 
-#[derive(Debug, derive_more::From, uniffi::Object)]
-pub struct MoveFunction(pub iota_graphql_client::query_types::MoveFunction);
+#[uniffi::remote(Record)]
+pub struct MoveFunction {
```
DaughterOfMars, line 748: Maybe we could just append `Query` to all of the queries for consistency? We should probably do this in the core crate too eventually.

DaughterOfMars (approved):

crates/iota-sdk-ffi/src/types/graphql.rs
Thoralf-M: Seems like there is no way to use `uniffi::remote(Record)` with a different name than in the core crate, so would require this change in the core crate
Don't want to do more changes than needed now, but I agree that this is something to consider

PR #92, opened 2025-08-12
feat(Makefile): add bindings make commands

Thoralf-M:
New commands in `make help` to make it easier to work with the bindings:
```
bindings                       Build all bindings
test-bindings                  Test all bindings
go                             Build Go bindings
test-go                        Test Go bindings
kotlin                         Build Kotlin bindings
test-kotlin                    Test Kotlin bindings
python                         Build Python bindings
test-python                    Test Python bindings
```

also updated the CI to make use of these commands and added *.so and *.dll files to the gitignore


DaughterOfMars (approved):
nice!

thibault-martinez:
Do you want to simplify the CI a tiny bit by using this?

Makefile
```diff
@@ -58,6 +58,78 @@ clean: ## Clean build artifacts
...
+	cd bindings/kotlin; \
+	./gradlew build clean; \
+	LD_LIBRARY_PATH=./lib ./gradlew run -q; \
+	cd -
+
+# Build Python bindings
+.PHONY: python
+python: ## Build Python bindings
+	cargo build -p iota-sdk-ffi --lib --release; \
+	case "$$(uname -s)" in \
```
thibault-martinez: Just wondering if this case can somehow be made into a function so we don't repeat it 3 times. Don't know if functions are a thing with Makefiles

Thoralf-M: Yes works https://github.com/iotaledger/iota-rust-sdk/pull/92/commits/d229f4c87cc464a9cef725ce5293cc42d418b4c0

Makefile
```diff
@@ -58,6 +58,65 @@ clean: ## Clean build artifacts
...
+cargo build -p iota-sdk-ffi --lib --release; \
+case "$$(uname -s)" in \
+	Darwin)   LIB_EXT=".dylib" ;; \
+	Linux)    LIB_EXT=".so" ;; \
+	MINGW*|MSYS*|CYGWIN*|Windows_NT) LIB_EXT=".dll" ;; \
+	*)        echo "Unsupported platform"; exit 1 ;; \
+esac;
+endef
+
+go: ## Build Go bindings
```
thibault-martinez, line 85: Do these not need a `.PHONY`?

Thoralf-M: Tbh I don't fully understand when to put it or not, I also wondered about it and tested, it works without it

thibault-martinez (approved):

Makefile
thibault-martinez: The rule of thumb seems to be that we want PHONY when the rule is not a physical entity like a file, so we probably want them here, I saw you did

Thoralf-M: Seems like one should have it if it's not actually a file, but not sure why, the example I found mentioned something related to `make clean` but for us this just calls `cargo clean` so it makes no difference
I added it anyways just to be on the safe side

Issue #49, opened 2025-07-30
Transaction types


Thoralf-M:
For each, create constructor and getter function for each field and

thibault-martinez:
Closed by https://github.com/iotaledger/iota-rust-sdk/pull/108

PR #98, opened 2025-08-13
feat(ffi): `ProgrammableTransaction`

thibault-martinez:
Part of https://github.com/iotaledger/iota-rust-sdk/issues/49

### DIFF
--- a/crates/iota-sdk-ffi/src/types/transaction/mod.rs
+++ b/crates/iota-sdk-ffi/src/types/transaction/mod.rs
@@ -9,9 +9,11 @@ use crate::types::{
     address::Address,
     digest::{CheckpointDigest, TransactionDigest, TransactionEventsDigest},
     execution_status::ExecutionStatus,
-    object::ObjectReference,
+    object::{ObjectId, ObjectReference},
     signature::UserSignature,
+    struct_tag::Identifier,
     transaction::v1::TransactionEffectsV1,
+    type_tag::TypeTag,
 };
 
 pub mod v1;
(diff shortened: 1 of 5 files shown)


crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -156,9 +158,457 @@ impl TransactionKind {
...
+/// ```text
+/// input = input-pure / input-immutable-or-owned / input-shared / input-receiving
+///
+/// input-pure                  = %x00 bytes
+/// input-immutable-or-owned    = %x01 object-ref
+/// input-shared                = %x02 object-id u64 bool
+/// input-receiving             = %x04 object-ref
+/// ```
+#[derive(Clone, Debug, derive_more::From, uniffi::Object)]
+pub struct Input(pub iota_types::Input);
```
DaughterOfMars, line 233: This could probably be an enum

DaughterOfMars: Same with Command. What do we think about the usability of that?

thibault-martinez: Can you have constructors on an enum?

DaughterOfMars: And more, I am seeing

crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -156,9 +158,457 @@ impl TransactionKind {
...
+    /// The amounts to split off
+    pub fn amounts(&self) -> Vec<Arc<Argument>> {
+        self.0
+            .amounts
+            .iter()
+            .cloned()
+            .map(Into::into)
+            .map(Arc::new)
+            .collect()
+    }
```
Thoralf-M, line 427: I'm wondering if a record wouldn't be better for these commands
At least with the current functions I don't see any reason to use an object

thibault-martinez: How is it any different than say `ProgrammableTransaction`? You need to construct them at some point in the bindings, either with fields if Record of with constructor if Object. Or are you saying they should all be Record? When do you start doing an Object?

DaughterOfMars (approved):

crates/iota-sdk-ffi/src/types/transaction/mod.rs
Thoralf-M: Object I would only do if I want methods on them that aren't just to set the fields for building or getters, because this is what we also have with records (just less code with records)
So object you want if you want a different method on it, like converting it to/from bcs for example

thibault-martinez: Would you also make `ProgrammableTransaction` a record then? It has the exact same setup, a constructor and two getters.

Thoralf-M: Just based on the current state, yes, but I feel like we might want to have specific methods on that in the future
For example in the core repo we have these methods on it https://github.com/iotaledger/iota/blob/4660e9505235ecd54e9a5e02fdaade071622cfaf/crates/iota-types/src/transaction.rs#L937-L1066

Thoralf-M (approved):

PR #110, opened 2025-08-18
chore(ffi): add `new_` prefix to some constructors

### LINKED ISSUE #88: Prefix all FFI constructors with `new_`
To avoid potential duplicate symbol duplicate between enum variants and constructors

thibault-martinez:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/88


crates/iota-sdk-ffi/src/types/struct_tag.rs
```diff
@@ -72,7 +72,7 @@ impl StructTag {
     }
 
     #[uniffi::constructor]
-    pub fn coin(type_tag: &TypeTag) -> Self {
+    pub fn new_coin(type_tag: &TypeTag) -> Self {
```
DaughterOfMars: These aren't needed as this isn't an enum

crates/iota-sdk-ffi/src/faucet.rs
```diff
@@ -27,19 +27,19 @@ impl FaucetClient {
 
     /// Set to local faucet.
     #[uniffi::constructor]
-    pub fn local() -> Self {
+    pub fn new_local() -> Self {
```
DaughterOfMars, line 30: also not needed

thibault-martinez: I would still keep as it's consistent with the `GraphQLClient`'s methods

crates/iota-sdk-ffi/src/types/struct_tag.rs
thibault-martinez: I debated that but I still prefer with. These would be the only "non-from" constructors without new_

crates/iota-sdk-ffi/src/faucet.rs
DaughterOfMars: fair enough

crates/iota-sdk-ffi/src/types/struct_tag.rs
DaughterOfMars: Well then we should remove `new_` from the others, because it doesn't add anything and isn't needed

thibault-martinez: It does add something, consistency. But ok I'll remove this one.

thibault-martinez: Plus it looks like a getter without it 🤷🏻

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/110/commits/ebea4e0285babffb3a7cc5a4dc430c6475a4cf40

DaughterOfMars (approved):

Thoralf-M (approved):

PR #113, opened 2025-08-18
feat(ffi): `uniffi::export(Hash)` types that are used in dictionaries

thibault-martinez:


Thoralf-M:
For what exactly is this needed? Do we maybe also need it for `Address` and all the types implemented with `impl_digest_wrapper!()`?

thibault-martinez:
> For what exactly is this needed? 

For any type that will be used as key to HashMaps. For some reasons it only make the Swift bindings to fail compile.

> Do we maybe also need it for `Address` and all the types implemented with `impl_digest_wrapper!()`?

I wouldn't add anything unless strictly required by bindings compilation tbh

Thoralf-M (approved):

DaughterOfMars (approved):

PR #147, opened 2025-08-28
chore(examples): Add `get_object` example

### LINKED ISSUE #146: Get an object with all possible data like bcs

DaughterOfMars:
## Description

Adds an example which gets an object by its ID and accesses the fields.

Closes #146


crates/iota-graphql-client/examples/get_object.rs
```diff
@@ -0,0 +1,47 @@
...
+    let registration = bcs::from_bytes::<IotaNamesRegistration>(bytes)?;
+
+    println!("registration name: {}", registration.name_str);
+
+    Ok(())
+}
+
+#[derive(Deserialize)]
+#[expect(unused)]
+struct IotaNamesRegistration {
```
thibault-martinez: It would make the example easier if we took a type already available in the SDK

DaughterOfMars: Which type would you recommend?

thibault-martinez: Maybe simply a Coin?

DaughterOfMars: Can't. The coin is not a type in the sdk that is deserializable

thibault-martinez: What prevents us from making it `Deserialize`?

thibault-martinez: Hmm, `TypeTag` I see

Thoralf-M: If the example is called get_object I would just get the existing data, you can print the json and the bcs values and that's it
Using the bcs and decode it into a struct I would only have in a separated example

bindings/go/examples/get_object.go
```diff
@@ -0,0 +1,42 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+
+	sdk "example.com/bindings/iota_sdk_ffi"
```
thibault-martinez:
```suggestion
	sdk "bindings/iota_sdk_ffi"
```

Makefile
```diff
@@ -114,8 +114,10 @@ go-example: ## Run a specific Go example. Usage: make go-example example
 	@true
 go-example:
 	cd bindings/go/examples; \
-	LD_LIBRARY_PATH="../../../target/release" CGO_LDFLAGS="-liota_sdk_ffi -L../../../target/release" go run $(word 2,$(MAKECMDGOALS)).go \
-	cd -
+	LD_LIBRARY_PATH="../../../target/release" CGO_LDFLAGS="-liota_sdk_ffi -L../../../target/release" go run $(word 2,$(MAKECMDGOALS)).go; \
+	code=$$?; \
+	cd -; \
+	exit $$code
```
thibault-martinez, line 120: Same for python and kotlin

thibault-martinez: Ah sorry, the diff was weird, kotlin is done

DaughterOfMars: python doesn't need it I think, and I did kotlin

bindings/kotlin/examples/GetObject.kt
thibault-martinez, line 1: Please add a copyright header, same for Example.kt as I see it's missing as well

crates/iota-graphql-client/examples/get_object.rs
thibault-martinez, line 1: I would harmonise this example to the bindings and also call `moveObjectContents`

bindings/kotlin/examples/GetObject.kt
DaughterOfMars: done

crates/iota-graphql-client/examples/get_object.rs
DaughterOfMars: right, done

bindings/kotlin/examples/GetObject.kt
thibault-martinez, line 1: We should rename the examples if we are not calling getObject

DaughterOfMars: to what?

Thoralf-M: MoveObjectContents?

DaughterOfMars: So in that case do we just not have a getObject example?

thibault-martinez: I mean, we don't have one in this PR but we probably should in the future

DaughterOfMars: But that's what this PR was supposed to be, and we decided it wasn't possible

Thoralf-M: It's not possible to decode the bytes into a struct you define on the bindings side at the moment, but there is more data to read. For getObject I would do something similar to this (we need to export the Display for some types)
```python

import asyncio


async def main():
    
    object = await client.object(ObjectId.from_hex("0x6"), None)
    print("Object ID:", object.object_id().to_hex())
    print("Object version:", object.version())
    print("Object previous transaction:", object.previous_transaction().to_base58())
    print("Object is shared:", object.owner().is_shared())
    print("Object storage rebate:", object.storage_rebate())
    print("Object type:", object.object_type())
    print("Object bcs bytes:", object.as_struct().contents.hex())

if __name__ == "__main__":
    asyncio.run(main())

```

DaughterOfMars: made those changes, thanks

bindings/kotlin/examples/GetObject.kt
```diff
@@ -0,0 +1,35 @@
...
+import iota_sdk.TransactionsFilter
+import kotlinx.coroutines.runBlocking
+
+fun main() = runBlocking {
+    try {
+        val client = GraphQlClient.newDevnet()
+
+        val objectId = ObjectId.fromHex("0x20c056090c3dd1604fcfd7ea759781de650aa45323738e799365d0c28bebeb1e")
+
+        val obj = client.`object`(objectId)!!
```
Thoralf-M, line 19: Looks amazing

bindings/kotlin/lib/iota_sdk/iota_sdk_ffi.kt
```diff
@@ -2141,6 +2141,15 @@ internal interface UniffiForeignFutureCompleteVoid : com.sun.jna.Callback {
...
 
+
+
+
+
+
+
+
+
+
```
Thoralf-M, line 2152: wtf is uniffi doing here?

DaughterOfMars: don't ask me lol

Alex6323: haha, I thought the same

Thoralf-M: Over 1000 newlines

bindings/go/examples/get_object.go
```diff
@@ -0,0 +1,51 @@
...
+
+	objType := "Package"
+	if obj.ObjectType().IsStruct() { 
+		objType = obj.ObjectType().AsStruct().String()
+	}
+
+	fmt.Println("Object ID:", obj.ObjectId().ToHex())
+    fmt.Println("Version:", obj.Version())
+    fmt.Println("Previous transaction:", obj.PreviousTransaction().ToBase58())
+    fmt.Println("Shared:", obj.Owner().IsShared())
```
Thoralf-M: How about `Owner: Shared` if that's true or `Owner : 0x...` in case of an address owned object?

DaughterOfMars: done

Thoralf-M (approved):

thibault-martinez (approved):

PR #165, opened 2025-08-28
feat(makefile): run all or specific bindings example

### LINKED ISSUE #160: Add chain_id.rs example to bindings

thibault-martinez:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/160

```
make go-examples
make go-example chain_id

make kotlin-examples
make kotlin-example chain_id

make python-examples
make python-example chain_id

make bindings-examples
make bindings-example chain_id
```


bindings/kotlin/examples/chain_id.kt
```diff
@@ -0,0 +1,13 @@
+import iota_sdk.GraphQlClient
```
Thoralf-M: Should we add a comment at the top for how to run it?

thibault-martinez: Assuming we are running it from where? Would get message if we need to add library paths etc

bindings/python/examples/chain_id.py
```diff
@@ -0,0 +1,14 @@
+from lib.iota_sdk_ffi import *
+
+import asyncio
+
+
+async def main():
+    client = GraphQlClient.new_devnet()
+    
+    chain_id = await client.chain_id()
+
```
Thoralf-M:
```suggestion
```

bindings/kotlin/examples/chain_id.kt
Thoralf-M: Would have said from the bindings language main folder
But yeah can be annoying with the library paths
Let's then at least add a section to each README then how to run an example using the make command?

bindings/python/examples/chain_id.py
thibault-martinez: I kind of prefer with the newline but ok

bindings/kotlin/examples/chain_id.kt
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/165/commits/3d600d0fadf4fd0b5f5d9a619189a5bd5103634d

will add the missing python readme

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/165/commits/71490f21dc780d53cb5a350f7e4edbc1313538e6

bindings/python/README.md
```diff
@@ -0,0 +1,37 @@
+# IOTA SDK - Python Bindings
+
```
Thoralf-M, line 2: We should probably show `make python` as first option, but can be done in another PR with the other README files

Thoralf-M (approved):

bindings/kotlin/examples/chain_id.kt
DaughterOfMars, line 1: Kotlin files are supposed to be named like `ChainId.kt` btw

thibault-martinez: Makes it simpler to run the examples if they have the same name across all languages

DaughterOfMars: Yes but my linter doesn't work when they are named wrong :c

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/165/commits/21dbe829d2a51e8b63d9b33776b7ff83bdf092c5

bindings/python/README.md
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/issues/168

bindings/kotlin/build.gradle.kts
```diff
@@ -16,11 +16,28 @@ dependencies {
...
+    // Get the example name from the command line argument -Pexample=<name>
+    val exampleProperty = "example"
+    inputs.property(exampleProperty, project.findProperty(exampleProperty) ?: "example")
+    
+    mainClass.set(provider {
+        val example = project.findProperty(exampleProperty)?.toString() ?: "example"
+        // Convert snake_case to CamelCase and append Kt
+        val className = example.split('_')
+            .map { it.replaceFirstChar { c -> c.uppercaseChar() } }
+            .joinToString("")
```
DaughterOfMars, line 33: Not needed anymore right?

thibault-martinez: It is, the argument still needs to be the same for the Makefile `make bindings-example chain_id` and then we transform it to CamelCase to find the Kotlin file

DaughterOfMars: Ah, I see

DaughterOfMars (approved):

PR #170, opened 2025-08-29
feat: Add support for IOTA-Names

DaughterOfMars:
## Descriptions

Adds three graphql queries and types to support IOTA-Names.


crates/iota-graphql-client/src/lib.rs
```diff
@@ -1740,6 +1746,100 @@ impl Client {
                 .collect::<Result<Vec<_>>>()?,
         ))
     }
+
+    pub async fn iota_names_lookup(&self, name: &str) -> Result<Option<Address>> {
```
Thoralf-M, line 1707: Please add the doc comments from the core repo for the methods, like `/// Return the resolved record for the given name.` here

crates/iota-sdk-types/src/type_tag/mod.rs
```diff
@@ -267,6 +302,85 @@ impl PartialEq<str> for Identifier {
...
+
+impl std::borrow::Borrow<IdentifierRef> for Identifier {
+    fn borrow(&self) -> &IdentifierRef {
+        self
+    }
+}
+
+#[derive(Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[repr(transparent)]
+pub struct IdentifierRef(str);
```
Thoralf-M, line 319: What's the benefit of this type over the ident_str macro?

DaughterOfMars: Well we still have this type in the core repo, it's just named `IdentStr` which I didn't like as much. I just moved the macro to the const constructor

DaughterOfMars: I should note that I found a pretty big issue with the macro while testing this. It does not detect invalid characters so long as they are valid ascii. I cannot figure out why tbh, it's weird.

Thoralf-M (approved):
`Default name: name.iota` :tada: 
```python

import asyncio


async def main():
    
    name = await client.iota_names_default_name(Address.from_hex("0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900"), None)
    print("Default name:", name.format(NameFormat.DOT))

if __name__ == "__main__":
    asyncio.run(main())
```

Thoralf-M (approved):

crates/iota-sdk-types/src/iota_names/error.rs
```diff
@@ -0,0 +1,32 @@
...
+    #[error("Hyphens are not allowed as first or last character of a label")]
+    HyphensAsFirstOrLastLabelChar,
+    #[error(
+        "Only lowercase letters, numbers, and hyphens are allowed as label characters. Got \'{0}\' at position {1}"
+    )]
+    InvalidLabelChar(char, usize),
+    #[error("Name must contain at least two labels, TLN and SLN")]
+    NotEnoughLabels,
+    #[error("Name must include only one separator")]
+    InvalidSeparator,
```
Alex6323: tbh this error is a bit confusing since we actually have two separators and it's not clear that this error and its message make only sense for the `@` separator. 

Can we be more explicit here?

DaughterOfMars: sure, done

Alex6323 (approved):

PR #180, opened 2025-09-04
chore(examples): Add transactions with shared object examples

### LINKED ISSUE #148: Fetch transactions that had a shared object as input

DaughterOfMars:
## Description

Adds examples which demonstrate fetching all transactions using a shared object.

Closes #148

### DIFF
--- a/bindings/go/examples/transactions_with_shared.go
+++ b/bindings/go/examples/transactions_with_shared.go
@@ -1,3 +1,6 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
 package main
 
 import (
--- a/bindings/go/examples/transactions_with_shared.go
+++ b/bindings/go/examples/transactions_with_shared.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"fmt"
+	"log"
+
+	sdk "bindings/iota_sdk_ffi"
+)
+
+func isNilError(err error) bool {
+	if sdkErr, ok := err.(*sdk.SdkFfiError); ok {
+		return sdkErr == nil
+	}
+	return false
+}
+
+func main() {
+	client := sdk.GraphQlClientNewDevnet()
+
+	sharedObjId, err := sdk.ObjectIdFromHex("0x07c59b37bd7d036bf78fa30561a2ab9f7a970837487656ec29466e817f879342")
+	if err != nil {
+		log.Fatalf("Failed to create object ID: %v", err)
+	}
+
+	transactions, err := client.Transactions(sdk.PaginationFilter{}, &sdk.TransactionsFilter{InputObject: &sharedObjId})
+	if !isNilError(err) {
+		log.Fatalf("Failed to get transactions: %v", err)
+	}
+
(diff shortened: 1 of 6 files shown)


bindings/go/examples/transactions_with_shared.go
thibault-martinez, line 1: copyright

DaughterOfMars: none of them have them

Thoralf-M (approved):

crates/iota-graphql-client/examples/transactions_with_shared.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    let shared_obj_id =
+        ObjectId::from_str("0x07c59b37bd7d036bf78fa30561a2ab9f7a970837487656ec29466e817f879342")?;
+    let transactions = client
+        .transactions(
+            TransactionsFilter {
+                input_object: Some(shared_obj_id),
+                ..Default::default()
+            },
+            PaginationFilter::default(),
```
thibault-martinez, line 22: Please also use direction forward to align with the other examples

DaughterOfMars: I dunno, this is how I would do it in rust. Doesn't really make sense esp since this is an example in the graphql client that has nothing to do with the bindings

bindings/kotlin/examples/TransactionsWithShared.kt
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+import iota_sdk.GraphQlClient
+import iota_sdk.ObjectId
+import iota_sdk.PaginationFilter
+import iota_sdk.Direction
+import iota_sdk.TransactionsFilter
```
thibault-martinez, line 8: Just curious, no way to group them?

crates/iota-graphql-client/examples/transactions_with_shared.rs
thibault-martinez: fair enough, there is no notion of default in the other languages?

thibault-martinez (approved):

bindings/kotlin/examples/TransactionsWithShared.kt
Thoralf-M: No, so bad
Only alternative is to do `*` import, not sure if we prefer that

crates/iota-graphql-client/examples/transactions_with_shared.rs
DaughterOfMars: we can't export it but we could just create a default method

bindings/kotlin/examples/TransactionsWithShared.kt
thibault-martinez: Nah, that's fine

crates/iota-graphql-client/examples/transactions_with_shared.rs
thibault-martinez: Ok, doesn't matter much

PR #187, opened 2025-09-05
chore(graphql): Improve graphql API

DaughterOfMars:
## Description

Some minor improvements, mostly to the rust graphql API.

### DIFF
--- a/crates/iota-graphql-client/src/lib.rs
+++ b/crates/iota-graphql-client/src/lib.rs
@@ -215,9 +215,10 @@ impl Client {
     pub async fn coins_stream(
         &self,
         address: Address,
-        coin_type: Option<String>,
+        coin_type: impl Into<Option<String>>,
         streaming_direction: Direction,
     ) -> impl Stream<Item = Result<Coin>> + '_ {
+        let coin_type = coin_type.into();
         stream_paginated_query(
             move |filter| self.coins(address, coin_type.clone(), filter),
             streaming_direction,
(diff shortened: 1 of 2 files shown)


crates/iota-graphql-client/src/lib.rs
```diff
@@ -291,11 +294,7 @@ impl Client {
     /// prevent access to objects that are owned by addresses other than the
     /// sender, and calling non-public, non-entry functions, and some other
     /// checks. Defaults to false.
-    pub async fn dry_run_tx(
-        &self,
-        tx: &Transaction,
-        skip_checks: Option<bool>,
-    ) -> Result<DryRunResult> {
+    pub async fn dry_run_tx(&self, tx: &Transaction, skip_checks: bool) -> Result<DryRunResult> {
```
Thoralf-M, line 297: Why make the skip_checks required? Also comment is outdated with that

Alex6323: I guess one could argue that it's better having to provide `false` instead of `None` as it's more explicit. Generally you want to avoid calls like `my_func(None, None, Some(a))`, when you can have `my_func(true, false, a)` instead.

Alex6323 (approved):

crates/iota-graphql-client/src/lib.rs
DaughterOfMars: Yeah exactly. Previously there were three options to control a bool, which makes no sense

Thoralf-M: But then please update the comment, it says `optional flag`, but maybe still mention that `false` is the default one would use here

DaughterOfMars: done

Thoralf-M (approved):

PR #188, opened 2025-09-05
chore(examples): Add example to create a txn for sending iota

### LINKED ISSUE #154: Send IOTA, Prepare transaction bytes (base64 encoded)
Just prepare the transaction bytes for a tx that sends IOTA and print them base64 encoded so it could then be signed

DaughterOfMars:
## Description

Adds an example which creates a transaction that sends some IOTA and then serializes the transaction for signing.

Closes #154


bindings/go/examples/prepare_send_iota.go
```diff
@@ -0,0 +1,72 @@
...
+	res, err := client.DryRunTx(txn, nil)
+	if err.(*sdk.SdkFfiError) != nil {
+		log.Fatalf("Failed to get gas price: %v", err)
+	}
+
+	if res.Error != nil {
+		log.Fatalf("Failed to send IOTA: %v", *res.Error)
+	}
+
+	log.Print("Send IOTA dry run was successful!")
```
Thoralf-M: It's not printing the serialized tx bytes

DaughterOfMars: right, fixed

bindings/go/examples/prepare_send_iota/main.go
```diff
@@ -0,0 +1,79 @@
...
+	txnBytes, err := txn.BcsSerialize()
+	if err != nil {
+		log.Fatalf("Failed to serialize transaction: %v", err)
+	}
+	log.Printf("Signing Digest: %v", sdk.HexEncode(txn.SigningDigest()))
+	log.Printf("Txn Bytes: %v", sdk.Base64Encode(txnBytes))
+
+	res, err := client.DryRunTx(txn, nil)
+	if err.(*sdk.SdkFfiError) != nil {
+		log.Fatalf("Failed to get gas price: %v", err)
```
Thoralf-M: Wrong error message

crates/iota-sdk-ffi/src/lib.rs
```diff
@@ -115,3 +117,23 @@ mod types;
...
+}
+
+#[uniffi::export]
+pub fn base64_decode(input: String) -> crate::error::Result<Vec<u8>> {
+    Ok(base64ct::Base64::decode_vec(&input)?)
+}
+
+#[uniffi::export]
+pub fn hex_encode(input: &[u8]) -> String {
+    hex::encode(input)
```
Thoralf-M, line 133: Is that with 0x prefix or without?

DaughterOfMars: without

Thoralf-M: Wondering if we should have it, but we can add it in another PR later if needed

Thoralf-M (approved):

Alex6323 (approved):

PR #192, opened 2025-09-08
chore(examples): Add stake and unstake example

### LINKED ISSUE #157: Stake/unstake

DaughterOfMars:
## Description

Adds an example which stakes and unstakes.

Closes #157


bindings/go/examples/stake.go
```diff
@@ -0,0 +1,113 @@
...
+	res, err := client.DryRunTx(txn, nil)
+	if !isNilError(err) {
+		log.Fatalf("Failed to get gas price: %v", err)
+	}
+
+	if res.Error != nil {
+		log.Fatalf("Failed to stake: %v", *res.Error)
+	}
+
+	log.Print("Successfully staked!")
```
Alex6323: Should we log the amount and the validator address, like "Successfully staked x IOTA to validator y"?

crates/iota-graphql-client/examples/unstake.rs
```diff
@@ -0,0 +1,78 @@
...
+    );
+    builder.add_gas_objects([Input::from(&gas_coin).with_owned_kind()]);
+    let txn = builder.finish()?;
+    let res = client.dry_run_tx(&txn, false).await?;
+
+    if let Some(err) = res.error {
+        anyhow::bail!("Failed to unstake: {err}");
+    }
+
+    println!("Successfully unstaked!");
```
Thoralf-M: Wondering if we should mention that it's just a dry run like `Unstake dry run was successful`

bindings/go/examples/stake.go
```diff
@@ -0,0 +1,113 @@
...
+
+	requestAddStakeFn, err := sdk.NewIdentifier("request_add_stake")
+	if err != nil {
+		log.Fatalf("Failed to parse identifier: %v", err)
+	}
+
+	builder := sdk.NewTransactionBuilder()
+	inputs := []*sdk.Argument{
+		builder.Input(sdk.UnresolvedInputNewShared(iotaSystemId, 1, true)),
+		builder.Input(sdk.UnresolvedInputFromObject(*coin).WithOwnedKind()),
```
Thoralf-M, line 86: If it's `UnresolvedInput` shouldn't we be able to construct it with the object id alone, without requesting the coin object first?

DaughterOfMars: We also need the digest and version

Thoralf-M: But isn't this the point of UnresolvedInputFromObject to resolve this data later in the build step? Or why is it called `Unresolved`?

crates/iota-graphql-client/examples/unstake.rs
DaughterOfMars: Changed them

bindings/go/examples/stake.go
DaughterOfMars: I added in a line to log the validator name. Logging the coin amount would require additional (fallible) code to convert the object to a coin and I don't think it's worth it

bindings/go/examples/stake.go
DaughterOfMars: You would think so but that is not the case

Thoralf-M (approved):

Alex6323 (approved):

PR #195, opened 2025-09-09
feat(examples): add transactions_with_function example

### LINKED ISSUE #152: Fetch transactions that called a specific move function

Thoralf-M:
Add examples to show to fetch transactions that called a specific move function

Fixes https://github.com/iotaledger/iota-rust-sdk/issues/152


bindings/go/examples/transactions_with_function.go
```diff
@@ -0,0 +1,34 @@
...
+	if sdkErr, ok := err.(*sdk.SdkFfiError); ok {
+		return sdkErr == nil
+	}
+	return false
+}
+
+func main() {
+	client := sdk.GraphQlClientNewDevnet()
+
+	function := "0x3::iota_system::request_add_stake"
```
Alex6323: you don't bind that string in the other examples. why here?

Thoralf-M: Because go is an amazing language :trollface: 
```
In Go, you cannot take the address of a string literal directly (e.g., [Function: &"..."] is not allowed). You must assign the string to a variable and then take its address, as in your current code.
```

Thoralf-M: If one need to define multiple strings one can make use of a function like this `func strPtr(s string) *string { return &s }` but that's not worth it for this example where it's a single string

crates/iota-graphql-client/examples/transactions_with_function.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
```
Alex6323: could use the `Result` from the error module?

DaughterOfMars (approved):

Alex6323 (approved):

crates/iota-graphql-client/examples/transactions_with_function.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/195/commits/fd5445221e60d0e5cd2ae48ed20146585c5571de

PR #196, opened 2025-09-09
chore(bindings): Reorder params involving pagination

DaughterOfMars:
## Description

Reorders the graphql function params which take a pagination filter so that it is always last, since it can be defaulted for most cases and most languages like null params at the end of the signature.


bindings/python/examples/transactions_with_shared.py
```diff
@@ -14,7 +14,6 @@ async def main():
         )
 
         transactions = await client.transactions(
-            PaginationFilter(direction=Direction.FORWARD),
```
Thoralf-M, line 17: Really not needed anymore?

DaughterOfMars: It hasn't been since my other PR but now it's at the end of the params list

Thoralf-M (approved):

Alex6323 (approved):

PR #197, opened 2025-09-09
chore(examples): Fix go folder layout

DaughterOfMars:
## Description

Updates the go examples folder structure to avoid lint issues.


bindings/go/common/common.go
Thoralf-M, line 1: Not sure about this change, remember we have the issue in the monorepo about examples that can't be copy pasted, because they rely on such a helper file

Alex6323: I agree. We should instead silence the linter in this case, because it's not able to see the goal and intent of the examples structure.

DaughterOfMars: No way to do that unfortunately

Thoralf-M (approved):

Alex6323 (approved):

PR #198, opened 2025-09-09
feat(examples): add prepare_send_iota_multi example

Thoralf-M:
add prepare_send_iota_multi example


Alex6323 (approved):

DaughterOfMars (commented):
Kind of a duplicate of `prepare_send_iota`, except with many recipients. Maybe we call it `prepare_send_iota_multi` and then also print out the txn bytes and stuff?

DaughterOfMars (approved):

PR #200, opened 2025-09-10
feat(examples): add pagination example

### LINKED ISSUE #189: Add pagination example
Add an example where there is more data than fits into a single page and then show in the example how to use the cursor to fetch all the data

Thoralf-M:
Add examples to show how to fetch all pages using the nextCursor of the pageInfo

Fixes https://github.com/iotaledger/iota-rust-sdk/issues/189

```
make bindings-example pagination
make[1]: Verzeichnis „/home/t/Desktop/iota-rust-sdk“ wird betreten

Running Go example "pagination"
Fetching page with cursor: nil
Fetching page with cursor: IAsCcO6dJ9oNsJZR5fczjfoyx+5kQczvofbjBXNbz8erj6o4BwAAAAA=
Fetching page with cursor: IC18jxuTcLkdf/Pb2BRKBdadf5d1jemMW7bDJXjNpiIUj6o4BwAAAAA=
3 objects fetched:
0x0b0270ee9d27da0db09651e5f7338dfa32c7ee6441ccefa1f6e305735bcfc7ab
0x2d7c8f1b9370b91d7ff3dbd8144a05d69d7f97758de98c5bb6c32578cda62214
0xd04077fe3b6fad13b3d4ed0d535b7ca92afcac8f0f2a0e0925fb9f4f0b30c699
```

### DIFF
--- a/bindings/go/examples/pagination/main.go
+++ b/bindings/go/examples/pagination/main.go
@@ -36,7 +36,7 @@ func main() {
 			break
 		}
 	}
-	fmt.Printf("Total objects fetched(%d):\n", len(allObjects))
+	fmt.Printf("%d objects fetched:\n", len(allObjects))
 	for _, obj := range allObjects {
 		fmt.Println(obj.ObjectId().ToHex())
 	}
--- a/bindings/go/examples/pagination/main.go
+++ b/bindings/go/examples/pagination/main.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"fmt"
+
+	sdk "bindings/iota_sdk_ffi"
+)
+
+func main() {
+	client := sdk.GraphQlClientNewDevnet()
+	address, err := sdk.AddressFromHex("0x611830d3641a68f94a690dcc25d1f4b0dac948325ac18f6dd32564371735f32c")
+	if err != nil {
+		panic(err)
+	}
+
+	// Limit to 1 to demonstrate pagination
+	limit := int32(1)
+	var allObjects []*sdk.Object
+	var nextCursor *string
+	for {
+		if nextCursor != nil {
+			fmt.Printf("Fetching page with cursor: %s\n", *nextCursor)
+		} else {
+			fmt.Printf("Fetching page with cursor: nil\n")
+		}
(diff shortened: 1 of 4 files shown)


crates/iota-graphql-client/examples/pagination.rs
```diff
@@ -0,0 +1,47 @@
...
+            .await?;
+        let (page_info, data) = owned_objects_page.into_parts();
+        all_objects.extend(data);
+        if page_info.has_next_page {
+            next_cursor = page_info.end_cursor.clone();
+        } else {
+            break;
+        }
+    }
+    println!("Total objects fetched({}):", all_objects.len());
```
DaughterOfMars: Bit of a weird message format. Maybe we can do something like this?
```suggestion
    println!("{} objects fetched:", all_objects.len());
```

Thoralf-M: Sure

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/200/commits/488b1ac3cdbbccf859a6e3d76261ff687f51fc96

DaughterOfMars (approved):

bindings/go/examples/pagination/main.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	sdk "bindings/iota_sdk_ffi"
+)
+
+func main() {
+	client := sdk.GraphQlClientNewDevnet()
+	address, err := sdk.AddressFromHex("0x611830d3641a68f94a690dcc25d1f4b0dac948325ac18f6dd32564371735f32c")
+	if err != nil {
+		panic(err)
```
Alex6323: I think in the other examples we use `log.Fatalf`. Wondering if we should do the same to be consistent?

Thoralf-M: Good point `log.Fatalf` actually seems to be the preferred case for these cases, changed it https://github.com/iotaledger/iota-rust-sdk/pull/200/commits/0a311554657ddbf6fe8c4422a436da6b774ef43b

Alex6323 (approved):

PR #203, opened 2025-09-11
chore(examples): Allow custom queries in bindings

### LINKED ISSUE #163: Add custom_query.rs example to bindings

Alex6323:
Fixes #163


crates/iota-graphql-client/src/lib.rs
```diff
@@ -403,6 +403,18 @@ impl Client {
         Ok(res)
     }
 
+    pub async fn run_custom_query<T: Serialize>(&self, query: &T) -> Result<serde_json::Value> {
```
DaughterOfMars: Since this can only accept json, maybe we should just take a `serde_json::Value`?

DaughterOfMars: Actually it may be better to take a specific type like `CustomQuery` to enforce the structure

Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/203/commits/e09d820dc70a5b8f622d62e8cf7fe408c5f29c59

bindings/go/examples/custom_query/main.go
```diff
@@ -0,0 +1,49 @@
...
+		Query:     query,
+		Variables: &variables,
+	}
+	res1, err := client.RunCustomQuery(payload1)
+	if err.(*sdk.SdkFfiError) != nil {
+		log.Fatalf("Failed to perform a custom query: %v", err)
+	}
+	fmt.Println(res1)
+
+	payload2 := sdk.CustomQuery{
```
Thoralf-M: Instead of payload1 and 2, how about naming them like `queryWithVariables` and `query`?

crates/iota-graphql-client/src/lib.rs
```diff
@@ -403,7 +411,12 @@ impl Client {
         Ok(res)
     }
 
-    pub async fn run_custom_query<T: Serialize>(&self, query: &T) -> Result<serde_json::Value> {
+    /// Run a query on the GraphQL server and return the response.
+    /// This method expects a valid GraphQL query string and valid JSON for the
+    /// variables. It performs no validation of the input. It returns
+    /// [`serde_json::Value`] and it is intended to be used with queries
+    /// that need to cross the FFI boundary.
```
DaughterOfMars: This fn is for the rust SDK, I wouldn't mention FFI

crates/iota-sdk-ffi/src/graphql.rs
```diff
@@ -865,9 +870,17 @@ impl GraphQLClient {
     }
 }
 
-#[derive(Debug, serde::Serialize, uniffi::Record)]
+#[derive(Debug, uniffi::Record)]
 pub struct CustomQuery {
     pub query: String,
-    #[serde(default)]
     pub variables: Option<serde_json::Value>,
```
DaughterOfMars, line 885: Make it default for uniffi?

DaughterOfMars (commented):
Rust example?

crates/iota-graphql-client/src/lib.rs
```diff
@@ -154,6 +154,14 @@ impl DynamicFieldOutput {
     }
 }
 
+/// A type to allow GraphQl queries across the FFI boundary.
+#[derive(Debug, serde::Serialize)]
+pub struct UncheckedQuery {
+    pub query: String,
+    #[serde(default)]
+    pub variables: Option<serde_json::Value>,
```
DaughterOfMars: Isn't this always supposed to be a HashMap?

Alex6323:
> Rust example?

Rust example exists already.

DaughterOfMars:
> > Rust example?
> 
> Rust example exists already.

Huh? But you just added the fn in this PR

crates/iota-graphql-client/src/lib.rs
Alex6323: I mentioned it so that people know better not to call it from Rust but go the route as in `custom_query.rs`, but sure I can remove it.

crates/iota-sdk-ffi/src/graphql.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/203/commits/f012e7b165fa0a5adbb17cafa5bccc31248ce185

crates/iota-graphql-client/src/lib.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/203/commits/f012e7b165fa0a5adbb17cafa5bccc31248ce185

crates/iota-graphql-client/src/lib.rs
Alex6323: yeah, but I found there's not much of a point converting it to a `HashMap` when it's just serialized back to Json in the query call, and if you stay on Rust side, you would alway define your custom queries with an actual type and therefore not need an `UncheckedQuery`.

Alex6323: While in Rust land we can guarantee valid queries through the cynic derives, it's not really doable for the bindings. So they would have to use their own graphql libraries before sending it accross the FFI boundary. At least that's my current understanding.

Thoralf-M (approved):

crates/iota-graphql-client/src/lib.rs
DaughterOfMars: I mean in the bindings we should be able to create a map of variables and pass it in, so this type should have a map too

Alex6323: done

crates/iota-sdk-ffi/Cargo.toml
```diff
@@ -20,7 +20,9 @@ derive_more = { version = "2.0", features = ["from", "deref", "display"] }
 hex = "0.4.3"
 rand = "0.8"
 roaring = { version = "0.11.2", default-features = false }
+serde = { version = "1.0.144" }
 serde_json = "1.0.95"
+strum = { version = "0.27.2", features = ["derive"] }
```
Thoralf-M: Not needed anymore?

Alex6323: good catch

Thoralf-M (approved):

PR #206, opened 2025-09-17
feat: Add new transaction builder

### LINKED ISSUE #130: Can we remove unresolved::Input?
This struct in iota-transaction-builder seems to be kind of useless. Can we remove it?

### LINKED ISSUE #134: Remove unused types in `iota-transaction-builder`?
There are several types, like `unresolved::Transaction` in this crate which are totally useless. We should remove them.

### LINKED ISSUE #219: Rework transaction builder
The current one is very low level and not nice to use
Inputs should be possible to be provided by object id alone and then be resolved by a client, right now the user has to do that manually, making even simple things quite complex

DaughterOfMars:
## Description

This PR implements an entirely new transaction builder based on experimental changes originally made on the core repository. Fundamentally, this allows a much simpler API in Rust, while also somewhat simplifying the bindings API.

### Features

- Simple API for calling basic functions and handling results.
- Methods take simpler types, like `&str` instead of `IdentStr` or `Identifier`.
- Methods use generic types to allow passing many usable data formats.
- Automatic resolution of objects using the client.
- Builder patterns for constructing move calls.
- Any type that implements `MoveParams` can be used for move call params. Tuples up to length 6 are implemented. Longer param lists can be achieved using nested tuples.
- Custom move types via `CustomMoveType` (with helper macro `move_type!`)
- Generic typing rather than constructing `TypeTag`s (via `MoveType` trait).
- Command results can be bound by a string name, and referenced by later calls with that name using `Res`.
- Parameters can be passed in directly without fiddling with inputs, and will be automatically handled (via `MoveParam` trait).
- Mixed parameters (new inputs and named references) can be defined in a single tuple.
- The builder is chainable, except for construction and finishing.


crates/iota-transaction-builder/src/builder.rs
```diff
@@ -0,0 +1,1070 @@
...
+        for gas_to_resolve in self.gas_to_resolve.drain(..) {
+            let obj = self
+                .client
+                .object(gas_to_resolve, None)
+                .await
+                .map_err(Error::Client)?
+                .ok_or_else(|| Error::Input(format!("missing object {gas_to_resolve}")))?;
+
+            let obj_ref = match obj.owner() {
+                Owner::Address(_) | Owner::Object(_) | Owner::Immutable => {
```
Thoralf-M: Immutable also can't be used as gas object
Also it would probably be good to make sure that the gas objects match the gas owner address that's set in `.finish()` `owner: self.sponsor.unwrap_or(self.sender),`

crates/iota-transaction-builder/src/builder.rs
```diff
@@ -0,0 +1,1070 @@
...
+            sender: self.sender,
+            gas_payment: GasPayment {
+                objects: self.gas,
+                owner: self.sponsor.unwrap_or(self.sender),
+                price,
+                budget: self.gas_budget.unwrap_or(0),
+            },
+            expiration: self.expiration,
+        };
+        println!("{txn:#?}");
```
Thoralf-M: Remove or maybe convert to a debug log?

crates/iota-transaction-builder/src/builder.rs
```diff
@@ -0,0 +1,1070 @@
...
+    /// A move object whose owner is "Shared"
+    Shared {
+        object_id: ObjectId,
+        /// Controls whether the caller asks for a mutable reference to the
+        /// shared object.
+        mutable: bool,
+    },
+    /// A move object that is attempted to be received in this transaction.
+    Receiving(ObjectId),
+    Resolved(Input),
```
Thoralf-M: Bit weird that it's `UnresolvedInput::Resolved()`

crates/iota-transaction-builder/src/builder.rs
```diff
@@ -0,0 +1,1070 @@
+// Copyright 2024 IOTA Stiftung
```
Thoralf-M: 2025 in all the new files

crates/iota-transaction-builder/src/builder.rs
```diff
@@ -0,0 +1,1070 @@
...
+    /// a gas object and use the [`add_gas_objects`](Self::add_gas_objects)
+    /// method to set the gas objects.
+    pub(crate) gas: Vec<ObjectReference>,
+    gas_to_resolve: Vec<ObjectId>,
+    /// The gas budget for the transaction.
+    gas_budget: Option<u64>,
+    /// The gas price for the transaction.
+    gas_price: Option<u64>,
+    /// The sender of the transaction.
+    sender: Address,
```
Thoralf-M: I'm wondering if this should also be made optional, because for some dev inspect txs the sender doesn't matter

crates/iota-transaction-builder/src/builder.rs
DaughterOfMars: Oops lol

crates/iota-transaction-builder/src/builder.rs
DaughterOfMars: Yeah. This is a bit of a hacky way to do this. I will give it some more thought.

crates/iota-transaction-builder/src/builder.rs
DaughterOfMars: Is it even possible to use an object owner for gas?

crates/iota-transaction-builder/src/builder.rs
DaughterOfMars: I renamed it `Input`, but the functionality is the same currently

crates/iota-transaction-builder/src/builder.rs
DaughterOfMars: I'm imagining a future scenario where we don't need to pass it at all, because we can look it up in a file or something like `client.yaml`

Alex6323:
What I like about this PR:

- relieves the user of so much boilerplate code (automatic object resolution) 
- the `NamedCommands` chaining by referencing previous results is great
- the implementation is Rust Art imho :heart: :crab: 

What I don't like that much:
- generics make this API very hard to explore without intense study: the former transaction builder may be a bit clunky, but the API clearly told the user what it expects as input arguments. Now if you look at the method signatures you see a lot of trait bounds, but it's not clear at all what you have to pass in. Sure, this can be solved by good documentation, but I fear that for developers this may be intimidating.
- I would remove the `dry_run` fn from the builder, because although it may be handy to skip the `finish` method, it's kind of an unexpected API. If you have a `TransactionBuilder`, I think Rust devs are conditioned to HAVE TO call a finish method at then end. It's still there I know, but I would prefer to enforce this, also because `builder.dry_run()` reads oddly.
- some naming
- higher complexity of the implementation (builders within builders)
- I don't understand yet the use-case for the `TransactionBuilder<()>` (non-client type). Maybe it needs examples for that as well to show why it's still useful to have?

crates/iota-transaction-builder/src/builder.rs
Thoralf-M: Why require it in `pub fn new(sender: Address) -> TransactionBuilder<()> {` then?

DaughterOfMars: Because right now there is no way to get the sender from the client

Thoralf-M: But if it's not optional now, this would be a breaking change later. And for dev inspect, one could just use an address like 0x0 as sender as default if not provided

DaughterOfMars: ehh, okay.

DaughterOfMars: I just really don't like that you can fail at runtime because you forgot to set the sender

crates/iota-transaction-builder/src/builder.rs
Thoralf-M: Seems like it's not allowed, so should be Owner::Address() only

DaughterOfMars: fixed

crates/iota-transaction-builder/src/builder.rs
Thoralf-M: Fair point, I just thought it's annoying for users to create an address they don't care about, but I guess it's fine since depending on what they call it can actually have an impact also for dev inspects

crates/iota-sdk-types/src/type_tag/mod.rs
```diff
@@ -267,6 +300,76 @@ impl PartialEq<str> for Identifier {
...
+        //     IdentStr(str) is #[repr(transparent)]).
+        // (3) both in and out lifetimes are 'static, so we're not widening the
+        // lifetime. (4) we've just asserted that the IdentStr passes the
+        // is_valid check.
+        unsafe { std::mem::transmute::<&'static str, &'static Self>(s) }
+    }
+    /// Returns true if this string is a valid identifier.
+    pub fn is_valid(s: &str) -> bool {
+        Identifier::is_valid(s)
+    }
```
Thoralf-M, line 338: Why have this? There isn't even a way to construct a IdentifierRef that would be invalid?

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,1066 @@
...
+        name.push_named_commands(self);
+    }
+
+    /// Get the value for the given string in the named commands map
+    pub fn get_named_command(&self, name: &str) -> Option<Argument> {
+        self.named_commands.get(name).copied()
+    }
+}
+
+impl TransactionBuilder<()> {
```
Thoralf-M: Why have a separated block like this with the `new()` and `impl<C> TransactionBuilder<C> {` in between them?

crates/iota-transaction-builder/src/builder.rs
```diff
@@ -0,0 +1,1066 @@
...
+
+        name.push_named_commands(ptb);
+
+        ptb
+    }
+}
+
+/// A builder for a move call command within a programmable transaction.
+#[derive(Debug)]
+pub struct PublishBuilder<'a> {
```
Thoralf-M: When do you create an own builder and when not? For example there is no builder for upgrade

crates/iota-transaction-builder/src/builder.rs
```diff
@@ -0,0 +1,1066 @@
...
+                tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+            }
+        }
+        Ok(res)
+    }
+}
+
+/// A builder for a move call command within a programmable transaction.
+#[derive(Debug)]
+pub struct MoveCallCommandBuilder<'a, C, G: MoveTypes = (), A = ()> {
```
Thoralf-M: How about moving these sub builders into an own file? The builder.rs is quite large

crates/iota-transaction-builder/src/builder.rs
```diff
@@ -0,0 +1,1066 @@
...
+        if let Some(arg) = ptb.named_commands.get(self.0) {
+            args.push(*arg);
+        } else {
+            panic!("no command named `{}` exists", self.0)
+        }
+    }
+}
+
+/// A trait that defines a named command, either a string or nothing.
+pub trait NamedCommand {
```
Thoralf-M: Can you show an example for what this is good for/how it's supposed to be used?

DaughterOfMars: Basically you name the output of a command with this trait and a string key, and then you can refer to that output in subsequent commands by using `Res`

crates/iota-transaction-builder/src/builder.rs
DaughterOfMars: Yeah, I will do that

crates/iota-transaction-builder/src/builder.rs
DaughterOfMars: I created them when multiple steps were needed, some of which are optional. For instance, in this case there are two alternate ways to finish the builder. In the move call builder two of the fns are optional.

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: I'll take a look at the formatting

crates/iota-sdk-types/src/type_tag/mod.rs
DaughterOfMars: Well this is an associated fn, not a method 🤷‍♀️

Thoralf-M: Ah right nvm

crates/iota-transaction-builder/src/builder.rs
```diff
@@ -0,0 +1,1066 @@
...
+            module: self.module,
+            function: self.function,
+            args: self.args,
+            generics: Err(tags.into_iter().collect()),
+            ptb: self.ptb,
+        }
+    }
+
+    /// Finish the move call and return the PTB.
+    pub fn end(self) -> &'a mut TransactionBuilder<Client> {
```
Thoralf-M: This looks a bit weird when used, wouldn't it be better to build a move call separated and provide it to the TransactionBuilder in a single method?

crates/iota-transaction-builder/src/builder.rs
Thoralf-M: Using `pub fn result(self, name: impl NamedCommands)`, what happens if I don't do that? Can the result then not be referenced?

DaughterOfMars: Yeah, it's the only way with the current system.

crates/iota-transaction-builder/src/builder.rs
DaughterOfMars: Maybe, but I wanted to keep everything in line. I was thinking about whether I could eliminate this fn but it's complicated

crates/iota-transaction-builder/src/builder.rs
Thoralf-M: Wouldn't it be nicer to have the result assigned to a variable the user can pass in somewhere?

DaughterOfMars: But then you couldn't chain the calls

crates/iota-graphql-client/examples/prepare_send_iota_multi.rs
```diff
@@ -3,78 +3,42 @@
...
-        (
+            Res("coin1"),
+        )
+        .transfer_objects(
             Address::from_str(
                 "0x2222b466a24399ebcf5ec0f04820812ae20fea1037c736cfec608753aa38b522",
             )?,
-            2_000_000_000u64,
-        ),
-    ];
```
Thoralf-M, line 43: I wanted to have the amounts and addresses be defined in an array so one can easily add more recipients without having to manually update the transaction builder
But with the named results this seems to be a bit complicated now
```Rust
    let recipients = [
        (
            Address::from_str(
                "0x111173a14c3d402c01546c54265c30cc04414c7b7ec1732412bb19066dd49d11",
            )?,
            1_000_000_000u64,
        ),
        (
            Address::from_str(
                "0x2222b466a24399ebcf5ec0f04820812ae20fea1037c736cfec608753aa38b522",
            )?,
            2_000_000_000u64,
        ),
    ];

    let mut builder = TransactionBuilder::new(sender_address).with_client(client.clone());

    // Extract amounts from recipients
    let amounts: Vec<u64> = recipients.iter().map(|(_, amt)| *amt).collect();
    // Annoying part now
    let labels: Vec<&'static str> = (0..recipients.len())
        .map(|i| Box::leak(format!("coin{}", i + 1).into_boxed_str()) as &'static str)
        .collect();

    builder.split_coins(gas_coin, amounts, labels.clone());

    // Transfer each split coin to the corresponding recipient
    for (i, (address, _)) in recipients.iter().enumerate() {
        builder.transfer_objects(*address, Res(labels[i]));
    }

    builder.gas(gas_coin).gas_budget(1000000000);

    let txn = builder.finish().await?;
```

DaughterOfMars: So, I could fix this by using a method `fn res(s: impl Into<String>) -> Res` instead of directly using the struct. Alternatively, I could possibly add some kind of wrapper around the builder which would allow follow-up commands like naming the output or returning the `Argument`. I will think about it.

crates/iota-transaction-builder/src/builder/command_builder.rs
```diff
@@ -0,0 +1,295 @@
...
+        recipient: Address,
+        amount: impl Into<Option<u64>> + Send,
+    ) -> &'a mut TransactionBuilder<C> {
+        self.end().transfer_iota(recipient, amount)
+    }
+}
+
+impl<'a, L: Into<Command>> CommandBuilder<'a, (), L> {
+    /// Set the gas coins that will be consumed. Optional.
+    pub fn gas(self, obj_ref: ObjectReference) -> &'a mut TransactionBuilder<()> {
```
Thoralf-M: I'm a bit confused by this, why is this on the CommandBuilder?

DaughterOfMars: The idea is to allow you to call these other methods in a chain without needing to call `end` or something to get the builder back

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,684 @@
...
+            sender,
+            sponsor: Default::default(),
+            expiration: Default::default(),
+            named_commands: Default::default(),
+            client: (),
+        }
+    }
+
+    /// Set the gas coins that will be consumed. Optional.
+    pub fn gas(&mut self, obj_ref: ObjectReference) -> &mut Self {
```
Thoralf-M: only accepts a single ObjectReference, but mentions coin**s**, right now one would need to call this multiple times to add multiple gas coins?

crates/iota-transaction-builder/src/builder/command_builder.rs
Thoralf-M: Ah I see

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,684 @@
...
+    /// The sponsor of the transaction. If None, the sender is also the sponsor.
+    sponsor: Option<Address>,
+    /// The expiration of the transaction. The default value of this type is no
+    /// expiration.
+    expiration: TransactionExpiration,
+    named_commands: HashMap<String, Argument>,
+    client: C,
+}
+
+impl<C> TransactionBuilder<C> {
```
Thoralf-M: Why do we need this and `impl TransactionBuilder<Client> {`?

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,684 @@
...
+    /// Set the client to enable automatic object resolution.
+    pub fn with_client(self, client: Client) -> TransactionBuilder<Client> {
```
Thoralf-M: Should this accept Arc<Client> instead so we don't need to clone the whole client?

crates/iota-graphql-client/examples/prepare_send_iota_multi.rs
Thoralf-M: So with your latest changes it's like this now, much better already since it doesn't need to be &'static str:
```Rust
    let recipients = [
        (
            Address::from_str(
                "0x111173a14c3d402c01546c54265c30cc04414c7b7ec1732412bb19066dd49d11",
            )?,
            1_000_000_000u64,
        ),
        (
            Address::from_str(
                "0x2222b466a24399ebcf5ec0f04820812ae20fea1037c736cfec608753aa38b522",
            )?,
            2_000_000_000u64,
        ),
    ];

    let mut builder = TransactionBuilder::new(sender_address).with_client(client.clone());

    // Extract amounts from recipients
    let amounts: Vec<u64> = recipients.iter().map(|(_, amt)| *amt).collect();

    let labels: Vec<String> = (0..recipients.len()).map(|i| format!("split{i}")).collect();

    builder.split_coins(gas_coin, amounts).name(labels.clone());

    // Transfer each split coin to the corresponding recipient
    for (i, (address, _)) in recipients.iter().enumerate() {
        builder.transfer_objects(*address, res(&labels[i]));
    }

    builder.gas(gas_coin).gas_budget(1000000000);

    let txn = builder.finish().await?;
```
Can you also update the example to this?

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,684 @@
...
+            client,
+        }
+    }
+
+    /// Transfer IOTA to a recipient address.
+    pub fn transfer_iota(
+        &mut self,
+        recipient: Address,
+        amount: impl Into<Option<u64>> + Send,
+    ) -> &mut Self {
```
Thoralf-M: Maybe optionally one should be able to provide the input that should be used? In case one doesn't want to transfer from the gas coin
A similar function could also be added for other coins than IOTA (there the input would be required, since gas coin is always IOTA)

DaughterOfMars: I think in that case you could just use a split and transfer quite easily

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: Well the client is already an `Arc` internally, but I do wonder if it's worth using a reference instead or something

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: This one is for any value so applies to both with and without the client

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: Yes, once per coin

Thoralf-M: Why not `gas_coins: impl IntoIterator<Item = ObjectId>`?

crates/iota-transaction-builder/src/builder/mod.rs
Thoralf-M: Ah didn't see the `Arc` since it's "hidden" in `reqwest::Client`

crates/iota-transaction-builder/src/builder/mod.rs
Thoralf-M: But that's the same with the gas, users could also just split from there and transfer
The thing is you just need to be familiar with these concepts and if there is such a method it's easier to use for new developers

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: I just think the builder pattern lends itself better to this method, but I could also add an alternative fn

crates/iota-graphql-client/examples/prepare_send_iota_multi.rs
DaughterOfMars: done

crates/iota-transaction-builder/src/builder/mod.rs
Thoralf-M: I think another fn that accepts multiple object(ids) would be convenient

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -490,6 +518,31 @@ impl<L> TransactionBuilder<Client, L> {
         self
     }
 
+    /// Transfer some coin value to a recipient address.
+    pub fn transfer_value(
+        &mut self,
+        coin: ObjectId,
```
Thoralf-M: Wondering if it should accept more than one coin id here, so first merge them all together and then split the amount

DaughterOfMars: done

crates/iota-transaction-builder/src/builder/named_commands.rs
```diff
@@ -0,0 +1,95 @@
...
+impl<T: NamedCommand> NamedCommands for Vec<T> {
+    fn push_named_commands(self, ptb: &mut TransactionBuildData) {
+        for (i, v) in self.into_iter().enumerate() {
+            let arg = Argument::NestedResult((ptb.commands.len() - 1) as _, i as _);
+            v.push_named_command(arg, ptb);
+        }
+    }
+}
+
+macro_rules! impl_named_command_tuple {
```
Thoralf-M: What is this good for?

DaughterOfMars: You can name the sub-results of a command using tuples :)

Thoralf-M: Ah I was using a vec and wondered why it still worked after removing it :sweat_smile:
Why did you limit it to 6? Anything against supporting at least 10 elements?

DaughterOfMars: No, I just wasn't sure what a good number was, I can add as many as we want

DaughterOfMars: done

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,954 @@
...
+        self.data.named_command(cmd, name);
+    }
+
+    /// Get the value for the given string in the named commands map
+    pub fn get_named_command(&self, name: &str) -> Option<Argument> {
+        self.data.get_named_command(name)
+    }
+
+    /// Transfer IOTA to a recipient address.
+    pub fn transfer_iota(
```
Thoralf-M: Should this also be called `send_iota()` to be aligned with `send_coins()`?

DaughterOfMars: sure

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,955 @@
...
+                tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+            }
+        }
+        Ok(res)
+    }
+}
+
+impl TransactionBuilder<(), MoveCall> {
+    /// Set the call params. Optional.
+    pub fn params(&mut self, params: impl IntoIterator<Item = Argument>) -> &mut Self {
```
Thoralf-M: Rename to arguments()

DaughterOfMars: done

crates/iota-transaction-builder/src/types/move_arg.rs
```diff
@@ -0,0 +1,104 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use iota_types::{Digest, ObjectId};
+
+use crate::types::ArgType;
+
+/// A trait which defines how types are serialized for move calls.
+pub trait MoveArg {
+    /// Get the param type.
+    fn param(&self) -> ArgType;
```
Thoralf-M: Should this also be renamed or not?

DaughterOfMars: sure, did that too

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,955 @@
...
+            data: self.data,
+            client,
+            last_command: self.last_command,
+        }
+    }
+}
+
+impl<C, L> TransactionBuilder<C, L> {
+    /// Apply the given parameters and return the generated arguments
+    pub fn apply_params<P: PTBArguments>(&mut self, param: P) -> Vec<Argument> {
```
Thoralf-M: Should this be also apply_arguments()?

DaughterOfMars: true true

bindings/python/examples/generic_move_function.py
```diff
@@ -12,67 +12,48 @@ async def main():
...
-                    (10_000_000).to_bytes(8, "little", signed=False)
-                )
-            ),
-            builder.input(
-                UnresolvedInput.new_pure(
-                    (20_000_000).to_bytes(8, "little", signed=False)
-                )
-            ),
+            PtbArgument.u64(10_000_000),
+            PtbArgument.u64(20_000_000),
```
Thoralf-M, line 40: So much better now 🚀

crates/iota-graphql-client/examples/dev_inspect.rs
```diff
@@ -22,98 +22,59 @@ async fn main() -> Result<()> {
...
-    builder.set_gas_budget(50000000);
-    builder.set_gas_price(
-        client
-            .reference_gas_price(None)
-            .await?
-            .ok_or_eyre("missing ref gas price")?,
-    );
-    let txn = builder.finish()?;
-
-    let res = client.dry_run_tx(&txn, true).await?;
```
Thoralf-M, line 116: Wondering if we should have this in the example, because one can and sometimes also want to dry run a transaction where one didn't construct the byte, but just got them from someone else

DaughterOfMars: I feel like it's good to have a specific example for that

Thoralf-M: Yes, look at the file name :joy:

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,955 @@
...
+    /// The sponsor of the transaction. If None, the sender is also the sponsor.
+    sponsor: Option<Address>,
+    /// The expiration of the transaction. The default value of this type is no
+    /// expiration.
+    expiration: TransactionExpiration,
+    named_commands: HashMap<String, Argument>,
+}
+
+impl TransactionBuildData {
+    fn set_input(&mut self, kind: InputKind, is_gas: bool) -> Argument {
```
Thoralf-M, line 70: Any reason to not have this public?
I tried to use the clock as argument and did this
```Rust
    let clock = builder.input(iota_types::Input::Shared {
        object_id: "0x6".parse()?,
        initial_shared_version: 1,
        mutable: false,
    });
    builder
        .move_call(Address::TWO, "clock", "timestamp_ms")
        .arguments(clock)
        .name("timestamp_ms");
```
Or maybe even have a method like `.set_shared_input()` so one also doesn't have to care about the is_gas param?
Because now I have to resolve the shared object initial_shared_version myself first or did I miss something?

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,955 @@
...
+        kind: impl Into<PublishType> + Send,
+    ) -> &mut TransactionBuilder<Client, Upgrade> {
+        let module = match kind.into() {
+            PublishType::Path(_path) => todo!("load the package from the path"),
+            PublishType::Compiled(m) => m,
+        };
+        let ticket = self.apply_arguments(upgrade_cap);
+        if ticket.len() != 1 {
+            // TODO: Maybe there's a better way
+            panic!("invalid upgrade cap");
```
Thoralf-M, line 670: Maybe just accept `Argument` instead of `U: PTBArguments`?

crates/iota-transaction-builder/src/error.rs
```diff
@@ -1,12 +1,14 @@
-// Copyright (c) Mysten Labs, Inc.
-// Modifications Copyright (c) 2025 IOTA Stiftung
+// Copyright 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+//! Client errors.
```
Thoralf-M: Transaction builder errors?

crates/iota-transaction-builder/src/types/mod.rs
```diff
@@ -0,0 +1,46 @@
+// Copyright 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Types for use with these tools.
```
Thoralf-M: `tools` sounds a bit off here

crates/iota-graphql-client/examples/dev_inspect.rs
DaughterOfMars: I'm not sure I understand

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: You can just use the id `0x6`

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: Well I want to accept the kind of types you would use for the normal args, I just don't want to accept multiple arguments. I was thinking about creating another trait to handle this but I figured it can wait

crates/iota-transaction-builder/src/types/mod.rs
DaughterOfMars: yeah. this and the other are remnants from the original implementation

crates/iota-graphql-client/examples/dev_inspect.rs
Thoralf-M: The name of the file is `dev_inspect.rs` I rather expect a more generic dev inspect call `client.dry_run_tx(&txn, true)` in it than the tx builder method

crates/iota-transaction-builder/src/builder/mod.rs
Thoralf-M: How?

Thoralf-M: `.arguments("0x6")` fails with `Error: Failed to send IOTA: ExecutionError: ExecutionError { inner: ExecutionErrorInner { kind: CommandArgumentError { arg_idx: 0, kind: InvalidUsageOfPureArg }, source: Some("Non-primitive argument at index 0. If it is an object, it must be populated by an object"), command: Some(0) } }`
`.arguments(ObjectId::from_str("0x6")?)` fails with `Error: Conversion error due to input issue: object 0x0000000000000000000000000000000000000000000000000000000000000006 was passed as owned or immutable, but is not`

crates/iota-graphql-client/examples/dev_inspect.rs
DaughterOfMars: Why?

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: You need to parse the object ID

crates/iota-graphql-client/examples/dev_inspect.rs
Thoralf-M: As I wrote earlier, you might get the transaction bytes from someone else and don't construct it yourself

crates/iota-transaction-builder/src/builder/mod.rs
Thoralf-M: I did that in the second example? `.arguments(ObjectId::from_str("0x6")?)`

crates/iota-graphql-client/examples/dev_inspect.rs
DaughterOfMars: Yeah but what about dev_inspect implies this at all?

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: Ohh, you also need to wrap it like `Mut(clock)`. Sorry, missed that

Thoralf-M: Doesn't need to be mutable for this call though `Error: Query error: [ErrorObject { code: InvalidParams, message: "Error checking transaction input objects: ImmutableParameterExpected { object_id: 0x0000000000000000000000000000000000000000000000000000000000000006 }", data: None }]`

Thoralf-M: I guess the issue is here, since the clock is not ImmutableOrOwned, but shared
```Rust
    fn push_args(&self, ptb: &mut TransactionBuildData, args: &mut Vec<Argument>) {
        let arg = match self.arg_type() {
            ArgType::Object(id) => ptb.set_input(InputKind::ImmutableOrOwned(id), false),
```

Thoralf-M: So maybe there should be a `Shared()` struct and then a way to set it mut?

DaughterOfMars: Ah, I guess I missed this case, I'll take a look

crates/iota-graphql-client/examples/dev_inspect.rs
Thoralf-M: The difference is that in the example with the transaction builder, it's not visible how to provide own tx bytes

Thoralf-M: We could also show both ways maybe, what do you think about that?
```Rust
    // Alternatively, one can use the client with the transaction bytes
    let tx_bytes = builder.finish().await?;
    let _res = client.dry_run_tx(&tx_bytes, true).await?;
```

DaughterOfMars: That's still not the bytes though. You'd have to show deserializing the bytes and then calling dry_run, which I still think should be a separate example

Thoralf-M: Fair, actually dry_run might be more interesting for that, but it's just a flag change anyways, created an issue https://github.com/iotaledger/iota-rust-sdk/issues/250

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,955 @@
...
+    /// The gas price for the transaction.
+    gas_price: Option<u64>,
+    /// The sender of the transaction.
+    sender: Address,
+    /// The sponsor of the transaction. If None, the sender is also the sponsor.
+    sponsor: Option<Address>,
+    /// The expiration of the transaction. The default value of this type is no
+    /// expiration.
+    expiration: TransactionExpiration,
+    named_commands: HashMap<String, Argument>,
```
Thoralf-M: Can you add a comment what this is about? It's only used to name results and pass them as argument in functions, right? Would `named_results` be a better name then?

crates/iota-transaction-builder/src/error.rs
```diff
@@ -1,12 +1,14 @@
-// Copyright (c) Mysten Labs, Inc.
-// Modifications Copyright (c) 2025 IOTA Stiftung
```
Thoralf-M, line 2: This should stay

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: Yeah that makes sense

crates/iota-transaction-builder/src/error.rs
DaughterOfMars: oops not sure how that happened

Thoralf-M (approved):

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,956 @@
...
+    pub fn apply_arguments<P: PTBArguments>(&mut self, param: P) -> Vec<Argument> {
+        param.args(&mut self.data)
+    }
+
+    fn set_input(&mut self, kind: InputKind, is_gas: bool) -> Argument {
+        self.data.set_input(kind, is_gas)
+    }
+
+    fn reset(&mut self) -> &mut TransactionBuilder<C> {
+        unsafe { core::mem::transmute(self) }
```
Alex6323, line 205: missing unsafe doc?

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,956 @@
...
+    }
+
+    fn state_change<U: Into<Command>>(&mut self, command: U) -> &mut TransactionBuilder<C, U> {
+        self.command(command.into());
+        unsafe { core::mem::transmute(self) }
```
Alex6323, line 212: missing unsafe doc?

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,956 @@
...
+    }
+
+    /// Get the current set gas coins.
+    pub fn get_gas(&self) -> Vec<ObjectId> {
+        self.data.get_gas()
+    }
+
+    /// Set the gas budget. Optional.
+    pub fn gas_budget(&mut self, gas_budget: u64) -> &mut Self {
```
Alex6323, line 221: take `mut self` and return `Self` so we don't have to call `.to_owned()`? That's probably one of the first issues users of this API will run into.

Same for all others that return a `&mut Self`.

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,956 @@
...
+        self.data.gas_budget(gas_budget);
+        self
+    }
+
+    /// Set the gas price. Optional.
+    pub fn gas_price(&mut self, gas_price: u64) -> &mut Self {
```
Alex6323, line 227: I would prefix those methods with `with_`. Makes it a bit easier in the editor to know how to manipulate the builder.

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,956 @@
...
+    }
+
+    /// Set the expiration. Optional.
+    pub fn expiration(&mut self, expiration: u64) -> &mut Self {
+        self.data.expiration(expiration);
+        self
+    }
+
+    /// Make a value available to the transaction as an input.
+    pub fn input(&mut self, input: iota_types::Input) -> Argument {
```
Alex6323, line 245: We can't chain those methods, because we need to return an `Argument`. An alternative could be to pass a mutable `Vec<Argument>`, that gets populated with each call, but haven't really tested the idea, whether it's an improvement in terms of API usabilty, or not.

DaughterOfMars: Yeah these methods aren't really meant to be used unless there is a special reason

Alex6323: You are right. Tried my idea, but it's not making things better actually. One would need a map in order to easily reference an argument by later calls in the chain. Not worth it.

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -0,0 +1,961 @@
...
+    /// Get the value for the given string in the named results map
+    pub fn get_named_result(&self, name: &str) -> Option<Argument> {
+        self.data.get_named_result(name)
+    }
+
+    /// Send IOTA to a recipient address.
+    pub fn send_iota(
+        &mut self,
+        recipient: Address,
+        amount: impl Into<Option<u64>> + Send,
```
Alex6323: nit: the `Send` bound restricts this argument unnecessarily, since it's not spawning anything? Probably left-over from being async?

DaughterOfMars: Yeah, left over, thanks

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: @Thoralf-M what do you think?

Alex6323 (approved):
lgtm

crates/iota-transaction-builder/src/builder/mod.rs
Thoralf-M: I was initially also looking for `with_` so I guess it makes sense

PR #218, opened 2025-09-25
feat: Use `eyre` instead of `anyhow`

### LINKED ISSUE #217: Use eyre instead of anyhow
It has more features

DaughterOfMars:
## Description

Uses the `eyre` crate instead of `anyhow`.

Closes #217

### DIFF
--- a/crates/iota-graphql-client/examples/prepare_send_iota_multi.rs
+++ b/crates/iota-graphql-client/examples/prepare_send_iota_multi.rs
@@ -3,8 +3,8 @@
 
 use std::str::FromStr;
 
-use anyhow::{Context, Result};
 use base64ct::Encoding;
+use eyre::{OptionExt, Result};
 use iota_graphql_client::Client;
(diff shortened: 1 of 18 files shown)


Thoralf-M (approved):

Alex6323 (approved):

crates/iota-graphql-client/examples/prepare_send_iota_multi.rs
```diff
@@ -71,7 +71,7 @@ async fn main() -> Result<()> {
         client
             .reference_gas_price(None)
             .await?
-            .context("missing ref gas price")?,
+            .ok_or_eyre("missing ref gas price")?,
```
Alex6323, line 74: I like `context` much more :(

Thoralf-M: Could actually be used with an additional trait import
```
NOTE: However, to help with porting we do provide a ContextCompat trait which implements context for options which you can import to make existing .context calls compile.
```
https://crates.io/crates/eyre

PR #224, opened 2025-09-26
feat: add sign_send_iota example

### LINKED ISSUE #158: Send IOTA, Sign and send

Thoralf-M:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/158


crates/iota-graphql-client/examples/sign_send_iota.rs
```diff
@@ -0,0 +1,71 @@
...
+    builder.gas_price(
+        client
+            .reference_gas_price(None)
+            .await?
+            .expect("missing ref gas price"),
+    );
+
+    let tx = builder.finish().await?;
+
+    let dry_run_result = client.dry_run_tx(&tx, false).await?;
```
DaughterOfMars, line 33: why not one call

DaughterOfMars: oh nvm lol

bindings/go/examples/sign_send_iota/main.go
```diff
@@ -0,0 +1,98 @@
...
+	sdk "bindings/iota_sdk_ffi"
+)
+
+func main() {
+	// Amount to send in nanos
+	amount := uint64(1000)
+	recipientAddress, err := sdk.AddressFromHex("0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900")
+	if err != nil {
+		log.Fatalf("Failed to parse recipient address: %v", err)
+	}
```
Alex6323, line 18: I dunno but to keep the example less cluttered, I would just assert that this can never fail

Thoralf-M: Assert where?

Alex6323: I am just wondering whether we can remove all those redunandent error handling code in all the examples when we know that the parsing cannot fail, because the focus of those examples is not to show correct error handling IMHO, but to show how to `sign + send iota` in this case. Why have this noise in the example? I think people who are interested in the Go example already know how error handling works.

Thoralf-M: Good point, looks like we can just do `recipientAddress, _ := sdk.AddressFromHex("0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900")` if we don't care about the error
But not sure if we should do that in all the examples, since these are examples people will probably also try to replace the address with their own one and if something goes wrong the error message is not very clear:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4ecf8c]

goroutine 1 [running]:
bindings/iota_sdk_ffi.FfiConverterAddress.Lower({}, 0x0)
        /home/t/Desktop/iota-rust-sdk/bindings/go/iota_sdk_ffi/iota_sdk_ffi.go:6850 +0x2c
bindings/iota_sdk_ffi.(*FaucetClient).RequestAndWait.func3(...)
        /home/t/Desktop/iota-rust-sdk/bindings/go/iota_sdk_ffi/iota_sdk_ffi.go:10995
bindings/iota_sdk_ffi.(*FaucetClient).RequestAndWait(0xc00019a150, 0x0)
        /home/t/Desktop/iota-rust-sdk/bindings/go/iota_sdk_ffi/iota_sdk_ffi.go:10995 +0x99
main.main()
        /home/t/Desktop/iota-rust-sdk/bindings/go/examples/faucet/main.go:18 +0x3f
exit status 2
```
the main.go:18 line is where the address would be used, not where it was defined. So it would be less code, but also less clear if something goes wrong

Alex6323: okay, let's keep it .... those Go errors ... wtf

Alex6323 (approved):

DaughterOfMars (approved):

PR #226, opened 2025-09-29
fix(iota-sdk-types): fix numbers and typo

Alex6323:


crates/iota-sdk-types/src/crypto/zklogin.rs
```diff
@@ -522,7 +522,7 @@ impl ZkLoginPublicIdentifier {
 
 /// A JSON Web Key
 ///
-/// Struct that contains info for a JWK. A list of them for different kids can
+/// Struct that contains info for a JWK. A list of them for different keys can
```
Thoralf-M: This seemed to be actually correct, https://www.googleapis.com/oauth2/v3/certs contains "kid" entries

Alex6323: Oh wow, thanks for checking, never seen the term "kid" used for this, only "child" and "parent". TIL

Thoralf-M: Apparently it's just an abbreviation for "key id", so not related to parent-child relationships

Thoralf-M (approved):

PR #240, opened 2025-09-29
chore: Add `iota-sdk` crate for publishing

### LINKED ISSUE #227: Create iota-sdk wrapper crate
reexport the other crates so we can publish it to crates.io

DaughterOfMars:
## Description

Adds a wrapper crate for the other crates in this repo, which can be published to use the SDK.

Closes #227

### DIFF
--- a/crates/iota-sdk/Cargo.toml
+++ b/crates/iota-sdk/Cargo.toml
@@ -4,7 +4,7 @@ version = "3.0.0"
 authors = ["IOTA Foundation <info@iota.org>"]
 edition = "2021"
 license = "Apache-2.0"
-description = "The official SDK for IOTA Rebased"
+description = "The official SDK for IOTA"
 
 [features]
 default = ["types"]
--- a/crates/iota-sdk/Cargo.toml
+++ b/crates/iota-sdk/Cargo.toml
@@ -7,7 +7,24 @@ license = "Apache-2.0"
 description = "The official SDK for IOTA"
 
 [features]
-default = ["types"]
+default = [
+  "crypto",
+  "ed25519",
+  "secp256r1",
+  "passkey",
+  "secp256k1",
+  "zklogin",
+  "pem",
+  "bls12381",
+  "graphql",
+  "txn-builder",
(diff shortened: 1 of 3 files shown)


crates/iota-sdk/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "iota-sdk"
+version = "3.0.0"
+authors = ["IOTA Foundation <info@iota.org>"]
+edition = "2021"
+license = "Apache-2.0"
+description = "The official SDK for IOTA Rebased"
```
Thoralf-M: I think we don't want to mention "Rebased" anymore, it was just used to make the upgrade more clear, but now it should be just `IOTA`

crates/iota-sdk/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "iota-sdk"
+version = "3.0.0"
+authors = ["IOTA Foundation <info@iota.org>"]
+edition = "2021"
+license = "Apache-2.0"
+description = "The official SDK for IOTA"
+
+[features]
+default = ["types"]
```
Alex6323: How about adding a `full` feature for convenience, so users can explore all the SDK functionality more easily? 

I would even consider making `full` the default, because just having the types is probably not what users will need in the beginning. They at least want to be able to interact with the network right away, and not first have to find out what features to enable to being able to do that.

Thoralf-M: I think I agree,if they care only about a specific thing then they can import that individually

DaughterOfMars: I could just add all of them as default

Alex6323: Yeah, I think that's even better. Let's do that!!

DaughterOfMars: done

Thoralf-M (approved):

Alex6323 (approved):

PR #246, opened 2025-09-30
feat!: add TransactionData enum, rename Transaction to TransactionDataV1

### LINKED ISSUE #14: graphql types: add TransactionData enum, rename Transaction to TransactionDataV1
```Rust
pub struct Event {
    pub package_id: ObjectId,
    pub module: Identifier,
```

```Rust
pub struct Event {
    pub package_id: ObjectID,
    pub transaction_module: Identifier,
```
iota repo has `transaction_` as name, is this fine or do we want to align?

Thoralf-M:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/14


crates/iota-graphql-client/src/lib.rs
```diff
@@ -188,14 +188,18 @@ impl Client {
     // Dry Run API
     // ===========================================================================
 
-    /// Dry run a [`Transaction`] and return the transaction effects and dry run
-    /// error (if any).
+    /// Dry run a [`TransactionData`] and return the transaction effects and dry
```
Alex6323: I would not rename  it like that, because dry-running a "TransactionData" sounds weird. Also the method name still indicates that it's a transaction. Maybe now we should just say: "Dry run a transaction and ..." ?

Alex6323: Can we not just call it `TransactionV1` or smth?

crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -27,6 +27,50 @@ use crate::{
 
 pub mod v1;
 
+#[derive(Clone, uniffi::Enum)]
+pub enum TransactionDataInner {
+    Version1(Arc<TransactionDataV1>),
```
Alex6323: I think I'ld prefer `V1(..)` as the variant name, or did you write it out for a specific reason?

Thoralf-M: I would also prefer that, but the go binding fails then, because there the variant name is added to the base name and then the variant would end up as `TransactionDataV1` which conflicts with the actual type

crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -27,6 +27,50 @@ use crate::{
...
+}
+
+// Wrapper around the enum, because we can't have methods on enums in uniffi
+#[derive(Clone, uniffi::Object)]
+pub struct TransactionData {
```
Alex6323: Can we call this struct `Transaction`? Or is that name still occupied? I'ld rather have users work with something they are probably more familiar with.

Thoralf-M: Yes, I just used `TransactionData` because it's called like that in the monorepo and the graphql spec, but I guess there is actually no good reason for it

Thoralf-M: `TransactionData` is still used in `IntentScope` for example, this might be a bit annoying, but probably still nicer

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/246/commits/fc85bf65b6c36050ee9a614effb5131d89fc46b0

crates/iota-graphql-client/src/lib.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/246/commits/fc85bf65b6c36050ee9a614effb5131d89fc46b0

crates/iota-sdk-types/src/hash.rs
```diff
@@ -365,25 +384,25 @@ mod signing_message {
 
     impl PersonalMessage<'_> {
         pub fn signing_digest(&self) -> SigningDigest {
-            const INTENT: Intent = Intent {
+            let intent = Intent {
```
DaughterOfMars: Why this change?

Alex6323 (approved):

crates/iota-sdk-types/src/hash.rs
Thoralf-M: Did some testing and thought I reverted that already, did it now https://github.com/iotaledger/iota-rust-sdk/pull/246/commits/102b835540303523c6ff4e8739dea30a4ddec7f7

bindings/go/examples/prepare_send_iota/main.go
```diff
@@ -26,11 +26,11 @@ func main() {
 		log.Fatalf("Failed to create transaction: %v", err)
 	}
 
-	txnBytes, err := txn.BcsSerialize()
+	txnBytes, err := txn.AsV1().BcsSerialize()
```
DaughterOfMars: Do we not want an impl on the enum?

Thoralf-M: Sure good idea

crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -27,6 +27,78 @@ use crate::{
 
 pub mod v1;
 
+#[derive(Clone, uniffi::Enum)]
+pub enum TransactionInner {
+    Version1(Arc<TransactionV1>),
+}
+
+// Wrapper around the enum, because we can't have methods on enums in uniffi
```
DaughterOfMars: Can't you just derive `uniffi::Object` on the enum?

Thoralf-M: Yes that works, will change

bindings/go/examples/prepare_send_iota/main.go
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/246/commits/8dfe64c32e64b9e3364054587b16467d211cdc0f

crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -27,22 +27,16 @@ use crate::{
...
-#[derive(Clone, uniffi::Enum)]
-pub enum TransactionInner {
-    Version1(Arc<TransactionV1>),
-}
-
-// Wrapper around the enum, because we can't have methods on enums in uniffi
 #[derive(Clone, uniffi::Object)]
-pub struct Transaction {
-    pub inner: TransactionInner,
+pub enum Transaction {
```
DaughterOfMars: Actually why not wrap the sdk type?

Thoralf-M: Initially wanted to represent the enum on the bindings side, but since that's not possible anyways with the methods I guess you're right and it's better to just wrap it

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/246/commits/371fee572d62934a20d434ce891b18cbad741494

crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -27,6 +27,75 @@ use crate::{
...
+    fn from(value: Transaction) -> Self {
+        value.0
+    }
+}
+
+impl From<&Transaction> for iota_types::Transaction {
+    fn from(value: &Transaction) -> Self {
+        value.clone().into()
+    }
+}
```
DaughterOfMars: I don't really think these are necessary, and can obscure some things like the clone

crates/iota-sdk-ffi/src/crypto/multisig.rs
```diff
@@ -87,7 +87,7 @@ impl MultisigAggregator {
         Self(
             iota_crypto::multisig::MultisigAggregator::new_with_transaction(
                 committee.0.clone(),
-                &transaction.0,
+                &transaction.clone().into(),
```
DaughterOfMars: We should no longer need to clone, right?

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/246/commits/87c46e573a93ab1b52886562999657c6cb9ee656

crates/iota-sdk-ffi/src/types/transaction/mod.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/246/commits/87c46e573a93ab1b52886562999657c6cb9ee656

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -743,25 +745,30 @@ impl<L> TransactionBuilder<Client, L> {
         if self.data.gas_budget.is_none() {
             let res = self
                 .client
-                .dry_run_tx_kind(&txn.kind, true, Default::default())
+                .dry_run_tx_kind(&txn.clone().as_v1().kind, true, Default::default())
```
DaughterOfMars: Clone shouldn't be needed

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -743,25 +745,30 @@ impl<L> TransactionBuilder<Client, L> {
...
+            }
         }
 
         Ok(txn)
     }
 
     /// Dry run the transaction.
     pub async fn dry_run(mut self, skip_checks: bool) -> Result<DryRunResult, Error> {
-        let txn = self.resolve_ptb(false).await?;
+        let txn = self.resolve_ptb(false).await?.as_v1();
```
DaughterOfMars: Can we instead operate on the enum via the methods?

crates/iota-transaction-builder/src/builder/mod.rs
Thoralf-M: It's needed here because we mutate it later

crates/iota-transaction-builder/src/builder/mod.rs
Thoralf-M: We don't have methods for all the fields of the v1 on the enum

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: I pushed a commit :)

Alex6323 (approved):

DaughterOfMars (approved):

PR #256, opened 2025-10-02
feat(transaction-builder): Refactor PTB args

DaughterOfMars:
## Description

I did another pass over the traits and made some refactors:

1. Split `PTBArguments` into two traits which solves the `todo` in the upgrade fn.
2. Split `MoveArg` into `MoveArgCollection` to separate collections and single values (which also avoids weirdness with collections of collections). `MoveArg` is now explicitly a trait for serializing pure BCS bytes.
3. Changed all the traits so that they accept `self` instead of `&self` which reduces the amount of cloning needed under certain circumstances. This did require some additional implementations, notably to support both owned and referenced data there must now be explicit impls for both. 
4. I removed the `ArgType` enum and pushed `ObjectId` handling up to the `PTBArgument` traits.

### DIFF
--- a/crates/iota-graphql-client/examples/prepare_split_coins.rs
+++ b/crates/iota-graphql-client/examples/prepare_split_coins.rs
@@ -23,7 +23,7 @@ async fn main() -> Result<()> {
     builder
         .split_coins(coin, [1000, 2000, 3000])
         .name(("coin1", "coin2", "coin3"))
-        .transfer_objects(sender, [res("coin1"), res("coin2"), res("coin3")])
+        .transfer_objects(sender, (res("coin1"), res("coin2"), res("coin3")))
         .gas(coin)
         .gas_budget(1000000000);
 
--- a/crates/iota-sdk-ffi/src/transaction_builder/ptb_arg.rs
+++ b/crates/iota-sdk-ffi/src/transaction_builder/ptb_arg.rs
@@ -3,120 +3,162 @@
 
 use std::sync::Arc;
 
-use iota_transaction_builder::{Receiving, Shared, SharedMut, res, types::PureBytes};
+use iota_transaction_builder::{
+    PureBytes, Receiving, Shared, SharedMut, builder::ptb_arguments::Res, res,
+};
 
 use crate::types::{address::Address, digest::Digest, object::ObjectId};
 
-type DynPTBArg = Box<dyn iota_transaction_builder::PTBArguments + Send + Sync>;
+#[derive(derive_more::From)]
+enum PTBArg {
+    ObjectId(iota_types::ObjectId),
+    Address(iota_types::Address),
+    Digest(iota_types::Digest),
+    U8(u8),
+    U16(u16),
+    U32(u32),
+    U64(u64),
+    U128(u128),
+    String(String),
+    Vector(Vec<PureBytes>),
+    Res(Res),
+    Shared(Shared<iota_types::ObjectId>),
+    SharedMut(SharedMut<iota_types::ObjectId>),
+    Receiving(Receiving<iota_types::ObjectId>),
+    Gas,
+}
 
 #[derive(uniffi::Object)]
-pub struct PTBArgument(DynPTBArg);
+pub struct PTBArgument(PTBArg);
 
 #[uniffi::export]
 impl PTBArgument {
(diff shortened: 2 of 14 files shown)


crates/iota-sdk-ffi/src/transaction_builder/ptb_arg.rs
```diff
@@ -3,122 +3,165 @@
...
+};
 
 use crate::types::{address::Address, digest::Digest, object::ObjectId};
 
-type DynPTBArg = Box<dyn iota_transaction_builder::PTBArguments + Send + Sync>;
+#[derive(derive_more::From)]
+enum PTBArg {
+    ObjectId(iota_types::ObjectId),
+    Address(iota_types::Address),
+    Digest(iota_types::Digest),
```
Thoralf-M, line 17: For what is this needed?

DaughterOfMars: The trait can no longer be dynamically allocated

Thoralf-M: Ah it's not shown, my question was specific about the `Digest` part, isn't that also `Vector(Vec<PureBytes>)` at the end?

DaughterOfMars: No, the `Vector` variant represents a move type `vector<T>` where `PureBytes` are the items in the vec serialized as BCS

crates/iota-sdk-ffi/src/transaction_builder/ptb_arg.rs
```diff
@@ -3,122 +3,165 @@
...
+    U8(u8),
+    U16(u16),
+    U32(u32),
+    U64(u64),
+    U128(u128),
```
Alex6323, line 22: U256 variant not needed?

DaughterOfMars: I don't know how to make it work

DaughterOfMars: I added it

crates/iota-graphql-client/examples/prepare_split_coins.rs
```diff
@@ -23,7 +23,7 @@ async fn main() -> Result<()> {
     builder
         .split_coins(coin, [1000, 2000, 3000])
         .name(("coin1", "coin2", "coin3"))
-        .transfer_objects(sender, [res("coin1"), res("coin2"), res("coin3")])
+        .transfer_objects(sender, (res("coin1"), res("coin2"), res("coin3")))
```
Thoralf-M, line 26: Why don't support arrays anymore?

DaughterOfMars: Hmm, I guess I'll add it back

Thoralf-M (approved):

crates/iota-transaction-builder/src/types/move_arg.rs
```diff
@@ -1,104 +1,113 @@
...
+    while value >= 0x80 {
+        // Write 7 (lowest) bits of data and set the 8th bit to 1.
+        let byte = (value & 0x7f) as u8;
+        res.push(byte | 0x80);
+        value >>= 7;
     }
+    // Write the remaining bits of data and set the highest bit to 0.
+    res.push(value as u8);
+    res
 }
```
Alex6323, line 113: do we have tests for this? Why do we have to do bit manipulation code ourselves? Isn't there some battle tested crate that does this already? Or was this just moved from somewhere?

crates/iota-sdk-ffi/src/transaction_builder/ptb_arg.rs
Alex6323: damn, gh didn't notify me about your response ... so weird

crates/iota-transaction-builder/src/types/move_arg.rs
DaughterOfMars: It was copied from `bcs`

Alex6323 (approved):

PR #257, opened 2025-10-02
fix(transaction-builder, examples): Fix gas budget calculation

### LINKED ISSUE #253: Tx builder: call dry_run_tx_kind() or dry_run() depending on what user set
Right now in ` pub async fn finish(mut self) -> Result<Transaction, Error> {` we always call dry_run() if no `gas_budget` was set to determine the gas budget.
If a gas coin was set, we should call `dry_run_tx_kind()` instead with the gas coins provided in `TransactionMetadata` (also the other fields if set)

DaughterOfMars:
## Description

Fixes the gas budget calculation by calling `dry_run_tx_kind` instead of `dry_run_tx` and remove arbitrary gas budgets from examples.

Closes #253

### DIFF
--- a/bindings/go/examples/prepare_send_iota_multi/main.go
+++ b/bindings/go/examples/prepare_send_iota_multi/main.go
@@ -15,7 +15,7 @@ func main() {
 
 	sender, _ := sdk.AddressFromHex("0x611830d3641a68f94a690dcc25d1f4b0dac948325ac18f6dd32564371735f32c")
 
-	gasCoinId, _ := sdk.ObjectIdFromHex("0x0b0270ee9d27da0db09651e5f7338dfa32c7ee6441ccefa1f6e305735bcfc7ab")
+	coinId, _ := sdk.ObjectIdFromHex("0x0b0270ee9d27da0db09651e5f7338dfa32c7ee6441ccefa1f6e305735bcfc7ab")
 
 	recipients := []struct {
 		address string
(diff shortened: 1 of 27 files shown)


Thoralf-M (approved):

crates/iota-graphql-client/examples/generic_move_function.rs
```diff
@@ -25,7 +25,8 @@ async fn main() -> Result<()> {
         .move_call(Address::TWO, "vec_map", "from_keys_values")
         .generics::<(Address, u64)>()
         .arguments((vec![address1, address2], vec![10000000u64, 20000000u64]))
-        .gas("0xa1d009e8dafe20b1cba05e08aea488aafae1f89d892c3eaef6c0994e155e441a".parse()?);
+        .gas("0xa1d009e8dafe20b1cba05e08aea488aafae1f89d892c3eaef6c0994e155e441a".parse()?)
+        .gas_budget(1000000000);
```
Thoralf-M: Why don't remove both things instead?

DaughterOfMars: We'd also have to skip checks. I dunno 🤷‍♀️

bindings/go/examples/prepare_send_iota_multi/main.go
```diff
@@ -36,15 +36,13 @@ func main() {
 	}
 
 	// Split the gas coin into multiple coins
```
Thoralf-M:
```suggestion
	// Split a coin into multiple coins
```

Thoralf-M (approved):

Alex6323 (approved):

PR #260, opened 2025-10-06
feat(transaction-builder): Add gas station support

### LINKED ISSUE #216: Transaction builder: support gas station
Similar to https://github.com/iotaledger/product-core/blob/main/product_common/src/transaction/transaction_builder.rs#L609-L630

DaughterOfMars:
## Description

Adds support for executing a transaction using a gas station sponsor. This is accomplished with a new set of builder methods which set the URL pointing to the gas station API and some other config. When this is set, the execution will pass the transaction to this endpoint rather than the graphql client.

Closes #216


bindings/go/examples/gas_station/main.go
```diff
@@ -0,0 +1,51 @@
...
+	builder.MoveCall(
+		package_id,
+		module_name,
+		function_name,
+		[]*sdk.PtbArgument{sdk.PtbArgumentU64(64)},
+		nil,
+		nil,
+	)
+
+	token := "test"
```
Thoralf-M: Please move that next to the gasStationUrl and maybe call it `gasStationAuthToken` so it's clearer

crates/iota-transaction-builder/src/builder/gas_station.rs
```diff
@@ -0,0 +1,445 @@
...
+            .entry(reqwest::header::CONTENT_TYPE)
+            .or_insert_with(|| HeaderValue::from_static("application/json"));
+
+        let version = self.gas_station_version(client).await?;
+        if version < GasStationVersion::MIN {
+            return Err(Error::InvalidGasStationVersion {
+                min_required_version: GasStationVersion::MIN,
+                version,
+            });
+        }
```
Thoralf-M, line 308: Version 0.2 doesn't work or why this check?

DaughterOfMars: It was in the core impl so I just copied it 🤷‍♀️

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -871,23 +888,73 @@ impl<L> TransactionBuilder<Client, L> {
...
+            while retries_left > 0
+                && client
+                    .transaction(txn.digest())
+                    .await
+                    .map_err(Error::Client)?
+                    .is_none()
+            {
+                tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+                retries_left -= 1;
+            }
```
Thoralf-M, line 934: I'm wondering if we should return an error if no retries are left (error should include the res data maybe)
Because if the caller wants to wait for finalization, he probably wants to do something afterwards with the tx and if it doesn't exist yet, because the gas station didn't actually submit it maybe, then this could cause some issues

DaughterOfMars: true, good point

crates/iota-transaction-builder/src/error.rs
```diff
@@ -75,4 +75,6 @@ pub enum Error {
     Client(iota_graphql_client::error::Error),
     #[error("Failed to dry run transaction: {0}")]
     DryRun(String),
+    #[error("Timeout waiting for transaction finalization")]
+    FinalizationTimeout,
```
Thoralf-M: Can you at least include the tx digest here?

DaughterOfMars: done

Thoralf-M (approved):

crates/iota-transaction-builder/src/error.rs
Thoralf-M: Thanks

Alex6323 (approved):

PR #261, opened 2025-10-06
feat(iota-crypto): add to/from_bech32 methods

### LINKED ISSUE #210: Support bech32 encoding for private keys
In the CLI and wallet we display private keys as bech32 encoded string with iotaprivkey as prefix https://github.com/search?q=repo%3Aiotaledger%2Fiota%20bech32&type=code
We should support the same here to encode and decode private keys with this format

Thoralf-M:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/210


crates/iota-crypto/src/ed25519.rs
```diff
@@ -105,6 +110,68 @@ impl Ed25519PrivateKey {
...
     }
 
+    #[cfg(feature = "bech32")]
+    #[cfg_attr(doc_cfg, doc(cfg(feature = "bech32")))]
+    /// Encode this private key as `flag || privkey` in Bech32 starting with
+    /// "iotaprivkey" to a string.
+    pub fn to_bech32(&self) -> Result<String, SignatureError> {
+        use bech32::Hrp;
+
+        const IOTA_PRIV_KEY_PREFIX: &str = "iotaprivkey";
```
DaughterOfMars: Maybe move this out of the fn

Thoralf-M: Sure, https://github.com/iotaledger/iota-rust-sdk/pull/261/commits/8ad446ed5deb107e32f08aa3c9c76f9dbb9a3262

crates/iota-crypto/src/lib.rs
```diff
@@ -154,3 +154,8 @@ impl<T: Verifier<UserSignature>> IotaVerifier for T {
         self.verify(&message, signature)
     }
 }
+
+#[cfg(feature = "bech32")]
+#[cfg_attr(doc_cfg, doc(cfg(feature = "bech32")))]
+/// Bech32 prefix for IOTA private keys
```
DaughterOfMars: swap?

DaughterOfMars: actually I see that the rest of these files are like this 😤 whatever

DaughterOfMars (commented):
It really feels like these could be a trait to reduce the duplicated code

crates/iota-crypto/src/lib.rs
Alex6323: that's ugly indeed

crates/iota-crypto/src/lib.rs
```diff
@@ -154,3 +154,60 @@ impl<T: Verifier<UserSignature>> IotaVerifier for T {
...
+    fn from_flagged_bytes(bytes: &[u8]) -> Result<Self, SignatureError>
+    where
+        Self: Sized;
+
+    /// Encode this instance in Bech32 format with "iotaprivkey" prefix
+    fn to_bech32(&self) -> Result<String, SignatureError> {
+        use bech32::Hrp;
+
+        let hrp = Hrp::parse(IOTA_PRIV_KEY_PREFIX)
+            .map_err(|e| SignatureError::from_source(format!("invalid HRP: {e}")))?;
```
Alex6323: since we control this string, unwrapping should be fine?

crates/iota-crypto/src/lib.rs
```diff
@@ -154,3 +154,60 @@ impl<T: Verifier<UserSignature>> IotaVerifier for T {
...
+
+    /// Decode an instance from Bech32 format with "iotaprivkey" prefix
+    fn from_bech32(value: &str) -> Result<Self, SignatureError>
+    where
+        Self: Sized,
+    {
+        use bech32::Hrp;
+
+        let expected_hrp = Hrp::parse(IOTA_PRIV_KEY_PREFIX)
+            .map_err(|e| SignatureError::from_source(format!("invalid HRP: {e}")))?;
```
Alex6323: just unwrap?

crates/iota-crypto/src/ed25519.rs
```diff
@@ -111,6 +116,40 @@ impl Ed25519PrivateKey {
     }
 }
 
+#[cfg(feature = "bech32")]
+impl crate::Bech32 for Ed25519PrivateKey {
+    fn to_flagged_bytes(&self) -> Vec<u8> {
+        let mut bytes = Vec::new();
+        bytes.push(SignatureScheme::Ed25519.to_u8());
+        bytes.extend_from_slice(&self.to_bytes());
```
DaughterOfMars: We could also have a trait (like `PrivateKeyExt`) which defines the dependencies of this one, like `fn to_bytes` and `fn scheme` so that we can just impl this once for `T: PrivateKeyExt`

crates/iota-crypto/src/secp256k1.rs
```diff
@@ -116,6 +121,40 @@ impl Secp256k1PrivateKey {
     }
 }
 
+#[cfg(feature = "bech32")]
+impl crate::Bech32 for Secp256k1PrivateKey {
+    fn to_flagged_bytes(&self) -> Vec<u8> {
+        let mut bytes = Vec::new();
```
Alex6323: I think the extra byte will cause an extra allocation of the vec to 64. Not that it matters much, but we could prevent that.

crates/iota-crypto/src/simple.rs
```diff
@@ -626,4 +714,88 @@ mod test {
         let from_pem = SimpleVerifyingKey::from_pem(&pem).unwrap();
         assert_eq!(pem, from_pem.to_pem().unwrap());
     }
+
+    #[cfg(feature = "bech32")]
+    #[test]
+    fn test_bech32_encode_decode_ed25519() {
```
Alex6323: I don't really see a difference between the encode/decode and the roundtrip tests. Also, only for ed25519 there's both.

crates/iota-crypto/src/simple.rs
```diff
@@ -626,4 +714,88 @@ mod test {
...
+        use rand::{SeedableRng, rngs::StdRng};
+
+        let keypair: SimpleKeypair = Ed25519PrivateKey::generate(StdRng::from_seed([0; 32])).into();
+        let encoded = keypair.to_bech32().unwrap();
+
+        // Verify the prefix is correct
+        assert!(encoded.starts_with("iotaprivkey1"));
```
Alex6323: what's the `1` at the end? the flag?

Thoralf-M: No, that's the separator for bech32 HRPs

crates/iota-crypto/src/simple.rs
Thoralf-M: Right, removed in https://github.com/iotaledger/iota-rust-sdk/pull/261/commits/b214ff6f0429e26af0d6190d0951b45965e78c0a

crates/iota-crypto/src/secp256k1.rs
Thoralf-M: Changed in https://github.com/iotaledger/iota-rust-sdk/pull/261/commits/b214ff6f0429e26af0d6190d0951b45965e78c0a

crates/iota-crypto/src/ed25519.rs
Thoralf-M: Added in https://github.com/iotaledger/iota-rust-sdk/pull/261/commits/b214ff6f0429e26af0d6190d0951b45965e78c0a

crates/iota-crypto/src/lib.rs
Thoralf-M: Probably yes, but I don't see why that would be better
And potentially if there are changes in the future in the bech32 crate or with the prefix it's better to have the Result already in place

crates/iota-crypto/src/lib.rs
Thoralf-M: Changed the order for the bech32 methods

crates/iota-crypto/src/lib.rs
Alex6323: less code = better :trollface: 

okay, then. Closing

Alex6323 (approved):

crates/iota-crypto/src/lib.rs
```diff
@@ -7,6 +7,32 @@
...
+impl std::fmt::Display for PrivateKeyError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PrivateKeyError::EmptyData(msg) => write!(f, "empty data: {msg}"),
+            PrivateKeyError::InvalidScheme(msg) => write!(f, "invalid signature scheme: {msg}"),
+            PrivateKeyError::Bech32(msg) => write!(f, "bech32 error: {msg}"),
+            PrivateKeyError::Bech32Hrp(msg) => write!(f, "bech32 HRP error: {msg}"),
+        }
+    }
+}
```
DaughterOfMars: might as well use `thiserror` right?

Thoralf-M: Wasn't sure about it because of `strive to have a minimal dependency footprint`

DaughterOfMars: yeah but thiserror is so common and minimal

DaughterOfMars: (and I already am using it in the transaction builder)

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/261/commits/cd80ef1c396617ce9d22e34d5949003c11b2757a

DaughterOfMars (approved):

PR #263, opened 2025-10-07
chore(cleanup): make graphql and faucet client api consistent

Alex6323:
I wanted to make both the graphql client and the faucet client constructor API consistent, so usage feels more similar to the user.


Thoralf-M (approved):

Thoralf-M:
CI fails

DaughterOfMars (approved):

PR #265, opened 2025-10-08
chore(examples): Add `transfer_objects` example

### LINKED ISSUE #236: Tx builder: add send_iota() example

### LINKED ISSUE #237: Tx builder: add transfer_objects() example

DaughterOfMars:
## Description

Add an example which transfers objects and fix the `send_iota` example to actually call the right fn.

Closes #237

### DIFF
--- a/bindings/go/examples/prepare_transfer_objects/main.go
+++ b/bindings/go/examples/prepare_transfer_objects/main.go
@@ -9,16 +9,29 @@ import (
 	sdk "bindings/iota_sdk_ffi"
 )
 
+func objIdFromHex(hex string) *sdk.ObjectId {
+	id, err := sdk.ObjectIdFromHex(hex)
+	if err != nil {
+		log.Fatalf("Failed to parse object ID: %v", err)
+	}
+	return id
+}
+
 func main() {
 	client := sdk.GraphQlClientNewDevnet()
 
 	fromAddress, _ := sdk.AddressFromHex("0x611830d3641a68f94a690dcc25d1f4b0dac948325ac18f6dd32564371735f32c")
 
 	toAddress, _ := sdk.AddressFromHex("0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900")
-	amount := uint64(5000000000)
+
+	objsToTransfer := []*sdk.PtbArgument{
+		sdk.PtbArgumentObjectId(objIdFromHex("0xd04077fe3b6fad13b3d4ed0d535b7ca92afcac8f0f2a0e0925fb9f4f0b30c699")),
+		sdk.PtbArgumentObjectId(objIdFromHex("0x0b0270ee9d27da0db09651e5f7338dfa32c7ee6441ccefa1f6e305735bcfc7ab")),
+		sdk.PtbArgumentObjectId(objIdFromHex("0x8ef4259fa2a3499826fa4b8aebeb1d8e478cf5397d05361c96438940b43d28c9")),
+	}
 
 	builder := sdk.TransactionBuilderInit(fromAddress, client)
-	builder.SendIota(toAddress, &amount)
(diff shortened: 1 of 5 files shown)


bindings/go/examples/prepare_send_iota/main.go
Alex6323, line 1: why do we have the `prepare_` prefix? can't we just drop it?

bindings/go/examples/prepare_transfer_objects/main.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+	return id
+}
+
+func main() {
+	client := sdk.GraphQlClientNewDevnet()
+
+	fromAddress, _ := sdk.AddressFromHex("0x611830d3641a68f94a690dcc25d1f4b0dac948325ac18f6dd32564371735f32c")
+
+	toAddress, _ := sdk.AddressFromHex("0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900")
```
Alex6323, line 25: in my PR I changed those to `sender` and `recipient`. anything that speaks against renaming also in the other examples?

bindings/kotlin/examples/PrepareTransferObjects.kt
```diff
@@ -0,0 +1,57 @@
...
+                                ObjectId.fromHex(
+                                        "0x0b0270ee9d27da0db09651e5f7338dfa32c7ee6441ccefa1f6e305735bcfc7ab"
+                                )
+                        ),
+                        PtbArgument.objectId(
+                                ObjectId.fromHex(
+                                        "0x8ef4259fa2a3499826fa4b8aebeb1d8e478cf5397d05361c96438940b43d28c9"
+                                )
+                        )
+                )
```
Alex6323, line 36: as a bindings user this looks unnecessarily cumbersome to create the correct types before. Could we not have constructors that simply take the hex strings or both (string or Address/ObjectID)? At least here in this example I feel like the user should be able to just pass in strings for the addresses and a list of strings for the object ids to the builder methods. Can we achieve this to make lives for user easier?

bindings/go/examples/prepare_send_iota/main.go
DaughterOfMars: Well we're not actually executing, and we will have tests that do execute so I feel like this is better to differentiate

bindings/go/examples/prepare_transfer_objects/main.go
DaughterOfMars: Not really, it's just semantics

bindings/kotlin/examples/PrepareTransferObjects.kt
DaughterOfMars: Sure!

DaughterOfMars: Can you create an issue though? This isn't really the right PR

Alex6323: I agree! will open the issue

bindings/go/examples/prepare_send_iota/main.go
Alex6323: hmm, okay, makes sense!

bindings/kotlin/examples/PrepareTransferObjects.kt
Alex6323: https://github.com/iotaledger/iota-rust-sdk/issues/279

bindings/go/examples/prepare_transfer_objects/main.go
Alex6323: I don't mind merging your PR as-is. Can rename later in cleanup.

Alex6323 (approved):

Thoralf-M (approved):

PR #270, opened 2025-10-08
feat(transaction-builder): Set gas if none was provided

### LINKED ISSUE #268: Tx builder: automatically add gas coins if none were provided
For example a tx like in https://github.com/iotaledger/iota-rust-sdk/pull/224 should work, right now it errors with 
```
Error: Query error: [Internal error occurred while processing request: Unable to execute transaction: Indexer generic error: `Rest Client Error: 400 Bad Request 'Transaction execution failed due to issues with transaction inputs, please review the errors and try again: Transaction gas payment missing.'`]
``` 

Doesn't need to be complicated, just get gas coins for an address and add them all, no need to check the amounts

DaughterOfMars:
## Description

Sets default gas coins from the sender if none were provided to the builder.

Closes #268

### DIFF
--- a/crates/iota-sdk-types/src/object.rs
+++ b/crates/iota-sdk-types/src/object.rs
@@ -392,6 +392,7 @@ impl Object {
     }
 
     /// Return this object's reference
+    #[cfg(feature = "serde")]
     pub fn object_ref(&self) -> ObjectReference {
         ObjectReference {
             object_id: self.object_id(),
--- a/crates/iota-sdk-types/src/object.rs
+++ b/crates/iota-sdk-types/src/object.rs
@@ -392,7 +392,7 @@ impl Object {
     }
 
     /// Return this object's reference
-    #[cfg(feature = "serde")]
+    #[cfg(all(feature = "hash", feature = "serde"))]
     pub fn object_ref(&self) -> ObjectReference {
         ObjectReference {
             object_id: self.object_id(),
--- a/crates/iota-sdk-types/src/object.rs
+++ b/crates/iota-sdk-types/src/object.rs
@@ -391,6 +391,15 @@ impl Object {
         }
     }
 
+    /// Return this object's reference
+    pub fn object_ref(&self) -> ObjectReference {
+        ObjectReference {
+            object_id: self.object_id(),
+            version: self.version(),
+            digest: self.digest(),
+        }
+    }
+
     /// Return this object's version
     pub fn version(&self) -> Version {
         match &self.data {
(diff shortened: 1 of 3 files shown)


crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -702,6 +702,30 @@ impl<L> TransactionBuilder<Client, L> {
...
+                    ObjectFilter {
+                        type_: Some(
+                            StructTag::coin(TypeTag::Struct(Box::new(StructTag::gas_coin())))
+                                .to_string(),
+                        ),
+                        owner: Some(self.data.sender),
+                        ..Default::default()
+                    },
+                    Default::default(),
+                )
```
Thoralf-M, line 726: Could also just use `.coins(address, None, Default::default())`

DaughterOfMars: Yes, but then I wouldn't get the whole object ref so I'd have to do another call to resolve them

Thoralf-M (approved):

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -702,6 +702,30 @@ impl<L> TransactionBuilder<Client, L> {
...
         let mut input_map = HashMap::new();
+        if !self.data.inputs.values().any(|i| i.is_gas) {
+            for coin in self
+                .client
+                .objects(
+                    ObjectFilter {
+                        type_: Some(
+                            StructTag::coin(TypeTag::Struct(Box::new(StructTag::gas_coin())))
+                                .to_string(),
+                        ),
```
Thoralf-M:
```suggestion
                        type_: Some(StructTag::gas_coin().to_string()),
```
`StructTag::gas_coin()` already has everything, right now it doesn't work

DaughterOfMars: oops

Thoralf-M (approved):

crates/iota-sdk-types/src/object.rs
```diff
@@ -391,6 +391,16 @@ impl Object {
         }
     }
 
+    /// Return this object's reference
+    #[cfg(all(feature = "hash", feature = "serde"))]
```
Alex6323, line 395: do we need to feature gate this?

DaughterOfMars: yes, because the digest generation needs both

Alex6323 (approved):

PR #274, opened 2025-10-09
feat: add tx_command_results example

### LINKED ISSUE #233: Tx builder: use nested result as argument

Thoralf-M:
Added tx_command_results examples for https://github.com/iotaledger/iota-rust-sdk/issues/233


crates/iota-sdk-ffi/src/transaction_builder/ptb_command.rs
```diff
@@ -0,0 +1,211 @@
...
+        dependencies: Vec<Arc<ObjectId>>,
+    },
+}
+
+#[uniffi::export]
+impl PTBCommand {
+    #[uniffi::constructor]
+    pub fn split_coins(coin: Arc<PTBArgument>, amounts: Vec<Arc<PTBArgument>>) -> Self {
+        Self(PTBCommandKind::SplitCoins { coin, amounts })
+    }
```
DaughterOfMars: These probably need a new prefix, right?

crates/iota-sdk/examples/tx_command_results.rs
DaughterOfMars, line 1: Honestly, I don't really like that `named_command` is exposed at all. Why would you do this rather than just calling the builder methods?

Thoralf-M: The builder methods currently doesn't allow to provide results from previous commands
We could change the params to make use of the `PTBArgument` trait, but not sure if that's better, because just providing numbers like this `.split_coins(coin, [1000, 2000, 3000])` wouldn't work anymore, you would have to specify them as u64 `.split_coins(coin, [1000u64, 2000, 3000])` and if you read the fn signature it's not clear at all

Thoralf-M: Would look like this, and we probably should adjust all the builder methods to that then https://github.com/iotaledger/iota-rust-sdk/pull/274/commits/8ec948373b53eb2d91f9946137182a05c1b2b4b1
What do you think?

DaughterOfMars: Okay, I see. This is definitely a bit of a weird case. And using PTBArgument means you can pass a lot of stuff that would be invalid :c Still, I guess the system allows you to make mistakes in any command if you don't know what you're doing, so maybe it's not such a big issue 🤷‍♀️

Thoralf-M: So you prefer the params with the `PTBArgument` trait?

Thoralf-M: Resolved in https://github.com/iotaledger/iota-rust-sdk/pull/276

crates/iota-sdk/examples/tx_command_results.rs
```diff
@@ -0,0 +1,56 @@
...
+        // Assign a name to the result of this command
+        .name("res0");
+    builder
+        .move_call(Address::ONE, "u64", "max")
+        .arguments((1000u64, 2000u64))
+        .name("res1");
+
+    builder
+        // Use the named results of previous commands to use as arguments
+        .split_coins(Argument::Gas, vec![res("res0"), res("res1")])
```
DaughterOfMars:
```suggestion
        .split_coins(Argument::Gas, [res("res0"), res("res1")])
```

crates/iota-sdk/examples/tx_command_results.rs
```diff
@@ -0,0 +1,56 @@
...
+        // For nested results, a tuple or vec can be used to name them
+        .name(vec!["coin0", "coin1"]);
+
+    // Use named results as arguments
+    builder.transfer_objects(sender_address, vec![res("coin0"), res("coin1")]);
```
DaughterOfMars:
```suggestion
    builder.transfer_objects(sender_address, [res("coin0"), res("coin1")]);
```

DaughterOfMars (approved):

crates/iota-sdk/examples/tx_command_results.rs
```diff
@@ -0,0 +1,56 @@
...
+    builder
+        // Use the named results of previous commands to use as arguments
+        .split_coins(Argument::Gas, [res("res0"), res("res1")])
+        // For nested results, a tuple or vec can be used to name them
+        .name(vec!["coin0", "coin1"]);
+
+    // Use named results as arguments
+    builder.transfer_objects(sender_address, [res("coin0"), res("coin1")]);
+
+    let tx = builder.to_owned().finish().await?;
```
thibault-martinez: Do we need this `to_owned()`?

Thoralf-M: No, only if we call all the builder methods without assigning the builder to a variable

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/274/commits/783acbadf4a63539ed9840f59f6d8475e165e9e6

bindings/go/examples/tx_command_results/main.go
```diff
@@ -0,0 +1,76 @@
...
+
+	txnBytes, err := txn.BcsSerialize()
+	if err != nil {
+		log.Fatalf("Failed to serialize transaction: %v", err)
+	}
+	log.Printf("Signing Digest: %v", sdk.HexEncode(txn.SigningDigest()))
+	log.Printf("Txn Bytes: %v", sdk.Base64Encode(txnBytes))
+
+	skipChecks := bool(false)
+	res, err := client.DryRunTx(txn, &skipChecks)
```
thibault-martinez: pass false directly

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/274/commits/f7d3ca04c3d3f20e1246b65f25c8903a2617ff9f

thibault-martinez (approved):

PR #275, opened 2025-10-09
chore(graphql): Make stream methods sync and add helpers

DaughterOfMars:
## Description

Makes the stream methods in the graphQL client sync as they did not need to be async.

Cleans up the coins methods and adds gas_coins methods.

Adds an additional helper fn `iota` for struct tag.


crates/iota-graphql-client/src/lib.rs
```diff
@@ -625,7 +636,14 @@ impl Client {
...
                         coin_type
                             .into()
-                            .unwrap_or_else(|| "0x2::coin::Coin".to_owned()),
+                            .map(StructTag::coin)
+                            .unwrap_or_else(|| StructTag {
+                                address: Address::TWO,
+                                module: IdentifierRef::const_new("coin").into(),
+                                name: IdentifierRef::const_new("Coin").into(),
+                                type_params: Default::default(),
+                            })
```
Thoralf-M, line 645: StructTag::coin() ?

Thoralf-M: Wait, can this even be valid without type_params?

DaughterOfMars: Apparently that's what graphql wants

Thoralf-M (approved):

thibault-martinez (approved):

crates/iota-graphql-client/src/lib.rs
```diff
@@ -644,6 +662,16 @@ impl Client {
         ))
     }
 
+    /// Get the list of gas coins for the specified address.
+    pub async fn gas_coins(
+        &self,
+        owner: Address,
+        pagination_filter: PaginationFilter,
+    ) -> Result<Page<Coin>> {
+        self.coins(owner, StructTag::iota(), pagination_filter)
```
thibault-martinez, line 671: Just curious, why not call `StructTag::gas_coins()`, same but would feel more consistent

thibault-martinez: Ah, read too fast, ignore me

PR #276, opened 2025-10-09
feat(transaction-builder): Use PTBArgument trait for all builder calls

DaughterOfMars:
## Description

Many builder calls used `ObjectId` or `ObjectReference` for params, but this did not allow passing the result of previous commands, which is a normal use-case. This PR adds support for that to all calls and also fills in some missing impls for `ObjectReference`.


bindings/python/examples/prepare_send_coins.py
```diff
@@ -19,20 +19,16 @@ async def main():
 
         # This is a coin of type
         # 0x3358bea865960fea2a1c6844b6fc365f662463dd1821f619838eb2e606a53b6a::cert::CERT
-        coin_id = ObjectId.from_hex(
+        coin_id = PtbArgument.object_id_from_hex(
```
Thoralf-M, line 22: I guess usually one would get the object id from the client as `ObjectId` class already, we should probably add a `.to_ptb_argument()` method on `ObjectId` then

crates/iota-transaction-builder/src/builder/ptb_arguments.rs
```diff
@@ -34,15 +35,37 @@ impl PTBArgument for ObjectId {
...
 }
 
-/// A trait which defines arguments for a
+/// A trait which defines a list of arguments for a
 /// [`TransactionBuilder`](crate::TransactionBuilder).
-pub trait PTBArguments {
+#[diagnostic::on_unimplemented(
+    message = "Provided value is not a valid list of arguments.",
+    note = "Expected a tuple, vector, array, or slice of types that implement `PTBArgument`."
+)]
```
Thoralf-M, line 73: Good one, didn't know that exists

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -443,80 +423,84 @@ impl<L> TransactionBuilder<(), L> {
...
+    /// Add multiple gas coins that will be consumed. Optional.
+    pub fn gas_coins(&mut self, obj_refs: impl IntoIterator<Item = ObjectReference>) -> &mut Self {
+        for obj_ref in obj_refs {
+            self.gas(obj_ref);
+        }
+        self
+    }
 
     /// Convert this builder into a transaction.
     pub fn finish(mut self) -> Result<Transaction, Error> {
```
Thoralf-M, line 506: Would add this to the finish gas coin selection
```Rust
                // Skip if the coin is already used as an input
                if self.data.inputs.values().any(
                    |i| matches!(i.kind, InputKind::ImmutableOrOwned(id) if id == coin.object_id()),
                ) {
                    continue;
                }
```

bindings/python/examples/prepare_send_coins.py
DaughterOfMars: But in that case you can also just wrap it the normal way. Not sure how many convenience methods we need for that

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: I think I solved it

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -866,6 +737,7 @@ impl<L> TransactionBuilder<Client, L> {
     pub async fn finish(mut self) -> Result<Transaction, Error> {
         let mut txn = self.resolve_ptb(true).await?;
         if self.data.gas_budget.is_none() {
+            println!("{txn:#?}");
```
Thoralf-M: :eye:

DaughterOfMars: oops lol

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -581,7 +581,25 @@ impl<L> TransactionBuilder<Client, L> {
...
+                    }
+                    _ => &[],
+                } {
+                    if let Argument::Input(idx) = arg {
+                        if let Some(obj_id) = self.data.inputs[idx].object_id() {
+                            unusable_object_ids.insert(*obj_id);
+                        }
+                    }
+                }
+            }
```
Thoralf-M, line 606: Please add a comment for this part like `Collect coin inputs that are not allowed to be replaced by the gas argument`

bindings/python/examples/prepare_send_coins.py
Thoralf-M: Hm right, the normal way might be sufficient then

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: sure, done

crates/iota-sdk-ffi/src/transaction_builder/ptb_arg.rs
```diff
@@ -22,154 +25,146 @@ enum PTBArg {
     U128(u128),
     U256(U256),
     String(String),
-    Vector(Vec<PureBytes>),
```
Thoralf-M, line 25: So how does one provide this now?

DaughterOfMars: I figured it's just a better experience using `make_move_vec` anyway, so why have this janky one

Thoralf-M: `make_move_vec` is an own PTB command, probably costs more gas and you also need to provide the result as argument instead of the value directly

DaughterOfMars: Okay, I'll think of another solution.

DaughterOfMars: I came up with something that is better I think

crates/iota-sdk-ffi/src/transaction_builder/ptb_arg.rs
```diff
@@ -93,54 +263,129 @@ impl PTBArgument {
...
+    }
+
+    #[uniffi::constructor]
+    pub fn u128_vec(values: Vec<String>) -> Result<Self> {
+        Ok(Self::Move(MoveArg::u128_vec(values)?))
+    }
+
+    #[uniffi::constructor]
+    pub fn u256_vec(values: Vec<String>) -> Result<Self> {
+        Ok(Self::Move(MoveArg::u256_vec(values)?))
```
Thoralf-M, line 393: Do we miss a `string_vec()` or `utf_8_vec()`?

DaughterOfMars: added it

Thoralf-M (approved):

Alex6323 (approved):

PR #277, opened 2025-10-09
chore(*): re-add commented proptest + consistent naming

### LINKED ISSUE #30: Examine proptest TODO
worth an issue?

_Originally posted by @Alex6323 in https://github.com/iotaledger/iota-rust-sdk/pull/28#discussion_r2121463200_

Alex6323:
Closes #30

### DIFF
--- a/crates/iota-crypto/src/bls12381.rs
+++ b/crates/iota-crypto/src/bls12381.rs
@@ -134,7 +134,7 @@ impl Verifier<Bls12381Signature> for Bls12381VerifyingKey {
 }
 
 #[cfg(test)]
-mod test {
+mod tests {
     use test_strategy::proptest;
 
     use super::*;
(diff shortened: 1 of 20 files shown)


DaughterOfMars (commented):
I think `mod test` is more idiomatic but 🤷‍♀️

Alex6323:
> I think `mod test` is more idiomatic but 🤷‍♀️

oh, and I always thought `mod tests` is the idiomatic way, which is why I made the change. If there's debate around that, then I don't mind undoing it. I checked the standard lib and that mostly uses `mod tests` though :man_shrugging:

Thoralf-M (approved):

Thoralf-M:
https://doc.rust-lang.org/book/ch11-03-test-organization.html and https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html use `mod tests`

DaughterOfMars (approved):

Issue #231, opened 2025-09-29
Expose primitive type + `String` bcs conversion to bindings


Thoralf-M:
Add `to/from_bcs` function for every primitive type + String

Edited by @Alex6323 to make the scope of the issue clearer.

DaughterOfMars:
Would also be good to have both serialize and deserialize fns

Alex6323:
I'll take this one @Thoralf-M

PR #282, opened 2025-10-13
feat(iota-sdk-ffi): Add `from/to_bcs` methods to all FFI exposed core types

### LINKED ISSUE #333: Expose IOTA core types bcs conversion to bindings

Alex6323:
Closes #333 

This PR here adds BCS conversion to all FFI exposed core types that have a BCS definition and exposes them. Since this is very repetitive I implemented this using a declarative macro, which also makes it easier in the future to add this to other types if that may be necessary.


crates/iota-sdk-ffi/src/types/checkpoint.rs
```diff
@@ -272,6 +276,18 @@ impl CheckpointCommitment {
     }
 }
 
+/// Data, which when included in a [`CheckpointSummary`], signals the end of an
```
DaughterOfMars:
```suggestion
/// Data which, when included in a [`CheckpointSummary`], signals the end of an
```

crates/iota-sdk-ffi/src/lib.rs
```diff
@@ -137,3 +137,43 @@ pub fn hex_encode(input: &[u8]) -> String {
 pub fn hex_decode(input: String) -> crate::error::Result<Vec<u8>> {
     Ok(hex::decode(input)?)
 }
+
+#[macro_export]
+macro_rules! export_record_enum_bcs_conversion {
```
DaughterOfMars: This makes me sad. Users of the lib will obviously not understand why some types have methods and some don't, and this will be confusing. But I don't know what to do other than just make everything an object, and that sucks. 😤

Alex6323: that's so true, I was dying inside

Alex6323: okay, so changed everything to free fns, for objects and records/enums, so for every type it's now the same consistent API.

DaughterOfMars (commented):
So we're not doing the primitive types in this PR?

DaughterOfMars:
Need to update the bindings

Alex6323:
> So we're not doing the primitive types in this PR?

I guess not, unless we want to expand issue #231 to mean all types (Thoralf said intended only the primitives wit that isse ... I'll edit it to make this clear). I can however create a separate issue for the iota types and link that in the description.

Alex6323:
> Need to update the bindings

with empty lines (kotlin) :sob:

Alex6323:
> > So we're not doing the primitive types in this PR?
> 
> I guess not, unless we want to expand issue #231 to mean all types (Thoralf said intended only the primitives wit that isse ... I'll edit it to make this clear). I can however create a separate issue for the iota types and link that in the description.

I created a separate issue #333  and updated #231 with the original intent as clarified in Slack with @Thoralf-M

DaughterOfMars (approved):

PR #283, opened 2025-10-13
chore(cleanup): use better names for default address constants

Alex6323:
Never liked the ONE, TWO, THREE cryptic constants. Instead I propose to give proper names as to what they represent. Makes the code more readable I think.


DaughterOfMars (commented):
Very much agree with this change. While you're at it, can you add these to the FFI?

bindings/go/examples/dev_inspect/main.go
```diff
@@ -13,13 +13,13 @@ import (
 func main() {
 	client := sdk.GraphQlClientNewDevnet()
 
-	sender, _ := sdk.AddressFromHex("0x0")
+	sender := sdk.AddressZero()
```
Thoralf-M, line 16: No need to revert here now, but I think for sender address I would prefer `sdk.AddressFromHex("0x0")` because that's easier to change (just replace the address string instead of finding the required other function)

bindings/go/examples/unstake/main.go
```diff
@@ -22,9 +22,9 @@ func main() {
 	}
 	stakedIota := stakedIotas.Data[0]
 
-	iotaSystemAddress, _ := sdk.AddressFromHex("0x3")
+	iotaSystemAddress := sdk.AddressSystem()
 
-	iotaSystemId, _ := sdk.ObjectIdFromHex("0x5")
+	iotaSystemId := sdk.ObjectIdSystem()
```
Thoralf-M, line 27: Does the FFI generates this different in go? Since we have `ObjectId.system()` in the other languages, not a single function

crates/iota-sdk-ffi/src/types/address.rs
```diff
@@ -85,3 +85,19 @@ impl Address {
...
     }
 }
+
+macro_rules! named_address {
+    ($($constant:ident),+ $(,)?) => {
+        paste::paste! {
+            #[uniffi::export]
+            impl Address {$(
+                #[uniffi::constructor]
+                pub fn [< $constant:lower >]() -> Self {
```
DaughterOfMars:
```suggestion
                pub const fn [< $constant:lower >]() -> Self {
```

crates/iota-sdk-ffi/src/types/object.rs
```diff
@@ -84,6 +84,22 @@ impl ObjectId {
     }
 }
 
+macro_rules! named_object_id {
+    ($($constant:ident),+ $(,)?) => {
+        paste::paste! {
+            #[uniffi::export]
+            impl ObjectId {$(
+                #[uniffi::constructor]
+                pub fn [< $constant:lower >]() -> Self {
```
DaughterOfMars:
```suggestion
                pub const fn [< $constant:lower >]() -> Self {
```

crates/iota-sdk-ffi/Cargo.toml
```diff
@@ -18,6 +18,7 @@ base64ct = { version = "1.6.0", features = ["alloc", "std"] }
 bcs = "0.1.6"
 derive_more = { version = "2.0", features = ["from", "deref", "display"] }
 hex = "0.4.3"
+paste = "1.0"
```
DaughterOfMars, line 21: Is it really worth adding a dependency for a small macro like this?

bindings/go/examples/unstake/main.go
DaughterOfMars: Yeah, go is weird

crates/iota-sdk-ffi/Cargo.toml
Alex6323: does it really matter? We have it in the dependency tree already, no?

bindings/go/examples/dev_inspect/main.go
thibault-martinez, line 16: I would have preferred this with Address as suffix but I guess there's not much we can do about it

Thoralf-M (approved):

bindings/go/examples/dev_inspect/main.go
Alex6323: I don't think we can do much about it. This is what `uniffi-bindgen-go` does if you impl a constructor method on a type. So we'ld have to create a `Zero` type and then impl an `address` method on it to have it like you prefer, but I don't think we want that.

crates/iota-sdk-ffi/Cargo.toml
DaughterOfMars: Yeah, I suppose so

DaughterOfMars (approved):

PR #284, opened 2025-10-13
fix(*): replace deprecated `.as_slice` call 

Alex6323:


Thoralf-M (approved):

crates/iota-sdk-types/src/hash.rs
```diff
@@ -29,7 +29,7 @@ impl Hasher {
         let mut buf = [0; Digest::LENGTH];
         let result = self.0.finalize();
 
-        buf.copy_from_slice(result.as_slice());
+        buf.copy_from_slice(result.as_ref());
```
DaughterOfMars:
```suggestion
        buf.copy_from_slice(&result);
```

DaughterOfMars (approved):

thibault-martinez:
What's up with the go bindings?

DaughterOfMars:
> What's up with the go bindings?

I think it is something I fixed in my other PR

PR #301, opened 2025-10-13
chore(examples): Add offline transfer example

### LINKED ISSUE #239: Tx builder: add example to build tx without client (offline)
Show that it's possible to construct a transaction offline (client is obviously needed to fetch the inputs, but it should be possible to construct a transaction without providing a client as parameter to the tx builder)

DaughterOfMars:
## Description

Add an example which uses the transaction builder without the client attached.

Closes #239


Thoralf-M (changes requested):
Should also be added for the bindings

DaughterOfMars:
> Should also be added for the bindings

So, I can't actually do that yet because the FFI version of the builder only supports building with the client

Thoralf-M:
> > Should also be added for the bindings
> 
> So, I can't actually do that yet because the FFI version of the builder only supports building with the client

Then please open an issue so we expose that there and then add the examples

DaughterOfMars:
> > > Should also be added for the bindings
> > 
> > 
> > So, I can't actually do that yet because the FFI version of the builder only supports building with the client
> 
> Then please open an issue so we expose that there and then add the examples

https://github.com/iotaledger/iota-rust-sdk/issues/304

Thoralf-M (approved):

Alex6323 (approved):

PR #303, opened 2025-10-13
feat(types): add Transaction::{to/from bcs, base_64}

### LINKED ISSUE #250: Add dry_run_tx_bytes example
Decode tx from base64 encoded bytes and dry run them

Thoralf-M:
Renamed bcs_serialize() to to_bcs() because I think that's clearer to understand for most users.
Also added from_bcs(), to_base64() and from_base64() to make these conversions easier and exposed them all to the bindings.
Added a dry_run_bytes example which makes use of `from_base64()`.
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/250


bindings/kotlin/examples/PrepareSendIotaMulti.kt
```diff
@@ -6,67 +6,70 @@ import kotlinx.coroutines.runBlocking
...
-        val client = GraphQlClient.newDevnet()
-        val sender =
-                Address.fromHex(
-                        "0x611830d3641a68f94a690dcc25d1f4b0dac948325ac18f6dd32564371735f32c"
-                )
-        val coinId =
-                ObjectId.fromHex(
-                        "0x0b0270ee9d27da0db09651e5f7338dfa32c7ee6441ccefa1f6e305735bcfc7ab"
-                )
+        try {
```
DaughterOfMars: What happened with the spacing

Thoralf-M: Oh no, already noticed a few times that it switches to 8 spaces, not sure why

Thoralf-M: Fixed https://github.com/iotaledger/iota-rust-sdk/pull/303/commits/576b13f22e4630af2524bbc2347de20240503e31

DaughterOfMars: Vscode sometimes gets confused. But it looks like the file is still not properly formatted because there are a lot of changes

Thoralf-M: Not in this file, only in gas sponsor that was formatted with 8 spaces instead of 4, I formatted it with 4 now

DaughterOfMars: ah, that must have been my workspace's fault then haha

DaughterOfMars (approved):

crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -82,8 +82,26 @@ impl Transaction {
         self.0.signing_digest().to_vec()
     }
 
-    pub fn bcs_serialize(&self) -> Result<Vec<u8>> {
-        Ok(bcs::to_bytes(&self.0)?)
+    /// Serialize the transaction as a `Vec<u8>` of BCS bytes.
+    pub fn to_bytes(&self) -> Result<Vec<u8>> {
```
Alex6323: if feel like this naming might conflict with non-encoded bytes. There is (raw) bytes and theres BCS encoded bytes. For future proofing I am wondering if it's better to call this `to_bcs`/`from_bcs`. In my opinion it's clearer because one immediately knows what kind of bytes to pass in or how to further process received bytes. Also would be consistent with other signatures that mention the encoding, see `to_base64`. (NOTE that base64 doesn't mean String necessarily, it just happens to map easily to characters), so it could also return bytes, i.e. a `Vec<u8>`.

Thoralf-M: But is there any use case for bytes that aren't BCS encoded for a transaction? I haven't looked into gRPC what encoding is used there and if that might be relevant, but apart from that I don't have an idea where this could even just theoretically be interesting. I went for this name because `BCS` is not common, but everyone understands what `bytes` are. I would rather use `from/to_bcs_bytes()` at least

Thoralf-M: Just checked what Sui does with gRPC and they also use BCS for the transactions there https://github.com/MystenLabs/sui-rust-sdk/blob/master/crates/sui-rpc/src/proto/sui/rpc/v2/transaction.rs

Alex6323: Even if there's no other use case (right now or forever), I still think it's good practice to mention the encoding as we do in general { to/from_hex, to/from_base64, to/from_json }. Like whenever you need an encoder/decoder to process the bytes or string I would as a personal preference mention that encoding in the method name. I wouldn't call it `from/to_bcs_bytes` for the same reason we don't call it `from/to_hex_string` or `from/to_base64_string`. It's kind of redundant and you can quickly look at the return type to see that it's bytes/string. It's self-documenting. The encoding however is not. But I don't know how others feel about it. :shrug:

Thoralf-M: Renamed in https://github.com/iotaledger/iota-rust-sdk/pull/303/commits/d8ee1e682880aed30fbb69f277594a10cb834fd6

Alex6323 (approved):

bindings/go/examples/dry_run_bytes/main.go
DaughterOfMars, line 1: Where did this come from?

crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -133,8 +156,26 @@ impl TransactionV1 {
...
+    }
+
+    /// Serialize the transaction as a base64-encoded string.
+    pub fn to_base64(&self) -> Result<String> {
+        Ok(self.0.to_base64()?)
+    }
+
+    /// Deserialize a transaction from a `Vec<u8>` of BCS bytes.
+    #[uniffi::constructor]
+    pub fn new_from_bcs(bytes: Vec<u8>) -> Result<Self> {
```
DaughterOfMars: I don't think we need the new prefix here

bindings/go/examples/dry_run_bytes/main.go
Thoralf-M: I added this in this PR, because before there was no way to decode a tx from base64
I can split in a separated PR if you want

crates/iota-sdk-ffi/src/types/transaction/mod.rs
Thoralf-M: It's technically not needed, but didn't we want to have this for all methods so they're easier to find?
On the client we also have `new_devnet()`

DaughterOfMars: Yeah but I think the `from_` already does that

Thoralf-M: Fine by me, removed in https://github.com/iotaledger/iota-rust-sdk/pull/303/commits/723dbf9526a4b89e63043ec738c8d8540ddefc14

bindings/go/examples/dry_run_bytes/main.go
DaughterOfMars: nah, just checking to make sure it wasn't on accident

DaughterOfMars (approved):

PR #305, opened 2025-10-13
chore: move examples to `iota-sdk` wrapper crate

thibault-martinez:


DaughterOfMars (approved):

crates/iota-sdk/build.rs
Thoralf-M, line 1: Does this mean that other crates that use the iota_graphql_client_build as dependency also have to add this to their build.rs?

thibault-martinez: Just having `iota_graphql_client_build` doesn't force you to add it to the `build.rs` but you can't use things like `#[cynic(schema = "rpc", graphql_type = "Epoch")]` without it.

Thoralf-M: Ah and that can't be put in the example?

crates/iota-sdk/examples/dev_inspect.rs
Thoralf-M, line 1: If we're already on it, how about moving all the transaction builder examples in a subfolder?

crates/iota-sdk/build.rs
thibault-martinez: Apparently has to be done in a build.rs.

```
error: Could not find a schema with this name.  Have you registered it in build.rs?  See the cynic documentation on regsistering schemas if you need help.
```

crates/iota-sdk/examples/dev_inspect.rs
thibault-martinez: We would need to start registering them manually in the Cargo.toml then, maybe a bit annoying

Thoralf-M: Ok maybe not now, but I think we should do it anyways once we're basically done with the examples, because from the names it's sometimes not so easy to find the example that was just a query or building a tx

Thoralf-M (approved):

crates/iota-sdk/examples/dev_inspect.rs
thibault-martinez: Fine by me

PR #307, opened 2025-10-14
chore(transaction-builder): Fix leftover `named-result` reference

DaughterOfMars:
## Description

I noticed that this method wasn't properly renamed.

### DIFF
--- a/crates/iota-transaction-builder/src/builder/mod.rs
+++ b/crates/iota-transaction-builder/src/builder/mod.rs
@@ -176,7 +176,7 @@ impl TransactionBuildData {
         name.push_named_results(self);
     }
 
-    /// Get the value for the given string in the named results map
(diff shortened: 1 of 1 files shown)


crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -306,7 +306,7 @@ impl<C, L> TransactionBuilder<C, L> {
 
     /// Get the value for the given string in the named results map
     pub fn get_named_result(&self, name: &str) -> Option<Argument> {
```
Thoralf-M, line 308: But we went from command -> result, no? Because the name is used to reference a result of a command and not the command itself

DaughterOfMars: Hm, whoops

PR #308, opened 2025-10-14
feat(iota-transaction-builder): improve send_iota() docs

### LINKED ISSUE #297: Tx builder: document TransactionBuilder::send_iota()

Thoralf-M:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/297


crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -393,6 +393,36 @@ impl<C, L> TransactionBuilder<C, L> {
...
+    ///     let client = Client::new_devnet();
+    ///     let from_address = Address::from_hex(
+    ///         "0x611830d3641a68f94a690dcc25d1f4b0dac948325ac18f6dd32564371735f32c",
+    ///     )?;
+    ///     let to_address = Address::from_hex(
+    ///         "0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900",
+    ///     )?;
+    ///
+    ///     let mut builder = TransactionBuilder::new(from_address).with_client(client);
+    ///     builder.send_iota::<u64>(to_address, 5000000000u64);
```
thibault-martinez: Is the explicit generic argument needed?

crates/iota-sdk-ffi/src/transaction_builder/mod.rs
```diff
@@ -165,6 +165,12 @@ impl TransactionBuilder {
     }
 
     /// Send IOTA to a recipient address.
+    ///
+    /// The `amount` parameter specifies the quantity in NANOS, where 1 IOTA
+    /// equals 1_000_000_000 NANOS.
+    /// If `amount` is provided, that amount is split from the gas coin and
+    /// sent.
+    /// If `amount` is `None`, the entire gas coin is transferred.
```
Alex6323: That feels dangerous, but probably you'll tell that it isnt bc of some good reason, right? So people forgetting in the bindings, where you can just leave it out, send the entire coin?

DaughterOfMars: Hmm, you have a point, since it can be the gas coin. With `send_coins` it is fine because you provide the coin, but in this case maybe we should make this non-optional.

DaughterOfMars: https://github.com/iotaledger/iota-rust-sdk/pull/338

PR #312, opened 2025-10-14
chore(generate_ed25519_address): add `*PublicKey::to_flagged_bytes` and harmonize examples

### LINKED ISSUE #310: PublicKey: add method to get a public key with scheme byte prepended
Something like `to_flagged_bytes()` we have in `PrivateKeyExt`
Should be also exposed to bindings and the generate_ed25519_address examples should use it

thibault-martinez:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/310


DaughterOfMars:
I'd like to merge #303 first

crates/iota-sdk-types/src/crypto/secp256r1.rs
```diff
@@ -66,6 +66,15 @@ impl Secp256r1PublicKey {
     pub fn from_bytes<T: AsRef<[u8]>>(bytes: T) -> Result<Self, std::array::TryFromSliceError> {
         <[u8; Self::LENGTH]>::try_from(bytes.as_ref()).map(Self)
     }
+
+    /// Returns the bytes with signature scheme flag prepended
+    pub fn to_flagged_bytes(&self) -> Vec<u8> {
```
Thoralf-M: Not sure if worth it, but could be a default trait fn to not repeat it

Thoralf-M (approved):

crates/iota-sdk-types/src/crypto/secp256r1.rs
Alex6323: I agree with Thoralf. Extension trait would be especially good for crypto stuff so you don't have to check in more than one place if there's a bug.

thibault-martinez: @DaughterOfMars opinion on this?

DaughterOfMars: Yeah I think it should be a trait too

thibault-martinez: Done, but please don't pay too much attention of where the trait was added, there is not better place atm. I think we need to sort out the clashes between iota-types/crypto and iota-crypto before.

Alex6323 (approved):

crates/iota-sdk-types/src/crypto/mod.rs
```diff
@@ -151,3 +151,23 @@ impl_base64_helper!(Base64Array34, Base64Display34, Base64FromStr34, test34, 34)
 impl_base64_helper!(Base64Array48, Base64Display48, Base64FromStr48, test48, 48);
 impl_base64_helper!(Base64Array64, Base64Display64, Base64FromStr64, test64, 64);
 impl_base64_helper!(Base64Array96, Base64Display96, Base64FromStr96, test96, 96);
+
+pub trait PublicKeyExt: Sized {
+    // Returns the public key as bytes.
+    fn as_bytes(&self) -> &[u8];
+
+    /// Tries to create a PublicKey from bytes.
+    fn from_bytes<T: AsRef<[u8]>>(bytes: T) -> Result<Self, std::array::TryFromSliceError>;
```
DaughterOfMars: Probably want a `type FromBytesErr` and return `Result<Self, Self::FromBytesErr>`

thibault-martinez: Why?

thibault-martinez: I mean, I understand why but I doubt we'll have public keys where the bytes are not parsed this way. Added it anyway.

DaughterOfMars (approved):

PR #315, opened 2025-10-15
feat(transaction-builder): Make client in builder extensible and passable by ref

DaughterOfMars:
## Description

This PR adds the `ClientMethods` trait, which makes it possible to implement new clients that can be used by the transaction builder. Additionally, these can now be passed in as a reference to the builder.

NOTE: In order to make this not depend on the graphql typess, specifically those for the `DryRunResult`, I moved them into the types crate next to the `TransactionEffects`.


crates/iota-sdk-ffi/src/types/transaction/mod.rs
```diff
@@ -1464,6 +1464,184 @@ impl TransactionEffects {
...
+            return_values: value.return_values.into_iter().map(Into::into).collect(),
+        }
+    }
+}
+
+/// The result of a simulation (dry run), which includes the effects of the
+/// transaction, any errors that may have occurred, and intermediate results for
+/// each command.
+#[derive(uniffi::Record)]
+pub struct DryRunEffects {
```
Thoralf-M: Why rename it to effects?

crates/iota-sdk-ffi/src/graphql.rs
```diff
@@ -955,3 +960,85 @@ pub struct Query {
     #[serde(default)]
     pub variables: Option<serde_json::Value>,
 }
+
+impl iota_transaction_builder::ClientMethods for GraphQLClient {
```
Thoralf-M: Can't we just provide a reference to the inner client in the places where this is needed?

crates/iota-sdk-ffi/src/types/transaction/mod.rs
DaughterOfMars: I didn't really like the old name but I don't care that much

crates/iota-sdk-ffi/src/graphql.rs
DaughterOfMars: yes but then we also need to move that logic to another place. Just makes more sense to consider this type a client

crates/iota-sdk-ffi/src/types/transaction/mod.rs
Thoralf-M: Better to keep it aligned with the graphl schema if there is no strong reason against it

Thoralf-M (approved):

Alex6323 (approved):
lgtm

Thoralf-M (approved):

PR #318, opened 2025-10-15
chore: rename `type_` fields

thibault-martinez:


crates/iota-sdk-graphql-client/src/query_types/events.rs
```diff
@@ -56,6 +56,7 @@ pub struct EventFilter {
 pub struct Event {
     pub sending_module: Option<MoveModuleQuery>,
     pub sender: Option<GQLAddress>,
+    // TODO thibault
     pub type_: MoveType,
```
thibault-martinez, line 60: Did this one not need a `#[cynic(rename = "type")]`?

crates/iota-sdk-types/src/object.rs
```diff
@@ -304,7 +304,7 @@ pub struct MoveStruct {
         feature = "serde",
         serde(with = "::serde_with::As::<serialization::BinaryMoveStructType>")
     )]
-    pub type_: StructTag,
+    pub type_tag: StructTag,
```
thibault-martinez: Should this be named `struct_tag`?

crates/iota-sdk-graphql-client/src/query_types/object.rs
```diff
@@ -56,7 +56,7 @@ pub struct Object {
 #[cynic(schema = "rpc", graphql_type = "ObjectFilter")]
 pub struct ObjectFilter {
     #[cynic(rename = "type")]
-    pub type_: Option<String>,
+    pub type_tag: Option<String>,
```
thibault-martinez: This is actually called "type_filter" in the monorepo because the field is a `TypeFilter` enum. Wdyt?

crates/iota-sdk-graphql-client/src/query_types/normalized_move/module.rs
```diff
@@ -134,7 +134,7 @@ pub struct MoveEnumVariant {
 pub struct MoveField {
     pub name: String,
     #[cynic(rename = "type")]
-    pub type_: Option<OpenMoveType>,
+    pub type_tag: Option<OpenMoveType>,
```
thibault-martinez, line 137:
```suggestion
    pub open_move_type: Option<OpenMoveType>,
```
?

crates/iota-sdk-types/src/object.rs
DaughterOfMars: I would say so

crates/iota-sdk-graphql-client/src/query_types/object.rs
DaughterOfMars: That would be preferable

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/318/commits/48810939ef28d4e937669a1147efd3a503275314

crates/iota-sdk-types/src/object.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/318/commits/54bd772b991ad8b1917000bd125d74cadb31837c

crates/iota-sdk-graphql-client/src/query_types/normalized_move/module.rs
Alex6323: `move_type` is probably also fine. Looking at the other fields they are named very chill as well, so not sure it's necessary to be super correct here.

crates/iota-sdk-graphql-client/src/query_types/events.rs
Alex6323: I actually verified, that trailing underscores do not prevent the match against the schema. You can add as many as you want and it still works. Mapping fails however at compile-time if you prepend it or change the field name in any other way. Not sure this is documented well. We probably should still cynic-rename it just in case that doesn't work anymore at some point.

Alex6323 (approved):

crates/iota-sdk-graphql-client/src/query_types/events.rs
```diff
@@ -56,6 +56,7 @@ pub struct EventFilter {
 pub struct Event {
     pub sending_module: Option<MoveModuleQuery>,
     pub sender: Option<GQLAddress>,
+    // TODO thibault
```
DaughterOfMars, line 59: TODO

PR #320, opened 2025-10-16
feat(iota-transaction-builder): improve send_coins docs

### LINKED ISSUE #292: Tx builder: document TransactionBuilder::send_coins()

Thoralf-M:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/292


crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -379,6 +379,42 @@ impl<C, L> TransactionBuilder<C, L> {
...
+    /// If `amount` is `None`, the entire coins are transferred.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use iota_graphql_client::Client;
+    /// use iota_transaction_builder::TransactionBuilder;
+    /// use iota_types::{Address, ObjectId};
+    ///
+    /// #[tokio::main(flavor = "current_thread")]
```
DaughterOfMars: Can we hide the method and stuff?

Thoralf-M: Also imports and client creation?

DaughterOfMars: I dunno, I'm never sure which is the most important stuff to show vs hide tbh 😆 @thibault-martinez or @Alex6323 have an opinion?

thibault-martinez: Idk, if it was only me I wouldn't hide anything so it's fully and easily copiable 🤷🏻

DaughterOfMars: It is copyable still, but now you don't need to remove the method and such

thibault-martinez: What method?
If we hide something from the docs, then the example is not compilable anymore, or what am I missing?

DaughterOfMars: The main method

thibault-martinez: Well, my opinion is that I wouldn't hide anything

DaughterOfMars: It's just simply cleaner without it, and it's not needed. If anything it makes it harder to copy and paste the example because most time you won't want the main method

Alex6323: I mean importing is kinda implied and user probably figure it out (or their editor). However I would also not show the `main` method, because that is only necessary to run it as a doc test. Therefore, I would also not show the `Ok(())` because that's very likely different for the user anyway.

So, IMHO, the method body except Ok(() should be documentation enough.

Alex6323: On a second thought, then `?` doesn't necessarily make sense then, so I am not sure, there's a best way here. We probably don't want people to run `main` single threaded, which they might think is necessary for the example, because they are beginners, I dunno.

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -443,6 +443,42 @@ impl<C, L> TransactionBuilder<C, L> {
 
     /// Transfer some coins to a recipient address. If multiple coins are
     /// provided then they will be merged.
+    ///
+    /// The `amount` parameter specifies the quantity in NANOS, where 1 IOTA
+    /// equals 1_000_000_000 NANOS.
+    /// If `amount` is provided, that amount is split from the provided coins
+    /// and sent.
+    /// If `amount` is `None`, the entire coins are transferred.
```
thibault-martinez, line 451: Add this section to FFI?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
Thoralf-M: I've hidden the main fn part now: https://github.com/iotaledger/iota-rust-sdk/pull/320/commits/ccb951f7c4162638556b3a4e36b2a99951fc5e01

crates/iota-sdk-transaction-builder/src/builder/mod.rs
Thoralf-M: Sure https://github.com/iotaledger/iota-rust-sdk/pull/320/commits/bcefb94002ab0f125cc57ed0c028e7d34866bf78

crates/iota-sdk-ffi/src/transaction_builder/mod.rs
```diff
@@ -180,6 +180,19 @@ impl TransactionBuilder {
...
+    /// If `amount` is provided, that amount is split from the provided coins
+    /// and sent.
+    /// If `amount` is `None`, the entire coins are transferred.
+    ///
+    /// All provided coins must have the same coin type. Mixing coins of
+    /// different types will result in an error.
+    ///
+    /// If you intend to transfer all provided coins to another address in a
+    /// single transaction, consider using
+    /// [`TransactionBuilder::transfer_objects()`] instead.
```
thibault-martinez: Not too sure about the link in the bindings

thibault-martinez (approved):

crates/iota-sdk-ffi/src/transaction_builder/mod.rs
DaughterOfMars: Yeah this should just be without the brackets

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/320/commits/627363d772c081b595be2d02a26953dc9e34eeb1

DaughterOfMars (approved):

PR #326, opened 2025-10-20
feat(ci): add coverage workflow

### LINKED ISSUE #288: CI: add coverage workflow
We have something in the monorepo, should be easy to copy

thibault-martinez:
Closes https://github.com/iotaledger/iota-rust-sdk/issues/288


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,148 @@
+name: Code Coverage (llvm-cov)
+
+on:
+  push:
+    branches: [develop, sdk-bindings]
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  schedule:
+    - cron: "0 1 * * *" # every day at 1am (UTC)
```
DaughterOfMars, line 9: Why are we scheduling this? Or maybe why run on push if we are scheduling it?

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,148 @@
...
+          else
+            echo "🚨 Error 🚨: llvm-profdata not found at $LLVM_PROFDATA, or is not executable."
+            exit 1
+          fi
+
+      - name: Run code coverage (nextest)
+        run: |
+          set +e
+
+          ./run_localnet.sh start ./iota
```
DaughterOfMars, line 95: Wait, why are you running the localnet here?

thibault-martinez: Because some tests need it?

.github/workflows/coverage.yml
thibault-martinez: The push/pull was only temporary to test it, will remove and leave scheduling/dispatch

.github/workflows/coverage.yml
DaughterOfMars: Ohhh, right 😆 of course

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,148 @@
...
+          ref: ${{ github.event.inputs.branch-or-commit || github.ref }}
+
+      - uses: bmwill/rust-cache@cb2cf0cc7c5198d3364b9630e2c3d457f160790c # v1.4.0
+
+      - name: Set Swap Space
+        uses: actionhippie/swap-space@73376950a0019f8e1f6a3d3d2673fe2aed74ae17 # v1.0.2
+        with:
+          size: 256G
+
+      - name: Get the IOTA testnet binary
```
DaughterOfMars, line 64: Why not use the `start_local_network` action? If we don't need the gas station we could just add a flag to the inputs to control that

thibault-martinez: Because coverage needs a self hosted runner and the action leaves a lot of stuff behind that would be annoying for future runs, the local network script takes care of reverting everything.

DaughterOfMars: Can we change the action then?

thibault-martinez: Idk, not sure it's worth it. I could make an attempt at the action using the localnet script but I'm not sure we gain anything from it.

DaughterOfMars: It's nice to not have two places to change if we need to fix something

Thoralf-M (approved):

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,148 @@
...
+            --failure-mode=warn \
+            --sparse \
+            --output target/nextest.profdata
+
+          if [ -s "target/nextest.profdata" ]; then
+            echo "Successfully collected nextest coverage data."
+          else
+            echo "🚨 Error 🚨: Collecting nexttest coverage failed."
+            exit 1
+          fi
```
Alex6323: we probably won't need that for the sdk tbh, at least I would try a few times without this. The problem in the monorepo was that running the code coverage on top of the tests often overpowered the runner's memory, and it was non-deterministic, so we needed this to make the coverage generation stable. In Sui they opted for just folding (in the sense of giving up and letting it fail). In our case I think the problem was more severe (less powerful hardware, or maybe also due to the transition from Stardust).

The SDK is much much smaller. I think creating coverage should almost never fail. We can even try to get the `branch coverage analysis`, not just only the line coverage working here. But that was too much for our runners against the monorepo 😿

DaughterOfMars (approved):

.github/workflows/coverage.yml
thibault-martinez: I know, I just wanted to keep the setup in case we have another source and it would be easier to merge. But I guess you are right, this is maybe slightly over-engineered for the SDK, I'll simplify.

thibault-martinez: Oh you weren't speaking about the merge. Ok yeah I'll remove.

thibault-martinez:
Closing this PR as the base branch change broke coveralls and I can't properly test anymore. Will open another one.

PR #329, opened 2025-10-20
chore: Add rust release CI

### LINKED ISSUE #242: Create release workflow to publish Rust crates to crates.io

DaughterOfMars:
## Description

Adds a CI job to publish the rust crates, and one to create a github release.  The first workflow uses [`cargo-release`](https://github.com/crate-ci/cargo-release) to do the hard work of bumping the versions, publishing the crates, and pushing the commit to the branch. The second workflow creates a github release with a changelog.

Renames the `iota-crypto` crate to `iota-crypto-2`.

The following crates will be published:
- `iota-crypto-2 v0.0.1`
- `iota-graphql-client v0.0.1`
- `iota-sdk v3.0.0`
- `iota-sdk-types v0.0.1`
- `iota-transaction-builder v0.0.1`

Closes #242

### DIFF
--- a/.github/workflows/pre_publish.yml
+++ b/.github/workflows/pre_publish.yml
@@ -30,12 +30,15 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
-      - run: cargo install cargo-release
+      - run: cargo install cargo-release --version 0.25.20
 
       - name: Cargo release publish
         run: |
           pkg_args="-p ${{inputs.package}}"
-          args="--allow-branch develop,sdk-bindings --no-confirm --execute"
+          args="--allow-branch develop,sdk-bindings --no-confirm"
+
+          echo "Dry-running publish"
+          eval "cargo release $pkg_args $args --all-features"
 
           echo "Commits since last tag:"
           eval "cargo release changes"
--- a/.github/workflows/pre_publish.yml
+++ b/.github/workflows/pre_publish.yml
@@ -18,39 +18,43 @@ on:
           - beta:    Increase the beta pre-version (x.y.z-beta.M)
           - alpha:   Increase the alpha pre-version (x.y.z-alpha.M)
       package:
-        type: string
+        type: choice
         required: true
         description: The package to publish.
+        options:
+          - iota-sdk
+          - iota-sdk-crypto
+          - iota-sdk-ffi
+          - iota-sdk-graphql-client
+          - iota-sdk-graphql-client-build
+          - iota-sdk-transaction-builder
+          - iota-sdk-types
 
 jobs:
   publish:
-    name: Publish to crates.io
+    name: Setup for publish to crates.io
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
-      - run: cargo install cargo-release --version 0.25.20
+      - name: Install cargo-release
+        run: cargo install cargo-release --version 0.25.20
 
-      - name: Cargo release publish
-        run: |
-          pkg_args="-p ${{inputs.package}}"
-          args="--allow-branch develop,sdk-bindings --no-confirm"
+      - name: Dry run publish
+        run: eval "cargo release -p ${{inputs.package}} --allow-branch develop,sdk-bindings --no-confirm --all-features"
 
-          echo "Dry-running publish"
-          eval "cargo release $pkg_args $args --all-features"
+      - name: Get commits since last tag
+        run: eval "cargo release changes"
 
-          echo "Commits since last tag:"
-          eval "cargo release changes"
+      - name: Bump versions
+        run: eval "cargo release version -p ${{inputs.package}} --allow-branch develop,sdk-bindings --no-confirm ${{inputs.level}}"
 
-          echo "Bumping versions"
-          eval "cargo release version $pkg_args $args ${{inputs.level}}"
+      - name: Commit packages
+        run: eval "cargo release commit --allow-branch develop,sdk-bindings --no-confirm"
 
-          echo "Committing packages"
-          eval "cargo release commit $args"
+      - name: Tag commit
+        run: eval "cargo release tag -p ${{inputs.package}} --allow-branch develop,sdk-bindings --no-confirm --tag-prefix '${{inputs.package}}-'"
 
-          echo "Tagging the commit"
-          eval "cargo release tag $pkg_args $args --tag-prefix '${{inputs.package}}-'"
-
-          echo "Pushing commit"
-          eval "cargo release push $pkg_args $args"
+      - name: Push commit
+        run: eval "cargo release push -p ${{inputs.package}} --allow-branch develop,sdk-bindings --no-confirm"
--- a/.github/workflows/pre_publish.yml
+++ b/.github/workflows/pre_publish.yml
@@ -1,0 +1,53 @@
+name: Pre-publish
+
+on:
+  workflow_dispatch:
+    inputs:
+      level:
+        type: string
+        required: true
+        description: |
+          The level of version bump to make.
+
+          Possible values:
+          - major:   Increase the major version (x.0.0)
+          - minor:   Increase the minor version (x.y.0)
+          - patch:   Increase the patch version (x.y.z)
+          - release: Remove the pre-version (x.y.z)
+          - rc:      Increase the rc pre-version (x.y.z-rc.M)
+          - beta:    Increase the beta pre-version (x.y.z-beta.M)
+          - alpha:   Increase the alpha pre-version (x.y.z-alpha.M)
+      package:
+        type: string
+        required: true
+        description: The package to publish.
+
+jobs:
+  publish:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - run: cargo install cargo-release
+
+      - name: Cargo release publish
+        run: |
+          pkg_args="-p ${{inputs.package}}"
+          args="--allow-branch develop,sdk-bindings --no-confirm --execute"
+
+          echo "Commits since last tag:"
+          eval "cargo release changes"
+
+          echo "Bumping versions"
+          eval "cargo release version $pkg_args $args ${{inputs.level}}"
+
+          echo "Committing packages"
+          eval "cargo release commit $args"
+
+          echo "Tagging the commit"
+          eval "cargo release tag $pkg_args $args --tag-prefix '${{inputs.package}}-'"
+
+          echo "Pushing commit"
+          eval "cargo release push $pkg_args $args"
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Cargo release publish
+        run: |
+          cmd="cargo release \
+            --workspace \
+            --exclude iota-graphql-client-build \
+            --exclude iota-sdk-ffi \
+            --all-features \
+            --allow-branch develop,sdk-bindings
+            --no-confirm \
+            --execute"
```
Alex6323, line 28: without `--execute` it would perform a dry-run?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,49 @@
...
+          if [ -n ${{inputs.package}} ]; then
+            cmd="$cmd -p ${{inputs.package}}"
+          fi
+          if [ -n ${{inputs.level}} ]; then
+            cmd="$cmd ${{inputs.level}}"
+          fi
```
Alex6323: can this be an `elif` since `cmd` is overwritten?

.github/workflows/publish.yml
DaughterOfMars: Yes, that is the default behavior

.github/workflows/publish.yml
DaughterOfMars: it's appending

DaughterOfMars: no longer relevant 😆

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  publish:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - run: cargo install cargo-release
```
Thoralf-M: Should we specify the version?
```suggestion
      - run: cargo install cargo-release --version 0.25.20
```

.github/workflows/pre_publish.yml
```diff
@@ -0,0 +1,53 @@
...
+      package:
+        type: string
+        required: true
+        description: The package to publish.
+
+jobs:
+  publish:
+    name: Publish to crates.io
+    runs-on: ubuntu-latest
+    steps:
```
Thoralf-M, line 36: Should we also do a dry run of cargo release here? If there is an issue it would create the commit and tag already and then only fail in the release.yml, right?

DaughterOfMars: I suppose we can

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+          echo "tag is: ${{ github.ref_name }}"
+
+      - name: Install node
+        uses: actions/setup-node@v6
+
+      - name: Install conventional-commits
+        run: npm -g i conventional-commits
+
+      - name: Generate Release Notes
+        run: conventional-changelog -p conventionalcommits -i notes-${{ github.ref_name }}.md -s
```
Thoralf-M: Do we want that? Wouldn't this include also the commits for other parts of the repo?

DaughterOfMars: Hmm, good point. Maybe we can filter it

.github/workflows/pre_publish.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - run: cargo install cargo-release --version 0.25.20
+
+      - name: Cargo release publish
+        run: |
+          pkg_args="-p ${{inputs.package}}"
```
Alex6323: can only be a single package, or does it also accept a comma separated list?

DaughterOfMars: it can be multiple, but in this case we want to do individual package releases

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v*.*.*"
```
thibault-martinez: Need to change that to include prefix

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Generate Release Notes
+        run: conventional-changelog -p conventionalcommits -i notes-${{ github.ref_name }}.md -s
+
+      - name: Create GitHub release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh release create $TAG --verify-tag --title $TAG --draft --notes-file notes-${{ github.ref_name }}.md
+
+      - name: Update major tag
```
thibault-martinez: Investigate if we really need this

DaughterOfMars: I believe we do not need it

.github/workflows/release.yml
DaughterOfMars: done

.github/workflows/release.yml
DaughterOfMars: I added a path filter

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+    tags:
+      - "*-v*.*.*"
+
+# We need this to be able to create releases.
+permissions:
+  contents: write
+
+jobs:
+  create-release:
+    name: create-release
```
thibault-martinez:
```suggestion
```
It's already the same name

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Show the tag
+        run: |
+          echo "tag is: ${{ github.ref_name }}"
```
thibault-martinez, line 21:
```suggestion
        run: echo "tag is: ${{ github.ref_name }}"
```
More consistent with the other one-line runs (except if there is a reason I don't know)

thibault-martinez, line 21: Why would the tag be ref_name? It's automatically set on push: tags ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Install node
+        uses: actions/setup-node@v6
```
thibault-martinez: Please pin commit + version comment

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: npm -g i conventional-commits
+
+      - name: Generate Release Notes
+        run: |
+          pkg=$(echo "${{github.ref_name}}" | sed -nr 's|(.*)-v.*|\1|p'
+          conventional-changelog -p conventionalcommits -i notes-${{ github.ref_name }}.md -s --commit-path "crates/$pkg"
+
+      - name: Create GitHub release
+        env:
+          GH_TOKEN: ${{ github.token }}
```
thibault-martinez: This seems odd

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: gh release create $TAG --verify-tag --title $TAG --draft --notes-file notes-${{ github.ref_name }}.md
```
thibault-martinez: Where has `$TAG` been defined?

.github/workflows/pre_publish.yml
```diff
@@ -0,0 +1,56 @@
...
+          - minor:   Increase the minor version (x.y.0)
+          - patch:   Increase the patch version (x.y.z)
+          - release: Remove the pre-version (x.y.z)
+          - rc:      Increase the rc pre-version (x.y.z-rc.M)
+          - beta:    Increase the beta pre-version (x.y.z-beta.M)
+          - alpha:   Increase the alpha pre-version (x.y.z-alpha.M)
+      package:
+        type: string
+        required: true
+        description: The package to publish.
```
thibault-martinez, line 23: Not necessarily now but at some point I'd make this a dropdown with the possible values so that it's less error-prone

.github/workflows/pre_publish.yml
```diff
@@ -0,0 +1,56 @@
...
+
+jobs:
+  publish:
+    name: Publish to crates.io
```
thibault-martinez: Doesn't really publish it though, right?

.github/workflows/pre_publish.yml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - run: cargo install cargo-release --version 0.25.20
+
+      - name: Cargo release publish
```
thibault-martinez: Same, maybe adapt the name a bit

.github/workflows/release.yml
DaughterOfMars: yes, afaik

.github/workflows/release.yml
DaughterOfMars: why?

thibault-martinez: Idk, we seem to always use `${{ secrets.GITHUB_TOKEN }}` instead

.github/workflows/release.yml
DaughterOfMars: there is an error when I remove this for some reason

.github/workflows/pre_publish.yml
```diff
@@ -0,0 +1,60 @@
...
+          - beta:    Increase the beta pre-version (x.y.z-beta.M)
+          - alpha:   Increase the alpha pre-version (x.y.z-alpha.M)
+      package:
+        type: choice
+        required: true
+        description: The package to publish.
+        options:
+          - iota-sdk
+          - iota-sdk-crypto
+          - iota-sdk-ffi
```
thibault-martinez: remove

DaughterOfMars: oop

thibault-martinez (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "*-v*.*.*"
```
Thoralf-M, line 6: So what about this, how do we filter for the rust crates only?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+      - run: cargo install cargo-release --version 0.25.20
+
+      - name: Cargo release publish
+        run: |
+          pkg=$(echo "${{github.event.release.tag_name}}" | sed -nr 's|(.*)-v.*|\1|p'
+          version=$(echo "${{github.event.release.tag_name}}" | sed -nr 's|.*-v(.*)|\1|p'
+
+          echo "Publishing $pkg at version $version"
+          eval "cargo release publish \
+            -p $pkg \
```
Thoralf-M, line 24: This would fail if we publish a release for a binding, are we fine with that or should we somehow filter and not run this in the first place?

thibault-martinez: I am fine merging this and adapt in other PRs for the bindings

Thoralf-M (approved):

PR #332, opened 2025-10-20
feat(ffi): Add bcs conversion for primitive types + `String`

### LINKED ISSUE #231: Expose primitive type + `String` bcs conversion to bindings

Alex6323:
Closes #231


thibault-martinez:
How does this close #231?

DaughterOfMars (commented):
Yeah this isn't doing anything. The issue is about adding `to_bcs` and `from_bcs` for every type we want to serialize

crates/iota-sdk-ffi/src/lib.rs
```diff
@@ -137,3 +138,24 @@ pub fn hex_encode(input: &[u8]) -> String {
 pub fn hex_decode(input: String) -> crate::error::Result<Vec<u8>> {
     Ok(hex::decode(input)?)
 }
+
+macro_rules! export_basic_types_bcs_conversion {
```
DaughterOfMars: Can we call this `export_primitive_types_bcs_conversion`

crates/iota-sdk-ffi/src/lib.rs
```diff
@@ -137,3 +138,24 @@ pub fn hex_encode(input: &[u8]) -> String {
...
+            #[uniffi::export]
+            pub fn [< $name:snake _from_bcs >](input: &[u8]) -> crate::error::Result<$name> {
+                Ok(bcs::from_bytes(input)?)
+            }
+
+            /// Convert this type to BCS encoded bytes.
+            #[uniffi::export]
+            pub fn [< $name:snake _to_bcs >](input: $name) -> crate::error::Result<Vec<u8>> {
+                Ok(bcs::to_bytes(&input)?)
+
```
DaughterOfMars, line 56:
```suggestion
```

crates/iota-sdk-ffi/src/lib.rs
Alex6323: I did not call it primitive on purpose because of `String`, which is a derived type. If nobody cares, we can rename it though.

DaughterOfMars: I guess, but tbh it's still more clear with primitive

thibault-martinez: Yeah I'd say primitive as well

crates/iota-sdk-ffi/src/lib.rs
```diff
@@ -43,3 +44,23 @@ pub fn hex_encode(input: &[u8]) -> String {
 pub fn hex_decode(input: String) -> crate::error::Result<Vec<u8>> {
     Ok(hex::decode(input)?)
 }
+
+macro_rules! export_primitive_types_bcs_conversion {
+    ($($name:ty),+ $(,)?) => {
+        paste::paste! {$(
+            /// Create this type from BCS encoded bytes.
```
thibault-martinez: Not a big deal but possible to use the type name instead of "this type"?

thibault-martinez (approved):

crates/iota-sdk-ffi/src/lib.rs
DaughterOfMars: I fixed it. Took waaaaaay to long to notice that paste supports doc strings

PR #335, opened 2025-10-22
feat(tx-builder): Unify gas methods and improve/update docs 

### LINKED ISSUE #291: Tx builder: document TransactionBuilder::gas_coins()

Alex6323:
1. removes `.gas()` for single gas coins
2. renames `.gas_coins()` to `.gas()` to now always allow to provide multiple gas coins.

Closes #291


DaughterOfMars:
Honestly I'd rather just accept an array in `gas` and remove the other fn

Alex6323:
> Honestly I'd rather just accept an array in `gas` and remove the other fn

sure about it? Most users will just pass one coin, and they'ld then have to deal with array syntax all the time. Providing an extra "multi" version is quite common in the monorepo, but :man_shrugging: @Thoralf-M @thibault-martinez opinions?

crates/iota-sdk-transaction-builder/src/lib.rs
```diff
@@ -128,7 +128,7 @@
 //! These methods set various metadata which may be needed for the execution.
 //!
 //! - [gas](TransactionBuilder::gas): Add a gas coin to pay for the execution.
-//! - [gas_coins](TransactionBuilder::gas_coins): Add gas coins to pay for the
+//! - [gas_multi](TransactionBuilder::gas_multi): Add gas coins to pay for the
```
Thoralf-M: This should be removed if we go for `gas` with multiple coins

Thoralf-M:
I also prefer a single `gas` method which accepts multiple gas coins, even if a user provides only a single gas coin it's not that annoying to make an array with just that and less methods make it less overwhelming

bindings/go/examples/gas_sponsor/main.go
```diff
@@ -30,7 +54,9 @@ func main() {
 		nil,
 	)
 
-	builder.Sponsor(sponsor)
+	gasObjId := objIdFromHex("0x0b0270ee9d27da0db09651e5f7338dfa32c7ee6441ccefa1f6e305735bcfc7ab")
+	gasObjIds := []*sdk.ObjectId{gasObjId}
+	builder.Gas(gasObjIds).Sponsor(sponsor)
```
DaughterOfMars: ???

Alex6323: don't ask, let's just say something got messed up with local and remote changes

Alex6323: oh wait, what are you actually asking? I am not changing the example's logic per se just now adapating to the Gas signature change. I see that the other gas_sponsor examples don't use it, and instead just rely on the sponsor. But why?

Alex6323: Aligned the go example with the other ones.

Thoralf-M (approved):

bindings/go/examples/gas_sponsor/main.go
DaughterOfMars: Ahhhh sorry I literally just was looking at a merge commit and got confused, which is why I resolved this immediately 🙈

crates/iota-sdk-transaction-builder/src/lib.rs
```diff
@@ -128,8 +128,6 @@
 //! These methods set various metadata which may be needed for the execution.
 //!
 //! - [gas](TransactionBuilder::gas): Add a gas coin to pay for the execution.
```
thibault-martinez: Maybe we want to slightly adapt this comment?

crates/iota-sdk-ffi/src/transaction_builder/mod.rs
```diff
@@ -68,18 +68,10 @@ impl TransactionBuilder {
...
 
-    /// Add a gas object to use to pay for the transaction.
-    pub fn gas(self: Arc<Self>, object_id: &ObjectId) -> Arc<Self> {
-        self.write(|builder| {
-            builder.gas(**object_id);
-        });
-        self
-    }
-
     /// Add gas objects to pay for the transaction.
```
thibault-martinez: Can we have the same one-liner as the base function?

Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/335/commits/c074b81a2ecfb54c807bde452eb5a377f53279b0

crates/iota-sdk-transaction-builder/src/lib.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/335/commits/c074b81a2ecfb54c807bde452eb5a377f53279b0

DaughterOfMars (approved):

thibault-martinez (approved):

PR #336, opened 2025-10-22
feat(tx-builder): improve `TransactionBuilder::make_move_vec()` docs

### LINKED ISSUE #296: Tx builder: document TransactionBuilder::make_move_vec()

Alex6323:
Closes #296


crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -514,6 +514,35 @@ impl<C, L> TransactionBuilder<C, L> {
...
+    ///
+    /// let address1 =
+    ///     Address::from_str("0xde49ea53fbadee67d3e35a097cdbea210b659676fc680a0b0c5f11d0763d375e")?;
+    /// let address2 =
+    ///     Address::from_str("0xe512234aa4ef6184c52663f09612b68f040dd0c45de037d96190a071ca5525b3")?;
+    ///
+    /// builder
+    ///     .move_call(Address::FRAMEWORK, "vec_map", "from_keys_values")
+    ///     .generics::<(Address, u64)>()
+    ///     .arguments((vec![address1, address2], vec![10000000u64, 20000000u64]));
```
DaughterOfMars: This isn't calling make_move_vec?

Alex6323: oopsie, thanks

Alex6323: fixed in https://github.com/iotaledger/iota-rust-sdk/pull/336/commits/bad7fb9cd4059a9b624feb1b32dd15c63d4a1ea2

crates/iota-sdk/examples/generic_move_function.rs
DaughterOfMars, line 1: This example isn't about make_move_vec and tbh it's a worse way to use the builder so I say revert this

crates/iota-transaction-builder/src/builder/mod.rs
```diff
@@ -514,6 +514,39 @@ impl<C, L> TransactionBuilder<C, L> {
     }
 
     /// Make a move vector from a list of elements.
```
DaughterOfMars, line 540: Maybe we should also mention something like:

```
/// Often it is possible (and more efficient) to pass a rust slice or `Vec` instead of calling this function, which will serialize the bytes into a move vector pure argument.

crates/iota-sdk/examples/generic_move_function.rs
Alex6323: okay, do you think we should have a separate example about `.make_move_vec` then?

crates/iota-transaction-builder/src/builder/mod.rs
Alex6323: yeah, makes sense to add this comment, Can you come up with a valid use-case? Then I'ld add another example just for that, because currently `.make_move_vec` is not covered I think.

crates/iota-sdk/examples/generic_move_function.rs
DaughterOfMars: yeah, maybe

crates/iota-transaction-builder/src/builder/mod.rs
DaughterOfMars: I don't know of a valid use-case other than if there's something we can't do with the normal syntax that I'm not thinking of

PR #338, opened 2025-10-22
fix(transaction-builder): Require `send_iota` amount param

DaughterOfMars:
## Description

The `amount` param for `send_iota` was optional, but this meant that not providing it would send the entire gas coin, which is probably undesirable. This PR makes it required.


thibault-martinez (approved):
I wouldn't say it was undesirable nor dangerous, seems pretty expected to me that no amount means the whole thing. But I'm fine changing it anyway.

DaughterOfMars:
> I wouldn't say it was undesirable nor dangerous, seems pretty expected to me that no amount means the whole thing. But I'm fine changing it anyway.

I agree with this in the case of `send_coins` where you specify the coins you want to send. But here the gas may be chosen for you and I think, even though it should be obvious that providing no amount would have SOME effect, it's not clear in this case what effect that would be.

PR #344, opened 2025-10-27
chore(examples): Add go helper methods to avoid ignoring errors

### LINKED ISSUE #286: Update go examples so that errors are not ignored when parsing hex
We changed the go examples to ignore these errors because it is simpler, but this means users copying these examples will get bad errors if they change the hex. We should add helper methods for parsing which will catch and print these errors as in https://github.com/iotaledger/iota-rust-sdk/blob/8cef843b80157ccc2c078500b34860fed5074827/bindings/go/examples/prepare_transfer_objects/main.go#L12

DaughterOfMars:
## Description

Adds go helper methods for all parsing so that errors can be properly reported.

Closes #286

### DIFF
--- a/bindings/go/examples/dev_inspect/main.go
+++ b/bindings/go/examples/dev_inspect/main.go
@@ -10,14 +10,38 @@ import (
 	sdk "bindings/iota_sdk_ffi"
 )
 
+func objIdFromHex(hex string) *sdk.ObjectId {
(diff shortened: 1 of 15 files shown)


Thoralf-M (approved):

thibault-martinez (approved):

thibault-martinez:
Conflicts, and apply this PR to the 2 new examples if need be

PR #345, opened 2025-10-27
feat(transaction-builder): Add high level stake and unstake fns

### LINKED ISSUE #311: Tx builder: add high level un/staking methods
Similar to what we have here https://github.com/iotaledger/iota/blob/develop/crates/iota-transaction-builder/src/stake.rs
Maybe one should just provide an amount instead of objects as parameter and then the amount is split from the gas coin and staked?

DaughterOfMars:
## Description

Adds two high level functions for staking to the transaction builder.

Closes #311

### DIFF
--- a/crates/iota-sdk-ffi/src/transaction_builder/mod.rs
+++ b/crates/iota-sdk-ffi/src/transaction_builder/mod.rs
@@ -310,6 +310,22 @@ impl TransactionBuilder {
         self
     }
 
+    /// Stake IOTA coins to a validator.
+    pub fn stake(self: Arc<Self>, stake: &PTBArgument, validator_address: &Address) -> Arc<Self> {
+        self.write(|builder| {
+            builder.stake(stake, **validator_address);
+        });
+        self
+    }
+
+    /// Unstake IOTA coins from their assigned validator.
+    pub fn unstake(self: Arc<Self>, staked_iota: &PTBArgument) -> Arc<Self> {
+        self.write(|builder| {
+            builder.unstake(staked_iota);
+        });
+        self
+    }
+
     /// Convert this builder into a transaction.
     pub async fn finish(&self) -> Result<Transaction> {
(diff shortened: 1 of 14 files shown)


Thoralf-M (dismissed):

thibault-martinez:
CI failing

thibault-martinez:
I am fine with this version but the issue suggested 

> Maybe one should just provide an amount instead of objects as parameter and then the amount is split from the gas coin and staked?

Have we considered it?

DaughterOfMars:
> I am fine with this version but the issue suggested
> 
> > Maybe one should just provide an amount instead of objects as parameter and then the amount is split from the gas coin and staked?
> 
> Have we considered it?

Didn't see that, but it's a good idea. I'll change it.

crates/iota-sdk/examples/unstake.rs
```diff
@@ -24,12 +24,13 @@ async fn main() -> Result<()> {
         .next()
         .ok_or_eyre("no staked iota found")?;
 
+    println!("{}", staked_iota.object_id());
+    println!("{}", *staked_iota.owner().as_address());
```
thibault-martinez: Was this debug? If not, add it to the bindings

DaughterOfMars: oh, yeah, oops

crates/iota-sdk-ffi/src/transaction_builder/mod.rs
```diff
@@ -310,6 +310,22 @@ impl TransactionBuilder {
         self
     }
 
+    /// Stake IOTA coins to a validator.
```
thibault-martinez: The one-line is different from the main function, can we have the same?
Same for unstake

thibault-martinez (approved):

Thoralf-M (approved):

PR #347, opened 2025-10-27
feat(crypto): Add ability to create a private key from a mnemonic

### LINKED ISSUE #266: Allow to create private key from mnemonic
Currently we have `Ed25519PrivateKey::generate()` method to randomly create a private key and derive an address. I think it would be interesting to also create a deterministic private key from a mnemonic you provide similar to what is possible in the monorepo.

DaughterOfMars:
## Description

Allow creating private keys from mnemonics. While working on this, it made sense to separate the `PrivateKeyExt` trait into several smaller traits.

Closes #266


crates/iota-crypto/src/ed25519.rs
```diff
@@ -132,6 +134,30 @@ impl crate::PrivateKeyExt for Ed25519PrivateKey {
...
+    const SCHEME: SignatureScheme = SignatureScheme::Ed25519;
+}
+
+#[cfg(feature = "mnemonic")]
+impl crate::FromMnemonic for Ed25519PrivateKey {
+    type Error = crate::PrivateKeyError;
+
+    fn from_mnemonic(phrase: &str) -> Result<Self, Self::Error> {
+        let mnemonic = bip32::Mnemonic::new(phrase, bip32::Language::English)?;
+        let seed = mnemonic.to_seed("");
```
Thoralf-M: It should be also possible to provide a password

crates/iota-crypto/src/ed25519.rs
```diff
@@ -132,6 +134,30 @@ impl crate::PrivateKeyExt for Ed25519PrivateKey {
...
+        Ok(Self::new(slip10_ed25519::derive_ed25519_private_key(
+            seed.as_bytes(),
+            &[
+                crate::DERIVATION_PATH_PURPOSE_ED25519,
+                crate::DERIVATION_PATH_COIN_TYPE,
+                0,
+                0,
+                0,
+            ],
```
Thoralf-M: Fine to have this as default, but we should also allow other paths

DaughterOfMars: The whole path or just the last three?

crates/iota-crypto/src/lib.rs
```diff
@@ -273,3 +332,14 @@ pub trait PrivateKeyExt {
...
     }
 }
+
+/// Defines a type which can be constructed from a mnemonic phrase
+#[cfg(feature = "mnemonic")]
+pub trait FromMnemonic {
+    type Error;
+
+    /// Create an instance from a mnemonic phrase
+    fn from_mnemonic(phrase: &str) -> Result<Self, Self::Error>
```
Alex6323: maybe generalize a bit for usability `impl AsRef<str>`?

crates/iota-crypto/src/ed25519.rs
Thoralf-M: Hm not sure about this, the wallets we have only increase the account index for new accounts, maybe we should only allow this then so it's compatible
People that want something else can use a bip39 library directly to derive a private key with other bip32 paths

crates/iota-crypto/src/lib.rs
DaughterOfMars: I don't like using AsRef really, &str is usable in almost all situations and avoids generics which should be used sparingly

crates/iota-sdk-types/src/crypto/signature.rs
```diff
@@ -212,16 +212,23 @@ impl SimpleSignature {
 /// zklogin-flag     = %x05
 /// passkey-flag     = %x06
 /// ```
-#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, derive_more::Display)]
```
Alex6323, line 215: derive_more::Display is a bit annoying with the manual lowercasing. One alternative would be to use `strum` and add:

```
#[derive(strum::Display)]
#[strum(serialize_all = "lowercase")]
```

to the enum to make it less noisy and typo-resistent.

crates/iota-crypto/src/lib.rs
Alex6323: fair enough

thibault-martinez (commented):
Maybe this deserves a new example or an addition to an existing one?

crates/iota-sdk-types/src/crypto/signature.rs
DaughterOfMars: I think I will address this in a new PR, because there are a few places we could do this

Alex6323: you can also let me do it ...you have enough tasks I think

crates/iota-crypto/src/ed25519.rs
DaughterOfMars: @thibault-martinez thoughts?

thibault-martinez: Is this question still relevant?
> but we should also allow other paths

I see we are taking a path as parameter?

crates/iota-sdk/examples/address_from_mnemonic.rs
```diff
@@ -0,0 +1,51 @@
...
+    let flagged_public_key = Base64::encode_string(&public_key.to_flagged_bytes());
+    let address = public_key.derive_address();
+
+    println!("\nSecp256k1\n---");
+    println!("Private Key: {private_key_bech32}");
+    println!("Public Key: {public_key}");
+    println!("Public Key With Flag: {flagged_public_key}");
+    println!("Address: {address}");
+
+    let private_key = Secp256r1PrivateKey::from_mnemonic(MNEMONIC, None, None)?;
```
thibault-martinez: Maybe at least one example with a password and/or path?

crates/iota-sdk-crypto/src/lib.rs
```diff
@@ -182,23 +188,66 @@ impl<T: Verifier<UserSignature>> IotaVerifier for T {
...
+        Self: Sized;
+}
+
+/// Defines the const scheme of a private key
+pub trait ConstPrivateKeyScheme {
     const SCHEME: iota_types::SignatureScheme;
+}
 
+/// Defines the scheme of a private key
+pub trait PrivateKeyScheme {
```
thibault-martinez: Splitting this in 2 feels over the top to me

crates/iota-sdk-crypto/src/lib.rs
```diff
@@ -182,23 +188,66 @@ impl<T: Verifier<UserSignature>> IotaVerifier for T {
...
+pub const DERIVATION_PATH_PURPOSE_SECP256R1: u32 = 74;
+
+/// Defines a type which can be converted to bytes
+pub trait ToBytes {
+    /// Returns the raw bytes of this type.
+    fn to_bytes(&self) -> Vec<u8>;
+}
+
+/// Defines a type which can be constructed from bytes
+pub trait FromBytes {
```
thibault-martinez: Why are these 2 traits when `ToFromBech32` is one?

crates/iota-crypto/src/ed25519.rs
Thoralf-M: This part not, as @DaughterOfMars added the path now, but the question is whether we want the full path or just an account index

crates/iota-sdk-crypto/src/lib.rs
DaughterOfMars: Well only some types can have a const scheme

crates/iota-sdk-crypto/src/lib.rs
DaughterOfMars: Because some types cannot impl FromBytes

crates/iota-sdk-crypto/src/lib.rs
thibault-martinez: Ok but you can't really implement `scheme()` if you don't have one? Or what am I missing?

crates/iota-sdk-crypto/src/lib.rs
thibault-martinez: Like what?

DaughterOfMars: `SimpleKeypair`

crates/iota-sdk-crypto/src/lib.rs
DaughterOfMars: `SimpleKeypair` can impl `PrivateKeyScheme` but not the const version

crates/iota-sdk/examples/address_from_mnemonic.rs
DaughterOfMars: done

crates/iota-sdk-crypto/src/secp256r1.rs
```diff
@@ -136,6 +139,39 @@ impl crate::PrivateKeyExt for Secp256r1PrivateKey {
...
+        use std::str::FromStr;
+
+        use crate::FromBytes;
+
+        let mnemonic = bip39::Mnemonic::parse_in_normalized(bip39::Language::English, phrase)?;
+        let seed = mnemonic.to_seed(password.into().unwrap_or_default());
+
+        let path = path.into().unwrap_or_else(|| {
+            format!(
+                "m/{}'/{}'/0'/0'/0'",
```
Alex6323: I am asking this because I am unsure about this myself, but are you sure about the hardening in this path? AFAIK this scheme allows safe non-hardened derivation as compared to ED25519. So just checking whether you've thought about this.

crates/iota-sdk-ffi/src/crypto/simple.rs
```diff
@@ -114,6 +114,21 @@ impl SimpleKeypair {
         Ok(self.0.to_pem()?)
     }
 
+    /// Construct the private key from a mnemonic phrase and the signature
+    /// scheme
+    #[uniffi::constructor(default(password = None, path = None))]
+    pub fn from_mnemonic(
+        scheme: SignatureScheme,
+        phrase: &str,
```
Alex6323: maybe the phrase should come first, then the scheme?

crates/iota-sdk/examples/address_from_mnemonic.rs
```diff
@@ -0,0 +1,55 @@
...
+};
+use iota_types::PublicKeyExt;
+
+const MNEMONIC: &str = "round attack kitchen wink winter music trip tiny nephew hire orange what";
+
+fn main() -> eyre::Result<()> {
+    let private_key = Ed25519PrivateKey::from_mnemonic(MNEMONIC, None, None)?;
+    let private_key_bech32 = private_key.to_bech32().unwrap();
+    let public_key = private_key.public_key();
+    let flagged_public_key = Base64::encode_string(&public_key.to_flagged_bytes());
```
Alex6323, line 17: maybe we should have a `public_key.to_string_with_flag()` or smth?

crates/iota-sdk-crypto/src/secp256r1.rs
DaughterOfMars: I have not thought about it and I do not know what the best way is

crates/iota-sdk-ffi/src/crypto/simple.rs
DaughterOfMars: I feel like the scheme is the most important thing because it determines the type of key, which is why I put it first

crates/iota-sdk/examples/address_from_mnemonic.rs
DaughterOfMars: maybe but I don't think that is in the scope of this PR

crates/iota-sdk-ffi/src/crypto/simple.rs
thibault-martinez: I agree with scheme being first since it's being matched on and the rest of the params are just being forwarded

Alex6323: isn't the method called `from_mnemonic` which is represented more by the `phrase` than by the scheme? 

Actually, I would even prefer having it in last position (even if its the the most important internally), reason being that the argument list should be consistent with the other methods. Wouldn't those signatures be better overall from a user standpoint?

`.from_mnemonic(phrase, password, path)`
`.from_menmonic(phrase, password, path, scheme)`

crates/iota-sdk/examples/address_from_mnemonic.rs
Alex6323: I agree

crates/iota-sdk-ffi/src/crypto/simple.rs
DaughterOfMars: TBH I would also be fine with simply removing this method and the impls on `SimpleKeypair`. You can just create the inner keys and construct `SimpleKeypair` from them 🤷‍♀️

Thoralf-M: > TBH I would also be fine with simply removing this method and the impls on `SimpleKeypair`. You can just create the inner keys and construct `SimpleKeypair` from them 🤷‍♀️

I think I agree with that, we also don't have a method to create a random `SimpleKeypair` with one of the schemes

crates/iota-sdk-crypto/src/secp256r1.rs
Thoralf-M: monorepo has this
```
/// Secp256k1 follows BIP-32/44 using path where the first 3 levels are
/// hardened: m/54'/4218'/0'/0/{index} Secp256r1 follows BIP-32/44 using path
/// where the first 3 levels are hardened: m/74'/4218'/0'/0/{index}.
```

Thoralf-M: Probably good to add these test cases https://github.com/iotaledger/iota/blob/beae5b82e5564ac4bdaaa68a608a46632cc1073f/crates/iota/src/unit_tests/keytool_tests.rs#L309-L411

DaughterOfMars: done, thanks

DaughterOfMars: I still wonder whether we should take the whole path. Maybe two functions?

crates/iota-crypto/src/ed25519.rs
DaughterOfMars: Now we have both :)

crates/iota-sdk-crypto/src/lib.rs
```diff
@@ -197,32 +197,29 @@ pub const DERIVATION_PATH_PURPOSE_SECP256K1: u32 = 54;
...
+    }
 }
 
 /// Defines a type which can be constructed from bytes
-pub trait FromBytes {
+pub trait ToFromBytes {
     type Error;
 
+    /// Returns the raw bytes of this type.
+    fn to_bytes(&self) -> Vec<u8>;
```
Alex6323: is returning a slice not possible? 

Since it's now `to` and `from` maybe the docs could be something like:
```
/// Defines a type which can be constructed from and turned into bytes
```

crates/iota-sdk-crypto/src/lib.rs
```diff
@@ -341,3 +338,95 @@ pub trait FromMnemonic {
...
+    use crate::{
+        ed25519::Ed25519PrivateKey, secp256k1::Secp256k1PrivateKey, secp256r1::Secp256r1PrivateKey,
+    };
+
+    #[cfg(feature = "mnemonic")]
+    #[test]
+    fn test_mnemonics_ed25519() {
+        const TEST_CASES: [[&str; 3]; 3] = [
+            [
+                "film crazy soon outside stand loop subway crumble thrive popular green nuclear struggle pistol arm wife phrase warfare march wheat nephew ask sunny firm",
```
Alex6323, line 370: damn, this mnemonic is deep

Alex6323: `wife phrase warfare` I know what you're talking about

crates/iota-sdk-crypto/src/lib.rs
DaughterOfMars: No, because we cannot use generics in the trait for the length of the array. But we could maybe have `as_bytes` which returns a `&[u8]`, I'll take a look

DaughterOfMars: So it turns out there is a way to use const arrays, but `as_bytes` cannot work as the `secp256` keys don't store their underlying bytes, they have to serialize

Alex6323 (approved):

Thoralf-M (approved):

PR #349, opened 2025-10-27
feat(graphql): Allow waiting for finalization in client

### LINKED ISSUE #316: Add Client::wait_for_transaction()
Similar to what we have in the TS SDK
```TS
    /**
     * Wait for a transaction block result to be available over the API.
     * This can be used in conjunction with `executeTransactionBlock` to wait for the transaction to
     * be available via the API.
     * This currently polls the `getTransactionBlock` API to check for the transaction.
     */
    waitForTransaction({ signal, timeout, pollInterval, ...input }: {
        /** An optional abort signal that can be used to cancel */
        signal?: AbortSignal;
        /** The amount of time to wait for a transaction block. Defaults to one minute. */
        timeout?: number;
        /** The amount of time to wait between checks for the transaction block. Defaults to 2 seconds. */
        pollInterval?: number;
    } & Parameters<IotaClient['getTransactionBlock']>[0]): Promise<IotaTransactionBlockResponse>;
```

On `TransactionBuilder::execute()` we also have `wait_for_finalization` param, the code should be probably reused there and the same param could be added to `Client::execute_tx()`

DaughterOfMars:
## Description

Adds the `wait_for_tx_finalization` fn and the `wait_for_finalization` boolean flag to the `execute_tx` fn to allow for waiting in the client. I also changed the way we wait, so it is now a duration timeout rather than a number of retries.

Closes #316


crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1477,12 +1478,41 @@ impl Client {
...
+                self.wait_for_tx_finalization(tx.digest(), None).await?;
+            }
+
             Ok(Some(effects))
         } else {
             Ok(None)
         }
     }
 
+    pub async fn wait_for_tx_finalization(
```
Thoralf-M: Add a doc comment

crates/iota-graphql-client/src/lib.rs
```diff
@@ -1477,12 +1478,44 @@ impl Client {
...
     }
 
+    /// Wait for the finalization of a transaction by its digest. An optional
+    /// timeout can be provided, which, if exceeded, will return an error.
+    /// Returns the [`TransactionEffects`].
+    pub async fn wait_for_tx_finalization(
+        &self,
+        digest: Digest,
+        timeout: impl Into<Option<Duration>>,
+    ) -> Result<Option<TransactionEffects>> {
```
Thoralf-M: No need to return an Option if we return an error because of the timeout otherwise?

crates/iota-graphql-client/src/lib.rs
```diff
@@ -1477,12 +1478,44 @@ impl Client {
...
+            }
+
             Ok(Some(effects))
         } else {
             Ok(None)
         }
     }
 
+    /// Wait for the finalization of a transaction by its digest. An optional
+    /// timeout can be provided, which, if exceeded, will return an error.
```
Thoralf-M: Maybe also mention that default timeout is 60 seconds

crates/iota-graphql-client/src/lib.rs
DaughterOfMars: Good call, and I also applied this to the execute fn

crates/iota-graphql-client/src/lib.rs
DaughterOfMars: done

thibault-martinez:
Doesn't compile, can't see it on GH because of the conflict

crates/iota-sdk-graphql-client/src/error.rs
```diff
@@ -91,6 +103,7 @@ impl std::fmt::Display for Kind {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         match self {
             Kind::Deserialization => write!(f, "Deserialization error:"),
+            Kind::Missing => write!(f, "Missing:"),
```
thibault-martinez: Triggered by not placing it at the same position as the enum

crates/iota-sdk/examples/sign_send_iota.rs
```diff
@@ -37,14 +37,10 @@ async fn main() -> Result<()> {
...
 
-    let effects = client.execute_tx(&[signature], &tx).await?;
-    if let Some(effects) = effects {
-        println!("Digest: {}", effects.digest());
-        println!("Transaction status: {:?}", effects.status());
-        println!("Effects: {effects:#?}");
-    } else {
-        println!("Transaction execution failed");
-    }
+    let effects = client.execute_tx(&[signature], &tx, false).await?;
```
thibault-martinez: Why not wait?

Thoralf-M (approved):

crates/iota-sdk-ffi/src/graphql.rs
```diff
@@ -630,10 +632,25 @@ impl GraphQLClient {
...
+                wait_for_finalization,
             )
             .await?
-            .map(Into::into)
-            .map(Arc::new))
+            .into())
+    }
+
+    #[uniffi::method(default(timeout = None))]
+    pub async fn wait_for_tx_finalization(
```
Alex6323: also add docs?

crates/iota-sdk-graphql-client/src/error.rs
Alex6323: should we order alphabetically? I know we sometimes enforce it and sometimes not.

thibault-martinez: I don't have a strong opinion on that, as long as we keep the order consistent for the same enum

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1477,12 +1478,43 @@ impl Client {
...
+        timeout: impl Into<Option<Duration>>,
+    ) -> Result<TransactionEffects> {
+        tokio::time::timeout(
+            timeout.into().unwrap_or_else(|| Duration::from_secs(60)),
+            async {
+                loop {
+                    if let Some(effects) = self.transaction_effects(digest).await? {
+                        break Ok(effects);
+                    } else {
+                        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
Alex6323: maybe refactor this fn to use `tokio::time:interval` instead to control the polling? I think it's preferred over sleep in these kinds of scenarios.

DaughterOfMars: sure, done

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1477,12 +1478,43 @@ impl Client {
...
+            if wait_for_finalization {
+                self.wait_for_tx_finalization(tx.digest(), None).await?;
+            }
+
+            Ok(effects)
         } else {
-            Ok(None)
+            Err(Error::from_message(
+                Kind::Missing,
+                format!("transaction for digest {}", tx.digest()),
```
Alex6323:
```suggestion
                format!("effects for transaction {}", tx.digest()),
```

Alex6323 (approved):

bindings/go/examples/sign_send_iota/main.go
```diff
@@ -53,7 +53,7 @@ func main() {
 	}
 	userSignature := sdk.UserSignatureNewSimple(signature)
 
-	effects, err := client.ExecuteTx([]*sdk.UserSignature{userSignature}, txn)
+	effects, err := client.ExecuteTx([]*sdk.UserSignature{userSignature}, txn, false)
```
thibault-martinez: Also set to true in the bindings examples?

crates/iota-sdk/examples/gas_station.rs
```diff
@@ -27,11 +27,9 @@ async fn main() -> Result<()> {
             HeaderValue::from_str(&format!("Bearer {gas_station_auth_token}"))?,
         );
 
-    let res = builder.execute(&keypair.into(), true).await?;
+    let effects = builder.execute(&keypair.into(), true).await?;
```
thibault-martinez: Wouldn't something like this need to be changed in the bindings as well? I can picture go returning a pointer for an option

DaughterOfMars: yeah, forgot

bindings/go/examples/sign_send_iota/main.go
DaughterOfMars: done

crates/iota-sdk/examples/gas_station.rs
thibault-martinez: The gas_station bindings examples don't need to be updated?

crates/iota-sdk-transaction-builder/src/error.rs
```diff
@@ -27,6 +27,8 @@ pub enum Error {
     MissingVersion(ObjectId),
     #[error("Missing digest for object {0}")]
     MissingDigest(ObjectId),
+    #[error("Missing transaction for digest {0}")]
+    MissingTransaction(Digest),
```
thibault-martinez, line 31: I guess that ship has sailed but I always feel a bit weird this kind of stuff being an error instead of an option

crates/iota-sdk/examples/gas_station.rs
DaughterOfMars: The builder execute fn already had this parameter, remember?

crates/iota-sdk-transaction-builder/src/error.rs
DaughterOfMars: Well it is an error because it should not be missing

crates/iota-sdk/examples/gas_station.rs
thibault-martinez: I was speaking about the return not being an option anymore

DaughterOfMars: ah, you are right about that

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1477,12 +1478,43 @@ impl Client {
             let bcs = base64ct::Base64::decode_vec(result.effects.bcs.0.as_str())?;
             let effects: TransactionEffects = bcs::from_bytes(&bcs)?;
 
-            Ok(Some(effects))
+            if wait_for_finalization {
```
thibault-martinez: I don't understand this, we're waiting for effects but still returning the effects we tried to parse before?

DaughterOfMars: Hmm, good point. Although I'm not entirely sure that the response from graphql is actually an indication that the txn is finalized. If so then we really don't need this at all?

crates/iota-sdk-graphql-client/src/query_types/transaction.rs
```diff
@@ -130,6 +129,7 @@ pub struct TxBlockEffects {
 #[cynic(schema = "rpc", graphql_type = "TransactionBlockEffects")]
 pub struct TransactionBlockEffects {
     pub bcs: Option<Base64>,
+    pub checkpoint: Option<Checkpoint>,
```
Thoralf-M, line 165: Shouldn't this be an own query with the checkpoint only?

DaughterOfMars: I suppose I could do that, yeah

Alex6323 (approved):

thibault-martinez:
I'll do a final review before this is merged please 🙏🏻

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1407,80 +1245,124 @@ impl Client {
...
+                .is_some()
+            {
+                return Ok(true);
+            }
         }
+        Ok(false)
+    }
+
+    /// Wait for the finalization or checkpoint inclusion of a transaction
+    /// by its digest. An optional timeout can be provided, which, if
```
Thoralf-M: `finalization or checkpoint inclusion of a transaction` that's the same, you mean indexed by node?

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1407,80 +1245,124 @@ impl Client {
...
-            let bcs = base64ct::Base64::decode_vec(result.effects.bcs.0.as_str())?;
-            let effects: TransactionEffects = bcs::from_bytes(&bcs)?;
+        Ok(effects)
+    }
 
-            Ok(Some(effects))
-        } else {
-            Ok(None)
+    /// Returns whether the transaction for the given digest has been indexed
+    /// on the node.
```
Thoralf-M: Probably good to add that this doesn't necessarily mean that it's finalized

crates/iota-sdk/examples/sign_send_iota.rs
```diff
@@ -37,14 +37,13 @@ async fn main() -> Result<()> {
...
-    let effects = client.execute_tx(&[signature], &tx).await?;
-    if let Some(effects) = effects {
-        println!("Digest: {}", effects.digest());
-        println!("Transaction status: {:?}", effects.status());
-        println!("Effects: {effects:#?}");
-    } else {
-        println!("Transaction execution failed");
-    }
+    let effects = client
+        .execute_tx(&[signature], &tx, WaitForTx::Indexed)
```
Thoralf-M: Why wait for indexed here? Would just provide `None` if we don't actually need to do more with it afterwards

crates/iota-sdk-transaction-builder/src/error.rs
thibault-martinez: I disagree, it could be missing and that's fine, it's a state, not an error

thibault-martinez: I'm willing to approve anyway, but I don't think we should do that

thibault-martinez:
> I'll do a final review before this is merged please 🙏🏻

Merging this first thing tomorrow morning, one last look 👀

crates/iota-sdk-graphql-client/src/query_types/transaction.rs
```diff
@@ -126,12 +147,24 @@ pub struct TxBlockEffects {
     pub effects: Option<TransactionBlockEffects>,
 }
 
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema = "rpc", graphql_type = "TransactionBlock")]
+pub struct TxBlockCheckpoint {
```
thibault-martinez, line 152: The naming is really annoying here, `TxBlockCheckpoint` vs `TransactionBlockCheckpoint` vs `TransactionBlockCheckpointQuery`... Let's try to explicit what they actually do and how they differ

DaughterOfMars: I'm just mimicking the other query, and tbh I'm not sure there's much we can do to make this better. These queries are bound to have weird names because of the nature of `cynic` requiring so many structs

thibault-martinez: I disagree, we can probably derive a sensible name from the `graphql_type` + fields we are trying to fetch. Having `TxBlockCheckpoint` and `TransactionBlockCheckpoint` is not acceptable. But this probably means changing the whole file to harmonise so I'll open an issue for later.

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/issues/360

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1407,80 +1245,126 @@ impl Client {
...
         let response = self.run_query(&operation).await?;
 
-        if let Some(errors) = response.errors {
-            return Err(Error::graphql_error(errors));
+        let result = response.execute_transaction_block;
+        let bcs = base64ct::Base64::decode_vec(result.effects.bcs.0.as_str())?;
+        let effects: TransactionEffects = bcs::from_bytes(&bcs)?;
+
+        if let Some(wait_for) = wait_for {
+            self.wait_for_tx(tx.digest(), wait_for, None).await?;
```
thibault-martinez, line 1302: Just checking, the effects can't change depending on what we are waiting for? Because we parsed them before waiting

thibault-martinez: Nah probably not, just thinking out loud

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1407,80 +1245,126 @@ impl Client {
...
+        digest: Digest,
+        wait_for: WaitForTx,
+        timeout: impl Into<Option<Duration>>,
+    ) -> Result<()> {
+        tokio::time::timeout(
+            timeout.into().unwrap_or_else(|| Duration::from_secs(60)),
+            async {
+                let mut interval = tokio::time::interval(tokio::time::Duration::from_millis(100));
+                loop {
+                    interval.tick().await;
```
thibault-martinez, line 1356: Should we do a first check before ticking? Maybe it's already indexed/finalized

thibault-martinez (approved):

crates/iota-sdk/examples/gas_station.rs
```diff
@@ -27,11 +27,9 @@ async fn main() -> Result<()> {
             HeaderValue::from_str(&format!("Bearer {gas_station_auth_token}"))?,
         );
 
-    let res = builder.execute(&keypair.into(), true).await?;
+    let effects = builder.execute(&keypair.into(), None).await?;
```
Alex6323, line 30: the only design flaw IMHO of Rust is that it doesn't allow for named arguments. I hate reading function calls, that have `true` or `None` in it.

crates/iota-sdk-graphql-client/src/lib.rs
DaughterOfMars: Yeah I don't think so

crates/iota-sdk-graphql-client/src/lib.rs
DaughterOfMars: The first tick completes immediately

crates/iota-sdk/examples/gas_station.rs
DaughterOfMars: yeah :c

crates/iota-sdk-graphql-client/src/lib.rs
thibault-martinez: Oh ok, nice

PR #363, opened 2025-11-03
chore(python): make examples fail the CI

thibault-martinez:
Remove the try/exception as usage was inconsistent and without an additional `sys.exit(1)` would prevent error code from reaching the shell and failing the CI in case of error. It is not needed in the first place and the current way of doing also removes the stacktrace.


bindings/python/examples/generic_move_function.py
```diff
@@ -1,10 +1,9 @@
 # Copyright (c) 2025 IOTA Stiftung
 # SPDX-License-Identifier: Apache-2.0
 
-import asyncio
-
 from lib.iota_sdk_ffi import *
 
+import asyncio
 
 async def main():
```
Thoralf-M, line 8: Should we also have try: here?

thibault-martinez: Idk, lots of examples don't have one. What's the logic of adding it or not in Python?

Thoralf-M: Just a way to catch the error, we could also remove it from all the examples if the `sys.exit(1)` is then not needed anymore

thibault-martinez: I tried locally and it's actually better without the try, we get a full stacktrace and it does return an error code, I will remove them

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/363/commits/d37856637833eb8a2566df60706b8a9da2849ce2

Thoralf-M (approved):

DaughterOfMars (approved):

PR #364, opened 2025-11-03
feat(FFI): Add offline transaction builder to FFI

### LINKED ISSUE #304: Add offline transaction builder to FFI
Currently the FFI transaction builder only works with the client. We should add the ability to call the builder without the client like in rust.
Should also add the example from https://github.com/iotaledger/iota-rust-sdk/pull/301 in the bindings then

DaughterOfMars:
## Description

Adds an offline transaction builder type to the FFI bindings.

Closes #304


crates/iota-sdk-ffi/src/transaction_builder/client_builder.rs
```diff
@@ -0,0 +1,383 @@
...
+        move_package::MovePackageData,
+        object::ObjectId,
+        struct_tag::Identifier,
+        transaction::{Argument, Transaction, TransactionEffects},
+        type_tag::TypeTag,
+    },
+};
+
+/// A builder for creating transactions. Use `finish` to finalize the
+/// transaction data.
```
Thoralf-M: Probably good to mention the client here and that it makes it easier, otherwise from looking at the doc comments  both tx builders are the same

crates/iota-sdk-ffi/src/transaction_builder/client_builder.rs
```diff
@@ -0,0 +1,383 @@
...
+    {
+        let mut lock = self.0.write().expect("error writing to builder");
+        f(&mut lock)
+    }
+}
+
+#[uniffi::export(async_runtime = "tokio")]
+impl ClientTransactionBuilder {
+    /// Create a new transaction builder and initialize its elements to default.
+    #[uniffi::constructor(name = "init")]
```
Alex6323: Can we remove the rename to `init` here as well?

DaughterOfMars: No, this method is async which is why it's needed. Though tbh I was considering removing this method completely

bindings/python/examples/unstake.py
```diff
@@ -17,7 +17,7 @@ async def main():
             raise Exception("no staked iotas found")
         staked_iota = staked_iotas.data[0]
 
-        builder = await TransactionBuilder.init(
+        builder = await ClientTransactionBuilder.init(
```
Thoralf-M: Why use `ClientTransactionBuilder.init()` in some examples and `TransactionBuilder(sender).with_client(client)` in others?

bindings/kotlin/examples/PrepareSendCoins.kt
```diff
@@ -24,7 +24,7 @@ fun main() = runBlocking {
                         "0x8ef4259fa2a3499826fa4b8aebeb1d8e478cf5397d05361c96438940b43d28c9"
                 )
 
-        val builder = TransactionBuilder.init(fromAddress, client)
+        val builder = TransactionBuilder(fromAddress).withClient(client)
```
thibault-martinez, line 27: Why is it `NewTransactionBuilder` in go, but `TransactionBuilder` in kotlin and python?

thibault-martinez: Actually I see a mix of `ClientTransactionBuilder.init(myAddress, client)` and `TransactionBuilder(sender).with_client(client)` this is a bit confusing

crates/iota-sdk-ffi/src/transaction_builder/client_builder.rs
thibault-martinez: How does async relate to the method name?

DaughterOfMars: Bindings like golang create constructors differently from other methods. Like `sdk.NewTransactionBuilder` vs `sdk.TransactionBuilderNew` and this breaks with async constructors.

bindings/python/examples/unstake.py
DaughterOfMars: I missed some I guess

bindings/kotlin/examples/PrepareSendCoins.kt
DaughterOfMars: Yeah I missed a few. Should be fixed now.

Thoralf-M (approved):

bindings/kotlin/examples/GasStation.kt
```diff
@@ -13,7 +13,7 @@ fun main() = runBlocking {
         var sender = keypair.publicKey().deriveAddress()
         var simpleKey = SimpleKeypair.fromEd25519(keypair)
 
-        val builder = TransactionBuilder.init(sender, client)
+        val builder = TransactionBuilder(sender).withClient(client)
```
thibault-martinez, line 16: How come we don't have to call the `new` method? Or why is it not appended to the name?

DaughterOfMars: kotlin creates a constructor not a new fn

bindings/kotlin/examples/PrepareTransferObjectsOffline.kt
```diff
@@ -0,0 +1,70 @@
...
+                        ObjectId.fromHex(
+                                "0x0b0270ee9d27da0db09651e5f7338dfa32c7ee6441ccefa1f6e305735bcfc7ab"
+                        ),
+                        ObjectId.fromHex(
+                                "0x8ef4259fa2a3499826fa4b8aebeb1d8e478cf5397d05361c96438940b43d28c9"
+                        )
+                )
+        val objsToTransfer =
+                objects.map {
+                    val obj = client.`object`(it)
```
thibault-martinez, line 33: Is `object` a keyword?

DaughterOfMars: yep

thibault-martinez: Do we think that's nice to use like that?

DaughterOfMars: no, but what would you prefer?

thibault-martinez: Can you investigate if there is a way to rename a method conditionally on the uniffi backend? I'd rename this to `get_object` for Kotlin I guess. If ever this is doable, another PR.

bindings/python/examples/prepare_transfer_objects_offline.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2025 IOTA Stiftung
+# SPDX-License-Identifier: Apache-2.0
+
+from lib.iota_sdk_ffi import *
+
+import asyncio
+
+
+async def main():
+    try:
```
thibault-martinez: remove the try/exception please

bindings/kotlin/examples/PrepareTransferObjectsOffline.kt
DaughterOfMars: I'm pretty sure there is not

thibault-martinez: What about a `get_object` alias in the FFI? We can just not use it in python/go but at least we have an alternative in Kotlin

DaughterOfMars: Sure but it's really weird and confusing. I assume that kotlin users know their language is shit and how to avoid pitfalls like this, just like how rust users know they can't use fields named `type`

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -869,6 +869,26 @@ impl<L> TransactionBuilder<(), L> {
         }
         .into())
     }
+
+    /// Execute the transaction using the gas station and return the JSON
+    /// transaction effects. This will fail unless data is set with
+    /// [`Self::gas_station_sponsor`].
+    ///
+    /// NOTE: These effects are not necessarily compatible with
+    /// [`TransactionEffects`]
```
thibault-martinez, line 878: Is this why it returns JSON? Why exactly are they not compatible?

crates/iota-sdk-ffi/src/transaction_builder/builder.rs
```diff
@@ -0,0 +1,349 @@
...
+        });
+        self
+    }
+
+    /// Convert this builder into a transaction.
+    pub fn finish(&self) -> Result<Transaction> {
+        Ok(Transaction(self.read(|builder| builder.clone().finish())?))
+    }
+
+    pub async fn execute_with_gas_station(
```
thibault-martinez, line 347: Maybe a lil doc comment?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: They simply don't have all of the same fields

crates/iota-sdk-ffi/src/transaction_builder/client_builder.rs
thibault-martinez, line 1: No way to avoid the code duplication between both builders?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: I feel like we have 2 types representing effects, none of them matches whatever this is?

crates/iota-sdk-ffi/src/transaction_builder/client_builder.rs
DaughterOfMars: none whatever

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: Not sure which the second one would be

crates/iota-sdk-ffi/src/transaction_builder/client_builder.rs
thibault-martinez: Most definitely doable with a macro

thibault-martinez (approved):

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: I am taking a note for later as I don't want to prevent merging this longer but I don't think we should be returning JSON

bindings/kotlin/examples/PrepareTransferObjectsOffline.kt
thibault-martinez: I don't think it's weird nor confusing. Can do it later though, taking a note.

crates/iota-sdk-ffi/src/transaction_builder/client_builder.rs
DaughterOfMars: ewww

DaughterOfMars: actually I can't because of the uniffi macros

PR #367, opened 2025-11-03
fix(CI): Make pre-publish CI create a PR

DaughterOfMars:
## Description

- Converts the string entry into a dropdown for the pre-publish version bump.
- Sets all versions to 0.0.0 to prepare for publish (except iota-sdk)
- Sets the pre-publish commands to actually execute
- Modifies the pre-publish CI so that it creates a PR instead of only a tag
- Makes the github release trigger on the PR merge


Alex6323 (approved):

.github/workflows/pre_publish.yml
```diff
@@ -4,19 +4,18 @@ on:
...
-          - beta:    Increase the beta pre-version (x.y.z-beta.M)
-          - alpha:   Increase the alpha pre-version (x.y.z-alpha.M)
+        options:
+          - major
+          - minor
+          - patch
+          - release
+          - rc
+          - beta
+          - alpha
```
thibault-martinez, line 31: I think it would still be nice somehow to keep the doc

DaughterOfMars: I can't find any good way to do it :c

thibault-martinez: Better to have an unformatted blob than an ambiguous and potentially dangerous choice

crates/iota-sdk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "iota-sdk"
-version = "3.0.0"
+version = "2.0.0"
```
thibault-martinez: Should probably stay at 3

crates/iota-sdk-types/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "iota-sdk-types"
-version = "0.0.1"
+version = "0.0.1-alpha.0"
```
thibault-martinez, line 3: Thinking out loud: is it common to have alpha before major 1 ? Since major 0 doesn't give any stability guarantee?

release.toml
```diff
@@ -0,0 +1,4 @@
+enable-all-features = true
+pre-release-hook = [
+  "../../gen-changelog.sh",
```
thibault-martinez: Is this a list of commands? Can't we put the 2 commands of the script instead of having another file?

Thoralf-M: Path is correct like this, not `"./gen-changelog.sh",`?

crates/iota-sdk-types/Cargo.toml
DaughterOfMars: I don't know. I don't really see the suffixes used that much tbh

release.toml
DaughterOfMars: I tried that, but it's actually only one command. The code builds a single [`std::process::Command`](https://doc.rust-lang.org/std/process/struct.Command.html) and executes it, which will fail if multiple commands are provided.

DaughterOfMars: Yes, because the working directory is actually in the published crate itself

Thoralf-M (approved):

.github/gen-changelog.sh
thibault-martinez, line 1: Can you move these 2 files to the workflow folder please? Annoying to clutter the root since this will never be used manually

crates/iota-sdk-ffi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "iota-sdk-ffi"
-version = "0.0.1"
+version = "0.0.1-alpha.0"
```
thibault-martinez: Does not need to be versioned ?

.github/gen-changelog.sh
DaughterOfMars: sure

crates/iota-sdk-ffi/Cargo.toml
DaughterOfMars: what do you mean?

.github/gen-changelog.sh
DaughterOfMars: Actually, I can't move the `release.toml` because of a limitation of the github action :c

crates/iota-sdk-ffi/Cargo.toml
thibault-martinez: We are not publishing this crate so let's not make it alpha?

DaughterOfMars: ohhhh

.github/gen-changelog.sh
thibault-martinez: schade

thibault-martinez (approved):

PR #368, opened 2025-11-04
feat(graphql-client): update query types

Thoralf-M:
Added missing types/fields that are defined in the graphql schema but were missing in the query types. Only the validator fields I left commented out as these would make the query too big and make it fail: `Query error: [Query payload is too large. The maximum allowed is 5000 bytes]`
Also changed the iota_total_supply type to work, as the number it holds is larger than i32::MAX and decoding failed without this change

### DIFF
--- a/bindings/go/iota_sdk_ffi/iota_sdk_ffi.go
+++ b/bindings/go/iota_sdk_ffi/iota_sdk_ffi.go
@@ -33344,10 +33344,6 @@ func (_ FfiDestroyerValidatorPage) Destroy(value ValidatorPage) {
 	value.Destroy()
 }
 type ValidatorSet struct {
-	// The current set of active validators.
-	ActiveValidators ValidatorConnection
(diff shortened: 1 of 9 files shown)


DaughterOfMars:
Considering that users can create their own queries and call `run_query` or `run_query_from_json`, do we really want to do this? Especially considering that queries can reach a byte limit, I feel that this is unnecessary.

Thoralf-M:
> Considering that users can create their own queries and call `run_query` or `run_query_from_json`, do we really want to do this? Especially considering that queries can reach a byte limit, I feel that this is unnecessary.

TBH I'm also not sure about it, but the query now still has the size that is fine and this way more data is available to the users (also the structs in the crate are a good example one can copy paste and modify). Building custom queries makes sense to make the queries more efficient, but is also a bit harder/more annoying to do

Thoralf-M:
Custom queries is the way to go

PR #373, opened 2025-11-05
feat: add json_query examples

Thoralf-M:


crates/iota-sdk/examples/json_query.rs
```diff
@@ -0,0 +1,147 @@
...
+use eyre::Result;
+use iota_graphql_client::Client;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let client = Client::new_devnet();
+
+    let json = serde_json::json!({
+        "query": r#"
+query getLatestIotaSystemState {
```
DaughterOfMars: I'm guessing it fails with indents?

Thoralf-M: Actually works fine, changed https://github.com/iotaledger/iota-rust-sdk/pull/373/commits/f6aaded5d42806c1df77981cd0cc539aff7dd6dd

crates/iota-sdk/examples/json_query.rs
```diff
@@ -0,0 +1,147 @@
...
+            atRisk
+            reportRecords {
+              nodes {
+                address
+              }
+            }
+            apy
+          }
+        "#,
+        "variables": {}
```
thibault-martinez, line 137: Is this from `serde_json::json`? Is it required?

bindings/kotlin/examples/JsonQuery.kt
```diff
@@ -0,0 +1,140 @@
...
+          nextEpochCommissionRate
+          atRisk
+          reportRecords {
+            nodes {
+              address
+            }
+          }
+          apy
+        }
+    """.trimIndent()
```
thibault-martinez: Would not work without?

crates/iota-sdk/examples/json_query.rs
Thoralf-M: The variables? They are for graphql in case you want to provide a parameter in the query like epoch id or an address

thibault-martinez (approved):

crates/iota-sdk/examples/json_query.rs
thibault-martinez: How come we have it in rust and not bindings?

bindings/kotlin/examples/JsonQuery.kt
Thoralf-M: Works also without, just cleans up the string, but removed https://github.com/iotaledger/iota-rust-sdk/pull/373/commits/10940c9b44114b41d63edf08355d6dc0b571e024

crates/iota-sdk/examples/json_query.rs
Thoralf-M: In Rust we have `run_query_from_json()` where you provide the full query yourself, including the variables, so I thought it would be good to show there, however in the bindings we have just `run_query()` which accepts the `Query` type that has this as optional field, so there it's more obvious how variables need to be added

thibault-martinez: Thanks

DaughterOfMars (approved):

PR #382, opened 2025-11-05
chore(ci): Allow manual release and fix branch checking

DaughterOfMars:
## Description

Allows manually calling the release CI as well as fixes the branch checking.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ permissions:
 
 jobs:
   create-release:
-    if: github.event.pull_request.merged == true && startsWith(github.base_ref, 'release/')
+    if: github.event.pull_request == null || (github.event.pull_request.merged == true && startsWith(github.base_ref, 'release/'))
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,13 +1,25 @@
 name: Release
 
 on:
+  workflow_dispatch:
+    inputs:
+      package:
+        type: choice
+        required: true
+        description: The package to publish.
+        options:
+          - iota-sdk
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,33 +1,55 @@
...
+        type: string
+        description: The exact version.
   pull_request:
     types: closed
-    branches:
-      - "release/*/*.*.*-alpha.*"
-      - "release/*/*.*.*-beta.*"
-      - "release/*/*.*.*-rc.*"
-      - "release/*/*.*.*"
+    branches: [develop]
```
thibault-martinez, line 22: I feel like this should be possible from GH actions... Maybe another trigger than `pull_request`?

Thoralf-M (approved):

.github/workflows/release.yml
DaughterOfMars: nope, it's not :c

.github/workflows/release.yml
```diff
@@ -1,33 +1,55 @@
...
 
 # We need this to be able to create releases.
 permissions:
   contents: write
 
 jobs:
   create-release:
-    if: github.event.pull_request.merged == true
+    if: github.event.pull_request == null || (github.event.pull_request.merged == true && startsWith(github.base_ref, 'release/'))
```
thibault-martinez, line 30: `release/` will be dangerous when we have bindings

thibault-martinez (approved):

.github/workflows/release.yml
DaughterOfMars: I'm not sure it will. Seems to me we could use this workflow for those too

thibault-martinez: Feel free to merge, we'll adapt when we have the other workflows

PR #386, opened 2025-11-06
chore(ci): Set owner of published packages

DaughterOfMars:
## Description

Set the owner of published packages.

### DIFF
--- a/.github/workflows/pre_publish.yml
+++ b/.github/workflows/pre_publish.yml
@@ -17,7 +17,6 @@ on:
           - rc:      Increase the rc pre-version (x.y.z-rc.M)
           - beta:    Increase the beta pre-version (x.y.z-beta.M)
           - alpha:   Increase the alpha pre-version (x.y.z-alpha.M)
-          - no-bump: Release the packages with the current version
         options:
           - major
           - minor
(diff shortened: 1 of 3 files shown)


.github/workflows/pre_publish.yml
```diff
@@ -82,7 +78,7 @@ jobs:
 
       - name: Add release notes
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
```
thibault-martinez: But this is not a GH token, why would this work? Completely unrelated

DaughterOfMars: Isn't it a GH PAT?

DaughterOfMars: I reverted it

Thoralf-M (approved):

release.toml
```diff
@@ -2,3 +2,5 @@ enable-all-features = true
 pre-release-hook = [
   "../../.github/gen-changelog.sh",
 ]
+owners = ["github:iotaledger:dev-tools"]
```
thibault-martinez, line 5: Hoping this is adding and not overwriting 😬

thibault-martinez (approved):

release.toml
DaughterOfMars: should be

PR #391, opened 2025-11-07
feat(ci): add Python binding release workflow

### LINKED ISSUE #243: Create workflow to publish Python package to pypi

Alex6323:
Closes #243 

This PR is not as straight-forward as it may seem, because of the following difficulties:
* a decision had to be made between `maturin` or `setuptools-rust` to create the platform specific Python packages, i.e. the wheels (`.whl`); both have downsides and I picked `maturin` + a small hack (more on that later) over the other, which is the way more complex approach. `maturin` supports `uniffi` out-of-the-box, so it places the generated `iota-sdk-ffi.py` and the `libiota-sdk-ffi.{LIB_EXT}` the way it's expected by `uniffi`. That means however that with `maturin` we can't use our `make` build script, because it is designed to do everything from start (building) to finish (packaging, ~~uploading~~)
* In order for `maturin` to produce a valid `iota-sdk-ffi.py` we need to install the exact same version of the `uniffi-bindgen` CLI which currently is at `0.29` in the `iota-sdk-ffi` crate deps. `maturin` looks for a binary named `uniffi-bindgen` in the `PATH` env var, so luckily we can control which binary it should use. We can't use a `pipx install uniffi-bindgen` in the CI because the latest version installable through `pip` is `0.28.3` (which tests showed is incompatible), which is why we need to build and `cargo install` that binary in the CI from the repo, which will always make sure that the versions correctly match. I had to rename the binary to `uniffi-bindgen` in the `iota-sdk-ffi` crate to make that work.
 
UPDATE: There was a naming incompatibility issue that was fixed upstream (with `maturin` 1.10), so the initially required `repair_whl.sh` script is no longer needed! :partying_face: FYI here's where they fixed it: https://github.com/PyO3/maturin/pull/2840/files#diff-ef3401ac6425fac36616570d22b89962c2c4c9b5033f849ce8e0b6ab904c9fe9L1186. That means that we must set the minimum maturin version to `1.10` in the `pyproject.toml`.

Other than that this PR can be tested locally easily with the following commands:

1. `pipx install maturin` (cargo install maturin fails to compile unless you install their latest main)
2. `cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen`
3. `cd bindings/python/src`
4. `maturin build --release` (you'll find the generated `.whl` in target/wheels)
5. create a python test project with a virtual environment
6. `pip install target/wheels/iota_sdk-0.1.0-<PLATFORM_SPECIFIC>.whl`
7. finally paste and run:

```
from iota_sdk import *

import asyncio

async def main():

    chain_id = await client.chain_id()
    print("Chain ID:", chain_id)

if __name__ == "__main__":
    asyncio.run(main())
```

The published Python bindings can be found on TestPyPi: https://test.pypi.org/project/iota-sdk/3.0.0a1.dev16/

Version like `3.0.0-alpha.1` are converted to `3.0.0a1` according to PEP-440 standard: https://peps.python.org/pep-0440/#public-version-identifiers


DaughterOfMars:
Since we're building maturin from source anyway, could we instead fork it and make some tweaks to suit our purposes?

Alex6323:
> Since we're building maturin from source anyway, could we instead fork it and make some tweaks to suit our purposes?

~~sorry, I think I had a mistakte in the PR description. We're not building `maturin`~~ since we're using the action, we actually do build it from source. Locally I installed a prebuilt binary. I wouldn't think it's worth the effort to fork it and then having to maintain the fork. Eventually they might fix the issue anyways, and we can just remove the repair script and the CI step then.

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,170 @@
...
+permissions:
+  contents: read
+
+jobs:
+  build-wheels:
+    name: Build platform specific wheels
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
```
Thoralf-M: What about the arm targets?

bindings/python/src/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[build-system]
+requires = ["maturin>=1.6,<2.0"]
+build-backend = "maturin"
+
+[project]
+name = "iota-sdk"
+version = "0.1.0"
+description = "IOTA SDK Python bindings"
+readme = "README.md"
+license = { file = "LICENSE" }
```
Thoralf-M: Can't we just link to the root LICENSE file?

.github/workflows/pypi_publish.yml
Alex6323: there was a compilation error, that doesn't seem trivial, which is why I removed them for now. Needs more time to fix this. I can create an issue and address it specifically.

bindings/python/src/pyproject.toml
Alex6323: we probably can

.github/workflows/pypi_publish.yml
Thoralf-M: Fine for me, but we definitely want to have them supported

bindings/python/src/pyproject.toml
Alex6323: not sure atm it's possible:
```
💥 maturin failed
  Caused by: Failed to build source distribution
  Caused by: Failed to add file from /home/me/dev/iota/iota3-rebased/iota-rust-sdk_publish-python-bindings/bindings/python/src/../../../LICENSE to sdist as iota_sdk-0.1.0/../../../LICENSE
  Caused by: paths in archives must not have `..` when setting path for iota_sdk-0.1.0
```

Thoralf-M: Ok then fine for me to keep it

.github/workflows/pypi_publish.yml
Alex6323: yeah, I investigated a little and pushed a potential fix, let's wait for the CI

Alex6323: fix worked: https://github.com/iotaledger/iota-rust-sdk/actions/runs/19261929481

bindings/python/src/pyproject.toml
DaughterOfMars: can we `cp` it in for the build or something?

.github/workflows/pypi_publish.yml
Thoralf-M: Nice

bindings/python/src/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[build-system]
+requires = ["maturin>=1.10,<2.0"]
+build-backend = "maturin"
+
+[project]
+name = "iota-sdk"
+version = "0.1.0"
```
Thoralf-M: Should start with `3.0.0.alpha.1`?
https://pypi.org/project/iota-sdk/#history

bindings/python/src/README.md
Thoralf-M, line 1: Why have this in src/ ?

Alex6323: that's because initially the generated python landed in there, but when I switched to `uniffi` in the `pyproject.toml` the python code started to get compiled into the wheel directly, so now it looks illogical.
It's unlikely, but if we ever switch back to another method, there will be `.py` files in there. However, I'ld also prefer renaming it. Suggestion? `pkg`?

bindings/python/src/pyproject.toml
Alex6323: yeah, will update with the other comment

bindings/python/src/README.md
Thoralf-M: Sorry I wasn't clear, the name `src` is fine for me, I'm just wondering why we have this README.md in src/ and another one in python/, shouldn't we have only one?

Alex6323: This README here is the one that gets displayed at the SDK's PyPi front page. The instructions are different (i.e. pip instructions). Mixing it with the instructions from the other README (ie. make commands) would confuse users I think.

Alex6323: I am glad you linked the old SDKs page though. We can re-use that.

Thoralf-M: But the example to show how to get the chain id would be good?

bindings/python/src/pyproject.toml
Alex6323: done

bindings/python/src/README.md
Alex6323: added

.github/workflows/bindings.yml
```diff
@@ -17,6 +17,7 @@ concurrency:
 jobs:
   go:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
```
thibault-martinez: remove?

bindings/python/README.md
```diff
@@ -11,14 +15,32 @@ Please follow the general install instructions for your platform.
...
-## Generate Python bindings
+## Generate Python bindings module `iota_sdk_ffi.py`
+
+1. Build the bindings: `make python`
+2. Test by running the following minimal example: `make python-example chain_id`
+
+## Generate Python bindings wheel `iota_sdk-*.whl`
+
+1. Install `maturin`: `pipx install maturin`
+2. Install `uniffi-bindgen`: `cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen`
```
thibault-martinez: Why do we need to install it? What other step relies on it being globally installed?

bindings/python/README.md
```diff
@@ -11,14 +15,32 @@ Please follow the general install instructions for your platform.
...
-## Run Python example
+       chain_id = await client.chain_id()
+       print("Chain ID:", chain_id)
 
-```sh
-make python-example chain_id
-```
+   if __name__ == "__main__":
+       asyncio.run(main())
+   ```
```
Thoralf-M: Remove the indentation?

bindings/python/README.md
thibault-martinez: I saw your README says it's maturin but how?

.github/workflows/bindings.yml
Alex6323: yeah, maybe, but do you think it's useful? Because I think this is an expensive workflow that shouldn't run on drafts (dev should built and run bindings examples locally)

But it could be a separate PR of course. Wdyt?

bindings/python/README.md
Alex6323: maturin looks for a `uniffi-bindgen` binary in the PATH

bindings/python/README.md
Alex6323: it belongs to step 7

bindings/python/README.md
```diff
@@ -11,14 +15,32 @@ Please follow the general install instructions for your platform.
...
+2. Test by running the following minimal example: `make python-example chain_id`
+
+## Generate Python bindings wheel `iota_sdk-*.whl`
+
+1. Install `maturin`: `pipx install maturin`
+2. Install `uniffi-bindgen`: `cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen`
+3. Switch to the Python bindings package: `cd bindings/python/src`
+4. Build the wheel: `maturin build --release` (you'll find the generated `.whl` in target/wheels)
+5. Create or switch to your Python project with an activated virtual environment
+6. Install the local wheel: `pip install target/wheels/iota_sdk-0.1.0-<PLATFORM_SPECIFIC>.whl`
```
Alex6323:
```suggestion
6. Install the local wheel: `pip install target/wheels/iota_sdk-<VERSION>-<PLATFORM>.whl`
```

.github/workflows/bindings.yml
thibault-martinez: Tbh I kinda misread this line at first. Ideally we'd be doing the same thing as the monorepo with draft + `[ci-run]` so  let's remove from this PR and I'll open an issue to do it properly.

bindings/python/README.md
Alex6323: if you look at the preview in VSCode it looks correct with the indentation.

.github/workflows/bindings.yml
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/issues/412

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,301 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to publish (leave empty for release version)"
+        required: false
+        type: string
+      dry_run:
+        description: "Dry run - publish to Test-PyPi repo only"
```
thibault-martinez: I don't know if I would call that a dry run

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,301 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform:
+          - target: x86_64
+          - target: i686
+          - target: aarch64
+          - target: armv7l
+          - target: ppc64le
+    # - target: s390x # fails here: crates/iota-sdk-types/src/u256.rs:51:5
```
thibault-martinez: Please open an issue 🙏🏻

.github/workflows/bindings.yml
Alex6323:
```suggestion
```

.github/workflows/bindings.yml
```diff
@@ -48,6 +49,7 @@ jobs:
         run: make go-examples
   kotlin:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
```
Alex6323:
```suggestion
```

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,301 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
thibault-martinez: Please pin the commit of all actions like the rest of the workflows (don't forget the version comment)

.github/workflows/bindings.yml
```diff
@@ -81,6 +83,7 @@ jobs:
         run: make kotlin-examples
   python:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
```
Alex6323:
```suggestion
```

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,301 @@
...
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install `uniffi-bindgen` for maturin
+        run: |
+          cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen
```
thibault-martinez: Please put the one liners directly on the same line

bindings/python/README.md
thibault-martinez: We don't usually ident code snippets like that though?

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,301 @@
...
```
Alex6323:
```suggestion
      - name: Install `uniffi-bindgen` for maturin
        run: |
          cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen
          
```

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,301 @@
...
+      - name: Update version in pyproject.toml
+        if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.version != '' }}
+        run: |
+          FILE="bindings/python/src/pyproject.toml"
+          grep -q '^version = ".*"' "$FILE" || { echo "version line missing"; exit 1; }
+          NEW_VERSION=${{ github.event.inputs.version }}
+          echo "New version: $NEW_VERSION"
+          sed -i '' -E "s|^version = \".*\"|version = \"${NEW_VERSION}\"|" "$FILE"
+          echo "Updated pyproject.toml:"
+          cat "$FILE"
```
thibault-martinez: What's the process of actually updating the file in the repository then?

.github/workflows/tests.yml
```diff
@@ -21,6 +21,7 @@ env:
 jobs:
   tests:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
```
Alex6323:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -38,6 +39,7 @@ jobs:
 
   tests-with-network:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
```
Alex6323:
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -73,6 +75,7 @@ jobs:
 
   doctests:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
```
Alex6323:
```suggestion
```

.github/workflows/wasm.yml
```diff
@@ -20,6 +20,7 @@ env:
 jobs:
   wasm:
     runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
```
Alex6323:
```suggestion
```

bindings/python/src/pyproject.toml
Alex6323: sorry, didn't see you comment as it was resolved, but yeah, we could do that

Alex6323: @thibault-martinez @Thoralf-M should we do what Chloé suggested?

.github/workflows/pypi_publish.yml
Alex6323: How do you want to call it? In my opinion it still fits the definition of a dry-run (a test that as closely as possible mimics the "real thing", but without actual consequences).

.github/workflows/pypi_publish.yml
thibault-martinez: What's changing here 👀 ?

bindings/python/src/pyproject.toml
thibault-martinez: Fine by me

.github/workflows/pypi_publish.yml
Alex6323: But I don't mind renaming it if you want.

.github/workflows/pypi_publish.yml
Alex6323: Honestly I think we should just remove it and spare us the effort? Seems not like a target regular people built for.

.github/workflows/pypi_publish.yml
Alex6323: good point

.github/workflows/pypi_publish.yml
Alex6323: Very good point. I also struggled with that question, so now's the time to clarify that. I assumed this was a requirement of these workflows to be able to manually override, so that we ensure a consistent release version. But in my opinion we should just fail the whole workflow here if the version in that file (pyproject.toml) doesn't match the workflow argument. So this step should be a safe-guard, that we have indeed committed the right version to the repo already that we're releasing for. Should I change it like that? 🤔

.github/workflows/pypi_publish.yml
thibault-martinez: Fine by me

.github/workflows/pypi_publish.yml
thibault-martinez: But it does have consequences, it's like saying sending a TX on the testnet was a dry run, it's not

.github/workflows/pypi_publish.yml
thibault-martinez: I think this should be doing what we do for Rust. The workflow dispatch takes a desired bump descriptor ("minor", "major", "patch", ...), read the version from the file, bump it, updates the file, do a PR to a update it. Having to do a PR ourselves before releasing is not very convenient.

bindings/python/README.md
Thoralf-M: Looks better in md view

.github/workflows/pypi_publish.yml
Alex6323: it's adding a blank line at 141, so each step is consistently separated

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,301 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install `uniffi-bindgen` for maturin
+        run: |
+          cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen
```
Alex6323:
```suggestion
      - name: Install `uniffi-bindgen` for maturin
        run: |
          cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen
          
```

.github/workflows/pypi_publish.yml
Alex6323:
```suggestion
        run: cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen
```

.github/workflows/pypi_publish.yml
Alex6323:
```suggestion
```

.github/workflows/pypi_publish.yml
Alex6323: no, it's different, you can delete previous pushes to TestPyPi, which you can't on PyPi. So you can completely reverse it.

thibault-martinez: Still doesn't make it a dry run, I'm pretty sure there is even a flag to actually dry run Test-PyPi.

bindings/python/src/pyproject.toml
Alex6323: ah, sorry, just realized we can't do that what she suggested, because then local `maturin build` fails, and there's no easiy way (if at all) to add a "pre-run" hook to copy the LICENSE where it expects it.

.github/workflows/pypi_publish.yml
Alex6323: I don't mind how we call it tbh (it was mainly to make it consistent with the Kotlin PR). What about `Test Run` then?

Thoralf-M: test_publish
Publish to Test-Pypi

Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/391/commits/4093ecaf9640def605a185f6b83fcc84e44d6edd

.github/workflows/pypi_publish.yml
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/391/commits/4093ecaf9640def605a185f6b83fcc84e44d6edd

.github/workflows/pypi_publish.yml
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/391/commits/4093ecaf9640def605a185f6b83fcc84e44d6edd

.github/workflows/pypi_publish.yml
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/391/commits/374b529dff7905874708f8e1626f49e3b2d4aea8

Thoralf-M (dismissed):

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,301 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install `uniffi-bindgen` for maturin
+        run: cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen
+      - name: Install Python
+        uses: actions/setup-python@v5
```
thibault-martinez: Many more to pin

thibault-martinez: Oh you did it just now :)

bindings/python/src/README.md
```diff
@@ -0,0 +1,30 @@
+# IOTA SDK Library - Python bindings
+
+This library provides Python bindings for the official [`IOTA SDK`](https://github.com/iotaledger/iota-rust-sdk).
+
+Note that these bindings were automatically generated using [`uniffi-rs`](https://github.com/mozilla/uniffi-rs) and [`maturin`](https://github.com/PyO3/maturin) and are incompatible with any IOTA SDK version prior `v3`.
```
thibault-martinez:
```suggestion
Note that these bindings were automatically generated using [`uniffi-rs`](https://github.com/mozilla/uniffi-rs).
```
I wouldn't say that, it's assumed by major versions

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,310 @@
...
+
+      - name: Install `uniffi-bindgen` for maturin
+        run: cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.9
+
+      - name: Update version in pyproject.toml
```
thibault-martinez: There is lots of duplicated code, maybe you can factorise with a better matrix. I factorised my workflow a lot after seeing Thoralf's matrix.

Alex6323: I started out with Thoralf's workflow but actually I found it clearer to have platform-specific jobs. That's because for each platform you use a different runner, and that may have an influence on which steps you can/have to run. I saw in your workflow, that you disable certain steps depending on the platform. But that makes the job harder to follow IMHO. The second reason is that I looked at many examples for python release workflows which is why I followed their lead. If you have `maturin` installed locally, please run `maturin generate-ci github` to see what I mean. That's essentially the template that most people follow. 

Also consider platform dependendent intricacies:
1. on MacOS you have to use `sed` differently than on Linux: `sed -i ...` vs `sed -i '' ...`
2. on Windows the default shell is `PowerShell`, so for platform independence I would have add `shell: bash`, which is redundant on Mac and Linux.

Alex6323: I looked at the examples linked here: https://github.com/PyO3/maturin-action

thibault-martinez: I respectfully disagree, the workflow is currently very heavy to process because of the repetition and maintenance is going to be pretty annoying.

> on MacOS you have to use sed differently than on Linux: sed -i ... vs sed -i '' ...

Not an issue, we're already using a platform independent `sed - i` in scripts and CI, just need to tweak it a bit

> on Windows the default shell is PowerShell, so for platform independence I would have add shell: bash, which is redundant on Mac and Linux.

Doesn't sound like a problem to me

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,310 @@
...
+        uses: dtolnay/rust-toolchain@0f44b27771c32bda9f458f75a1e241b09791b331
+        with:
+          toolchain: stable #1.91.1
+
+      - name: Install `uniffi-bindgen` for maturin
+        run: |
+          cargo install --path crates/iota-sdk-ffi --bin uniffi-bindgen
+
+      - name: Install Python
+        uses: actions/setup-python@v5
```
thibault-martinez: 👀

.github/workflows/python_publish.yml
```diff
@@ -0,0 +1,310 @@
...
+          FILE="bindings/python/src/pyproject.toml"
+          grep -q '^version = ".*"' "$FILE" || { echo "version line missing"; exit 1; }
+          NEW_VERSION=${{ github.event.inputs.version }}
+          echo "New version: $NEW_VERSION"
+          sed -i -E "s|^version = \".*\"|version = \"${NEW_VERSION}\"|" "$FILE"
+          echo "Updated pyproject.toml:"
+          cat "$FILE"
+
+      - name: Build wheel
+        uses: PyO3/maturin-action@v1
```
thibault-martinez: 👀

.github/workflows/pypi_publish.yml
```diff
@@ -0,0 +1,310 @@
...
```
thibault-martinez: 👀

thibault-martinez: Lots of others

thibault-martinez: Because of the repetition :trollface:

Alex6323: repeated things are now factored out

.github/workflows/python_publish.yml
Alex6323: fixed

.github/workflows/pypi_publish.yml
Alex6323: fixed

.github/workflows/pypi_publish.yml
Alex6323: So, I spent a little bit of time to find a compromise here, because IMHO conditional steps are also repetitive, which is why I factored out the repeated steps into an action and call it with different args instead.

thibault-martinez: Alright I won't push for it more but I honestly fail to see how juggling between 3 files is easier to read than a few `if`s, if anything it makes it worse IMHO

bindings/python/README.md
```diff
@@ -11,14 +15,32 @@ Please follow the general install instructions for your platform.
 
 Verify by running `make --version` and `python3 --version`.
 
-## Generate Python bindings
+## Generate Python bindings module `iota_sdk_ffi.py`
+
+1. Build the bindings: `make python`
+2. Test by running the following minimal example: `make python-example chain_id`
+
+## Generate Python bindings wheel `iota_sdk-*.whl`
```
thibault-martinez: I don't think we need to show this to the general public tbh, no one cares but us, it's confusing otherwise

thibault-martinez: We should show how to install the SDK from pip instead

thibault-martinez: Ah but we have 2 READMEs, tbh I would just link one to the other, no need for that stuff IMO, even for us actually, the CI will take care of it

bindings/python/src/pyproject.toml
thibault-martinez: Sorry what happens exactly if you have a step just before that copies the license?

.github/workflows/pypi_publish.yml
Alex6323: Okay, let me try to make an argument for this solution:

1. IMHO it scales a bit better than the `if` guarded steps, bc instead of adding a new step with a different `if`, repeating most of it and just passing some other variable, you essentially create callable functions (action.yml) with input arguments and return values (outputs). That makes it like any other function call where that function lives in another file (we do it everywhere in Rust)
2. `if` steps are not visually indented like regular code is, which makes it hard to quickly understand which parts run and which don't given a certain condition

Alex6323: Although I still stand behind this arguments, practically in this case it wasn't required, because the steps between the platforms (especially how you invoke the maturin action) were not that different, so even `if` guards weren't necessary.

bindings/python/README.md
Alex6323: removed

bindings/python/src/pyproject.toml
Alex6323: I can't remember why it didn't work when I tried it, but seems like there was something else wrong back then. Added the `copy` step.

bindings/python/src/README.md
```diff
@@ -0,0 +1,30 @@
...
+This library provides Python bindings for the official [`IOTA SDK`](https://github.com/iotaledger/iota-rust-sdk).
+
+Note that these bindings were automatically generated using [`uniffi-rs`](https://github.com/mozilla/uniffi-rs).
+
+## Installation
+
+To add this library as a dependency to your 3.9+ Python project, first make sure you have created a [`virtual environment`](https://docs.python.org/3/library/venv.html), then run the command:
+
+```sh
+pip install iota-sdk
```
Thoralf-M:
```suggestion
pip install iota-sdk==3.0.0a1
```
We should have the version here until a new stable release is published, otherwise users will get the stardust package

Alex6323: make sense

bindings/python/src/pyproject.toml
Alex6323: Can be resolved

.github/workflows/pypi_publish.yml
Alex6323: can be resolved

Thoralf-M (approved):

bindings/python/src/pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[build-system]
+requires = ["maturin>=1.10,<2.0"]
+build-backend = "maturin"
+
+[project]
+name = "iota-sdk"
+version = "3.0.0a1"
+description = "IOTA SDK Python bindings"
+readme = "README.md"
+license = { file = "LICENSE" }
```
thibault-martinez: You could have just put "Apache-2.0" btw, not even needing a file

.github/workflows/python_publish.yml
```diff
@@ -0,0 +1,180 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@0f44b27771c32bda9f458f75a1e241b09791b331
+        with:
+          toolchain: stable
+
+      - name: Install `uniffi-bindgen` for maturin
+        shell: bash
```
thibault-martinez: probably don't need shell for most of the steps

thibault-martinez (approved):

.github/workflows/python_publish.yml
Alex6323: this makes windows use bash as well instead of powershell. I agree that for this command that's likely the same on windows.

bindings/python/src/pyproject.toml
Alex6323: Didn't know that was possible. Thanks

Alex6323: done

.github/workflows/python_publish.yml
Alex6323: removed

PR #405, opened 2025-11-10
fix(workflows): fix CI

Thoralf-M:


Thoralf-M:
devnet indexer was just not reachable

PR #413, opened 2025-11-18
feat(ci): bindings release trigger

### LINKED ISSUE #241: CI: create release workflows

thibault-martinez:
Closes https://github.com/iotaledger/iota-rust-sdk/issues/241


.github/workflows/release.yml
```diff
@@ -50,6 +60,9 @@ jobs:
           echo "PKG=$pkg" >> $GITHUB_ENV
           echo "VERSION=$version" >> $GITHUB_ENV
 
+      # TODO why is it taking the whole changelog file?
+      # TODO handle bindings changelog
+      # TODO bring back notes "--notes-file "crates/$PKG/CHANGELOG.md"
```
Thoralf-M: `bring back notes "--notes-file "crates/$PKG/CHANGELOG.md"` why was it even removed?

thibault-martinez: Because the path was wrong for bindings

.github/workflows/go_publish.yml
```diff
@@ -128,30 +112,13 @@ jobs:
...
-
-      - name: Set version according to inputs
+      - name: Set version to input or release tag
         id: set-version
         run: |
           if [ -n "${{ inputs.version }}" ]; then
             echo "version=${{ inputs.version }}" >> $GITHUB_OUTPUT
           else
-            echo "version=${{ steps.bump-semver.outputs.new_version }}" >> $GITHUB_OUTPUT
+            echo "version=$(echo "${{github.event.release.tag_name}}" | sed -nr 's|.*-v(.*)|\1|p')" >> $GITHUB_OUTPUT
```
DaughterOfMars: Can we do this `sed` stuff in a separate command?

.github/workflows/pre_publish.yml
```diff
@@ -65,9 +71,62 @@ jobs:
         with:
           tool: git-cliff
 
+      # TODO does this actually work?
```
DaughterOfMars: It does sign it but github doesn't consider it verified :c

.github/workflows/release.yml
```diff
@@ -50,6 +60,9 @@ jobs:
           echo "PKG=$pkg" >> $GITHUB_ENV
           echo "VERSION=$version" >> $GITHUB_ENV
 
+      # TODO why is it taking the whole changelog file?
```
DaughterOfMars: We could instead generate the string and use that, then just append it to the file in a separate step

.github/workflows/pre_publish.yml
```diff
@@ -65,9 +71,62 @@ jobs:
...
+          echo "Updated pyproject.toml with new version: $NEW_VERSION_PEP440"
+
+          # Generate changelog for Python
+          ./.github/gen-changelog.sh bindings/python "$NEW_VERSION"
+
+          # Set output for the next step
+          echo "NEW_VERSION=$NEW_VERSION" >> $GITHUB_ENV
+
+      - name: Create PR for Rust
+        if: startsWith(inputs.package, 'iota-sdk')
```
DaughterOfMars, line 147: maybe move this to an ENV var? then we have this logic in one place at least

.github/workflows/go_publish.yml
thibault-martinez: Why?

.github/workflows/pre_publish.yml
thibault-martinez: What do you mean?

.github/workflows/go_publish.yml
DaughterOfMars: just a lot in line

.github/workflows/pre_publish.yml
DaughterOfMars: We could add at the top:
```
env:
  IS_SDK: ${{ startsWith(inputs.package, 'iota-sdk') }}
```

.github/workflows/release.yml
DaughterOfMars: still using a file

release.toml
```diff
@@ -1,6 +1,7 @@
 enable-all-features = true
 pre-release-hook = [
   "../../.github/gen_changelog.sh \"$CRATE_ROOT\" \"$NEW_VERSION\"",
+  "git add \"*CHANGELOG.md\"",
```
DaughterOfMars: I don't think this will work

thibault-martinez: Why not?

.github/workflows/release.yml
thibault-martinez: Whoops

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/413/commits/ee5d14a1990fc3f0a70ebf2707f7275691aed9c9

.github/gen_changelog.sh
```diff
@@ -0,0 +1,19 @@
...
+
+git fetch --tags
+
+LATEST_TAG=$(git tag -l "$PKG-v*" --sort=-v:refname | head -n 1)
+echo "Latest tag: $LATEST_TAG"
+
+COMMIT_FROM=$(git show-ref $LATEST_TAG | cut -f 1 -d' ')
+echo "Commit from: $COMMIT_FROM"
+
+COMMIT_TO=$(git show-ref $VERSION | cut -f 1 -d' ')
```
thibault-martinez: I think I messed up here, investigating

release.toml
DaughterOfMars: Because this only accepts a single command

thibault-martinez: Weird that it's an array but I trust you https://github.com/iotaledger/iota-rust-sdk/pull/413/commits/6919672b597e7a2509ecab4ad10517bb63eda6a2

.github/gen_changelog.sh
```diff
@@ -3,6 +3,7 @@
 PKG=$1
 PKG_ROOT=$2
 VERSION=$3
+ADD=$4
```
DaughterOfMars: can also default it to false

release.toml
DaughterOfMars: It's still there :c

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/413/commits/f8841606218ff07cd18da6585531578f98316ca9

.github/gen_changelog.sh
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/413/commits/c32766370e868a423a9391807a088f2fd752e53e

.github/gen_changelog.sh
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/413/commits/af4f4c91f82b478456668b598547238619326761

release.toml
```diff
@@ -1,6 +1,6 @@
 enable-all-features = true
 pre-release-hook = [
-  "../../.github/gen-changelog.sh",
+  "../../.github/gen_changelog.sh \"$CRATE_ROOT\" \"$NEW_VERSION\"",
```
DaughterOfMars: Needs more args

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/413/commits/dac1c4fda7c14868f026dda395e159c1e8d107c6

.github/workflows/pre_publish.yml
thibault-martinez: Think I prefer to keep it that way

DaughterOfMars (approved):

.github/workflows/rust_publish.yml
Alex6323, line 1: shouldn't this have a dispatch trigger like the others as well?

thibault-martinez: I don't think so, and tbh the others probably shouldn't either but I'll re-evaluate later

Alex6323: I thought we wanted them as a fail-safe.

thibault-martinez: We should trigger from pre_publish otherwise we don't have changelogs, version updates, releases etc and it creates a mess we manually need to fix. I'm not removing them anyway, it's very much possible we need them for whatever reason but I'd rather see the need first before adding it here as well.

Alex6323 (approved):

PR #418, opened 2025-11-19
feat(workflows): add Python part

Alex6323:


.github/bump_version.sh
```diff
@@ -119,6 +119,75 @@ handle_prerelease() {
     fi
 }
 
+semver_to_pep440() {
+    local version="$1"
+
+    # Remove leading 'v' if present
```
Thoralf-M: We only want to remove and not add it back the other way around?

.github/workflows/pre_publish.yml
```diff
@@ -80,12 +80,36 @@ jobs:
...
+
+          NEW_VERSION_PEP440=$(semver_to_pep440 "$NEW_VERSION")
+
+          # Update the version in pyproject.toml
+          sed -i "s/version = \"$CURRENT_VERSION_PEP440\"/version = \"$NEW_VERSION_PEP440\"/" "$CONFIG"
+          echo "Updated pyproject.toml with new version: $NEW_VERSION_PEP440"
+
+          # Generate changelog for Python
+          export CRATE_ROOT="bindings/python"
+          ./.github/gen-changelog.shpython
```
Thoralf-M:
```suggestion
          ./.github/gen-changelog.sh
```

.github/workflows/pre_publish.yml
```diff
@@ -80,12 +80,36 @@ jobs:
...
+          # Read current version from pyproject.toml
+          CONFIG="bindings/python/src/pyproject.toml"
+          CURRENT_VERSION_PEP440=$(grep '^version = ' $CONFIG | sed 's/version = "\(.*\)"/\1/')
+
+          source ./.github/bump_version.sh
+          CURRENT_VERSION=$(pep440_to_semver "$CURRENT_VERSION_PEP440")
+          echo "Current version: $CURRENT_VERSION"
+
+          # Bump the version using the input level
+          NEW_VERSION=$(bump_semver "$CURRENT_VERSION" "${{ inputs.level || 'alpha' }}")
```
Thoralf-M:
```suggestion
          NEW_VERSION=$(bump_semver "$CURRENT_VERSION" "${{ inputs.level }}")
```
did you want to keep this for a specific reason? In my branch I only added it for testing

Alex6323: no, I thought it's necessary and missed some information.

.github/bump_version.sh
Alex6323: I think we should remove that. AI just wanted to make it more resilient. It didn't know from the prompt that we run the script ourselves.

Thoralf-M: No problem with having this part, just wondered if we need to add the `v` before the version, but sounds like it's not needed then

Thoralf-M (approved):

Thoralf-M (approved):

thibault-martinez (approved):

PR #421, opened 2025-11-20
feat(ffi): Export more derivable traits

### LINKED ISSUE #322: Allow direct object comparison in bindings if types impl `Eq`
Many types can be compared for equality, which currently is not exposed to the bindings but can easily be done through:
`#[uniffi::export(Eq)]`. As an example in a Python binding example it was necessary to compare the `StructTag` of a type.
This was only possible so far by comparing the string (Display) representation, but that is not ideal.

Updated struct:
```
#[derive(derive_more::From, derive_more::Display, uniffi::Object, PartialEq)]
#[uniffi::export(Display, Eq)]
pub struct StructTag(pub iota_types::StructTag);
```

let's you then use `==` comparision directly on the types in the bindings language.

BLOCKED BY #433

### LINKED ISSUE #433: Kotlin/Java JNA limitation requires smaller bindings interface
While implementing https://github.com/iotaledger/iota-rust-sdk/pull/421 which exposes more standard traits like `Eq, Hash, Debug, Display` to the bindings, the Kotlin bindings CI failed with:

```sh
Error: Exception in thread "main" jdk.internal.org.objectweb.asm.MethodTooLargeException: Method too large: jdk/proxy2/$Proxy4.<clinit> ()V
	at java.base/jdk.internal.org.objectweb.asm.MethodWriter.computeMethodInfoSize(MethodWriter.java:2120)
	at java.base/jdk.internal.org.objectweb.asm.ClassWriter.toByteArray(ClassWriter.java:543)
	at java.base/java.lang.reflect.ProxyGenerator.generateClassFile(ProxyGenerator.java:506)
	at java.base/java.lang.reflect.ProxyGenerator.generateProxyClass(ProxyGenerator.java:178)
	at java.base/java.lang.reflect.Proxy$ProxyBuilder.defineProxyClass(Proxy.java:544)
	at java.base/java.lang.reflect.Proxy$ProxyBuilder.build(Proxy.java:657)
	at java.base/java.lang.reflect.Proxy.lambda$getProxyConstructor$0(Proxy.java:429)
	at java.base/jdk.internal.loader.AbstractClassLoaderValue$Memoizer.get(AbstractClassLoaderValue.java:329)
	at java.base/jdk.internal.loader.AbstractClassLoaderValue.computeIfAbsent(AbstractClassLoaderValue.java:205)
	at java.base/java.lang.reflect.Proxy.getProxyConstructor(Proxy.java:427)
	at java.base/java.lang.reflect.Proxy.newProxyInstance(Proxy.java:1034)
	at com.sun.jna.Native.load(Native.java:624)
	at com.sun.jna.Native.load(Native.java:596)
	at iota_sdk.UniffiLib$Companion.INSTANCE_delegate$lambda$0(iota_sdk_ffi.kt:2595)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:86)
	at iota_sdk.UniffiLib$Companion.getINSTANCE$iota_sdk(iota_sdk_ffi.kt:5497)
	at iota_sdk.Ed25519PrivateKey$Companion.generate(iota_sdk_ffi.kt:21389)
	at GenerateEd25519AddressKt.main(GenerateEd25519Address.kt:9)
	at GenerateEd25519AddressKt.main(GenerateEd25519Address.kt)

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':example'.
> Process 'command '/opt/hostedtoolcache/Java_Oracle_jdk/21/x64/bin/java'' finished with non-zero exit value 1
```

Upon investigation it turned out, that the `MethodTooLargException` exception is misleading because the issue is not with a single method. The actual issue is that the bindings interface size hits a Kotlin/Java JNA (Java Native Access) specific limitation. This limitation has been encountered before here: https://github.com/mozilla/uniffi-rs/issues/2340 and a bandage on `uniffi-rs` side has been merged with https://github.com/mozilla/uniffi-rs/pull/2344. This however, could not fix the underlying limitation.

The suggested way to circumvent this in our case is apparently to split up the `iota-sdk-ffi` crate into multiple, e.g. `iota-sdk-types-ffi`, etc

Alex6323:
Forwards some utility traits to the bindings by deriving them from the inner core type and exporting them. `uniffi` supports trait exports as mentioned here https://mozilla.github.io/uniffi-rs/latest/types/uniffi_traits.html

Closes #322 , but there's no point in not exporting other traits as well if we can to match as closely as possible the convenient Rust API for bindings users, so I'll do that as well.

Not sure why yet, but `Ord` (supported according to the docs) doesn't work.

Also fixes #433, which is necessary to make the CI pass.


DaughterOfMars (approved):

bindings/kotlin/split_uniffi_interface.py
```diff
@@ -0,0 +1,159 @@
...
+    parser = argparse.ArgumentParser(description="Split large UniffiLib interface into batches.")
+    parser.add_argument(
+        "--file",
+        default="bindings/kotlin/lib/iota_sdk/iota_sdk_ffi.kt",
+        help="Path to generated Kotlin bindings file.",
+    )
+    parser.add_argument(
+        "--batch-size",
+        type=int,
+        default=500,
```
Thoralf-M, line 55: How did you end up with 500?

bindings/kotlin/split_uniffi_interface.py
```diff
@@ -0,0 +1,159 @@
...
+    for start_idx, end_idx, sig_lines in moved:
+        signature = sig_lines[0]
+        m = re.match(r"^\s*fun\s+([A-Za-z0-9_`]+)\s*\(", signature)
+        if not m:
+            raise SystemExit(f"Could not parse method name from line {start_idx+1}: {signature}")
+        name = m.group(1)
+        # skip methods that can't be moved
+        print(f"name={name}")
+        if name == "uniffi_iota_sdk_ffi_fn_init_callback_vtable_transactionsignerfn":
+            continue
```
Thoralf-M: Only this single function can't be moved? Why?

Makefile
```diff
@@ -128,6 +128,7 @@ kotlin: ## Build Kotlin bindings
 	@$(build_binding) \
 	cargo run --bin uniffi-bindgen -- generate --library "target/release/libiota_sdk_ffi$${LIB_EXT}" --language kotlin --out-dir bindings/kotlin/lib --no-format -c bindings/kotlin/uniffi.toml || exit $$?; \
 	cp target/release/libiota_sdk_ffi$${LIB_EXT} bindings/kotlin/lib/
+	python bindings/kotlin/split_uniffi_interface.py --batch-size 500 || exit $$?
```
Thoralf-M, line 131: No need for --batch-size 500 if that's the default

bindings/kotlin/split_uniffi_interface.py
Alex6323: yeah, only that one. Because it's the only called inside of a `register` function on the function argument. Just search through the Kotlin binding file and you'll see what I mean.

Thoralf-M: Ah so probably if we add a new fn that takes a callback as argument we also have to ignore it?

bindings/kotlin/split_uniffi_interface.py
Alex6323: there are ~1600 methods in total in the interface, 500 just seems to be a relative safe number to move out of the original interface, but this number probably has to be adjusted now and then.

Makefile
Alex6323: I wanted to make it explicit because we have to adjust this number eventually.

bindings/kotlin/split_uniffi_interface.py
Alex6323: yes. I opted for ignoring those functions instead of making the script more complex, because it was only 1 function.

Thoralf-M: Sounds good to me, but maybe mention in the comment that it can't be moved because it's called in the registry function

Thoralf-M (approved):

thibault-martinez:
Having a look, don't merge just yet pls

bindings/kotlin/split_uniffi_interface.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env python3
+"""
+Post-process the generated Kotlin bindings to split a chunk of methods
+out of `UniffiLib` into `UniffiLibBatch2`, reducing interface size to
+avoid JNA proxy MethodTooLargeException. Run this after `make kotlin`.
```
thibault-martinez:
```suggestion
avoid JNA proxy MethodTooLargeException.
```
Not really since it's part of it already

bindings/kotlin/split_uniffi_interface.py
```diff
@@ -0,0 +1,158 @@
...
+import re
+from pathlib import Path
+
+
+def find_block(lines: list[str], start_line: int) -> int:
+    """Return the line index of the matching closing brace for a block starting at start_line."""
+    depth = 0
+    for i in range(start_line, len(lines)):
+        depth += lines[i].count("{")
+        depth -= lines[i].count("}")
```
thibault-martinez, line 20: What if there are some curly braces within quotes or comments?

bindings/kotlin/split_uniffi_interface.py
```diff
@@ -0,0 +1,158 @@
...
+        if depth == 0:
+            return i
+    raise ValueError("Did not find matching brace for block starting at line %d" % start_line)
+
+
+def parse_methods(lines: list[str], start: int, end: int) -> list[tuple[int, int, list[str]]]:
+    """
+    Collect method declarations (assumed to span exactly two lines) directly inside the interface block.
```
thibault-martinez, line 28: Is this really a safe assumption?

thibault-martinez: Just worried about the formatter maybe creating a 3rd like at some point maybe

thibault-martinez (approved):

bindings/kotlin/split_uniffi_interface.py
Alex6323: it doesn't do that on method signatures, at least not for the 1600 functions we already have.

bindings/kotlin/split_uniffi_interface.py
Alex6323: I was worried about that as well, which is why I checked carefully. If it fails at some point, we can change the script but I think the sheer number of functions we have makes me confident, that this does in fact not happen. I had a variant prior that was robust against that, but I removed it because of that. Also the code was way more complex then.

bindings/kotlin/split_uniffi_interface.py
DaughterOfMars: but this is called on the whole interface, which contains types, which have comments which may contain an opening or closing brace with no pair

bindings/kotlin/split_uniffi_interface.py
thibault-martinez: Fair enough, I've seen very long function definition that don't get wrapped. I think we would have seen wrapped now if that was a possibility.

bindings/kotlin/split_uniffi_interface.py
Alex6323: Can we clarify what part of the file you mean by `whole interface`? Maybe you assume it's doing that on the whole file?

DaughterOfMars: is it not? what is it working on then?

Alex6323: it's only used from the line `internal interface UniffiLib : Library {` until its matching closing brace, and there you just have the method signatures, no comments.

DaughterOfMars: Okay, fair enough then

PR #424, opened 2025-11-20
feat(iota-sdk-crypto): add generate_mnemonic()

### LINKED ISSUE #354: Allow creating a new random mnemonic
https://github.com/iotaledger/iota-rust-sdk/pull/347 adds the possibility to derive a private key from a mnemonic, but the SDK currently doesn't provide a way to generate a mnemonic, we should add that (english wordlist)

Thoralf-M:
Add generate_mnemonic() so one can generate a mnemonic and can use that later.

Fixes https://github.com/iotaledger/iota-rust-sdk/issues/354

Tested by running the examples:
```
24 word mnemonic: speed trick unique category check chef save rose blanket keep crush version divorce head paper drip lake what audit problem paddle leg away flat
12 word mnemonic: above audit want boost mammal balance dry rifle pride flush know filter
```

### DIFF
--- a/bindings/go/examples/generate_mnemonic/main.go
+++ b/bindings/go/examples/generate_mnemonic/main.go
@@ -12,7 +12,7 @@ import (
 func main() {
 	mnemonic := iota_sdk.GenerateMnemonic(nil)
 	fmt.Println("24 word mnemonic:", mnemonic)
-	wordCount := iota_sdk.MnemonicWordCountWords12
+	wordCount := iota_sdk.MnemonicLengthWords12
 	mnemonic = iota_sdk.GenerateMnemonic(&wordCount)
 	fmt.Println("12 word mnemonic:", mnemonic)
 }
--- a/bindings/go/examples/generate_mnemonic/main.go
+++ b/bindings/go/examples/generate_mnemonic/main.go
@@ -12,7 +12,7 @@ import (
 func main() {
 	mnemonic := iota_sdk.GenerateMnemonic(nil)
 	fmt.Println("24 word mnemonic:", mnemonic)
-	wordCount := iota_sdk.MnemonicWordCountTwelve
+	wordCount := iota_sdk.MnemonicWordCountWords12
 	mnemonic = iota_sdk.GenerateMnemonic(&wordCount)
 	fmt.Println("12 word mnemonic:", mnemonic)
 }
--- a/bindings/go/examples/generate_mnemonic/main.go
+++ b/bindings/go/examples/generate_mnemonic/main.go
@@ -1,0 +1,18 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"fmt"
+
+	"github.com/iotaledger/iota-rust-sdk/bindings/go/iota_sdk"
+)
+
+func main() {
+	mnemonic, err := iota_sdk.GenerateMnemonic(nil)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println("Mnemonic:", mnemonic)
+}
--- a/bindings/go/examples/generate_mnemonic/main.go
+++ b/bindings/go/examples/generate_mnemonic/main.go
@@ -11,5 +11,8 @@ import (
 
 func main() {
 	mnemonic := iota_sdk.GenerateMnemonic(nil)
-	fmt.Println("Mnemonic:", mnemonic)
+	fmt.Println("24 word mnemonic:", mnemonic)
+	wordCount := iota_sdk.MnemonicWordCountTwelve
+	mnemonic = iota_sdk.GenerateMnemonic(&wordCount)
+	fmt.Println("12 word mnemonic:", mnemonic)
 }
--- a/bindings/go/examples/generate_mnemonic/main.go
+++ b/bindings/go/examples/generate_mnemonic/main.go
@@ -10,9 +10,6 @@ import (
 )
 
 func main() {
-	mnemonic, err := iota_sdk.GenerateMnemonic(nil)
-	if err != nil {
-		panic(err)
-	}
+	mnemonic := iota_sdk.GenerateMnemonic(nil)
 	fmt.Println("Mnemonic:", mnemonic)
 }
--- a/crates/iota-sdk-crypto/src/mnemonic.rs
+++ b/crates/iota-sdk-crypto/src/mnemonic.rs
@@ -12,8 +12,8 @@ pub enum MnemonicWordCount {
 
 /// Generate a new BIP-39 mnemonic in English.
 /// Supported word counts are 12 and 24 (default).
-pub fn generate_mnemonic(word_count: Option<MnemonicWordCount>) -> String {
-    let count = word_count.unwrap_or(MnemonicWordCount::TwentyFour) as usize;
+pub fn generate_mnemonic(word_count: impl Into<Option<MnemonicWordCount>>) -> String {
+    let count = word_count.into().unwrap_or(MnemonicWordCount::TwentyFour) as usize;
     Mnemonic::generate(count)
         .expect("mnemonic generation failed") // Safe to unwrap since the word count is controlled
         .to_string()
--- a/crates/iota-sdk-crypto/src/mnemonic.rs
+++ b/crates/iota-sdk-crypto/src/mnemonic.rs
@@ -1,0 +1,14 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use bip39::Mnemonic;
+
+use crate::PrivateKeyError;
+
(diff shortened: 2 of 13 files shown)


crates/iota-sdk-crypto/src/mnemonic.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use bip39::Mnemonic;
+
+use crate::PrivateKeyError;
+
+const DEFAULT_WORD_COUNT: usize = 24;
+
+/// Generate a new BIP-39 mnemonic in English.
+/// Supported word counts are 12, 15, 18, 21, and 24 (default).
+pub fn generate_mnemonic(word_count: Option<usize>) -> Result<String, PrivateKeyError> {
```
thibault-martinez: Feeling like we should have an enum for the word count so we can make this function infallible

thibault-martinez: Or are there other reasons it can fail than a wrong word count?

Thoralf-M: In the bip39 crate it's also a Result because of the entropy, but the way it's called from this function it should be safe to unwrap I guess

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/424/commits/bc36ce6c71526993e2c7bf057cf2b8650c42811b

Thoralf-M: I only added the 12 and 24 word variants as they are most common and our wallet also only supports these two options

crates/iota-sdk-crypto/src/mnemonic.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use bip39::Mnemonic;
+
+#[derive(Debug, Clone, Copy)]
+pub enum MnemonicWordCount {
```
thibault-martinez: please make it non exhaustive if we ever want to add the others

bindings/go/examples/generate_mnemonic/main.go
```diff
@@ -0,0 +1,15 @@
...
+package main
+
+import (
+	"fmt"
+
+	"github.com/iotaledger/iota-rust-sdk/bindings/go/iota_sdk"
+)
+
+func main() {
+	mnemonic := iota_sdk.GenerateMnemonic(nil)
```
thibault-martinez, line 13: Can you maybe change to use a variant? Or maybe do with AND without?

Thoralf-M: Sure https://github.com/iotaledger/iota-rust-sdk/pull/424/commits/c1fee3cac8b315c5d3af0c416fb3a47317e9aab6

crates/iota-sdk-crypto/src/mnemonic.rs
Thoralf-M: Added https://github.com/iotaledger/iota-rust-sdk/pull/424/commits/c1fee3cac8b315c5d3af0c416fb3a47317e9aab6

crates/iota-sdk/examples/generate_mnemonic.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use iota_crypto::mnemonic::{MnemonicWordCount, generate_mnemonic};
+
+fn main() {
+    let mnemonic = generate_mnemonic(None);
+    println!("24 word mnemonic: {mnemonic}");
+    let mnemonic = generate_mnemonic(Some(MnemonicWordCount::Twelve));
```
thibault-martinez: Can make the API a tiny bit nicer without the Some if you take a `impl Into<Option<MnemonicWordCount>>` (not in the FFI crate though)

Thoralf-M: Sure https://github.com/iotaledger/iota-rust-sdk/pull/424/commits/547e812dce9a41269d7c1cd9ad1828f458d4bf5b

thibault-martinez (approved):

crates/iota-sdk-crypto/src/mnemonic.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use bip39::Mnemonic;
+
+#[derive(Debug, Clone, Copy)]
+#[non_exhaustive]
+pub enum MnemonicWordCount {
+    Twelve = 12,
+    TwentyFour = 24,
```
Alex6323: proposing to rename the variants to something including the numbers, maybe `Wc12, Wc24` or `N12, N24`, but no requirement for approval. 

Reasoning:
(1) more readable
(2) shorter
(3) better aligned

Alex6323 (approved):
lgtm

crates/iota-sdk-crypto/src/mnemonic.rs
thibault-martinez: I'm going to go ahead and merge, not sure I agree on this one 😬

thibault-martinez: Out of curiosity, what did `N`, mean?

Thoralf-M: I think `WC` is not really clear and actually more confusing instead of more readable. It would be "repeated" with the enum name `MnemonicWordCount::WC12`, just `12` would have been nice but is not allowed since it's just a number
The tiny-bip39 crate has an enum like this
```Rust
pub enum MnemonicType {
    Words12
```
do you prefer that over `MnemonicWordCount::Twelve`?

thibault-martinez: I wouldn't say it's a "type" though

Thoralf-M: Agree, that was also the reason why I didn't copy it initially
But one could use also just use the variants like `Words12` without their enum name if that's preferred

Alex6323: Words12, Words24 works for me

thibault-martinez: > But one could use also just use the variants like Words12 without their enum name if that's preferred

Not sure I understand

Alex6323: @thibault-martinez `N` is just what I learned at university to use to represent the total number of things. { apple, orange, banana }, N = 3

Thoralf-M: Now `Word` is also duplicated, but probably still better because there is `12` and `24` now, one would usually not search for these written out
Changed in https://github.com/iotaledger/iota-rust-sdk/pull/424/commits/b8fcb5de1058c6ba500ab64f1f185233aea8ce31

Alex6323: I only had to pick something to prefix the number.

thibault-martinez: @Thoralf-M I don't think I like `MnemonicWordCountWords12`...

thibault-martinez: @copilot what would be a good name for the enum and its variants?

Alex6323: I have (partly joking here!) another reason for using numbers ... What if there's a mnemonic with 1337 words? I know it's not happening ... but writing out numbers doesn't scale IMHO.

Thoralf-M:
```enum
enum MnemonicLength {
    Words12
    Words24 
}
```
What about this?

Alex6323: One could argue that this enum represents the `size` or `length` of a Mnemonic, so it'ld become `MnemonicSizeWords12` or `MnemonicLengthWords12`

Alex6323: @Thoralf-M didn't see your message in time. I fully approve on that!!r

Thoralf-M: One more option :trollface: 
```Rust
pub enum MnemonicStrength {
    Bits128,  // 12 words
    Bits256,  // 24 words
}
```

Alex6323: now that we're considering `Strength` I also want to mention `MnemonicPower` :trollface: :trollface:

Alex6323:
> @thibault-martinez I've opened a new pull request, #435, to work on those changes. Once the pull request is ready, I'll request review from you.

copilot my friend, I think you just cheated and read our comments.

PR #426, opened 2025-11-20
feat(workflows): add Kotlin part

Thoralf-M:


.github/workflows/pre_publish.yml
```diff
@@ -72,13 +74,31 @@ jobs:
...
-      # TODO @Thoralf-M
-      # - name: Create PR for Kotlin
-      #   if: inputs.package == 'kotlin-sdk'
-      #   run: |
-      #     echo "Creating PR for Kotlin SDK - Not yet implemented."
-      # TODO @Thoralf-M  get version, bump it, update it
-      # TODO @Thoralf-M generate changelog
+      - name: Bump Kotlin version
+        if: inputs.package == 'iota-kotlin-sdk'
+        id: bump-kotlin
```
Alex6323, line 79: It think we need to give it the same `id` for Kotlin and for Python, so the create pullrequest step below works for both?

.github/workflows/pre_publish.yml
```diff
@@ -72,13 +74,31 @@ jobs:
...
+        run: |
+          # Read current version from build.gradle.kts
+          CURRENT_VERSION=$(grep '^version = ' bindings/kotlin/build.gradle.kts | sed 's/version = "\(.*\)"/\1/')
+          echo "Current version: $CURRENT_VERSION"
+
+          # Bump the version using the input level
+          source ./.github/bump_version.sh
+          export NEW_VERSION=$(bump_semver "$CURRENT_VERSION" "${{ inputs.level || 'alpha' }}")
+          # export NEW_VERSION=$(bump_semver "$CURRENT_VERSION" "${{ inputs.level }}")
```
Alex6323: `shellcheck` again here :trollface: 

apparently one is supposed to create a variable, and then export it.
```
NEW_VERSION=$(bump_semver "$CURRENT_VERSION" "${{ inputs.level || 'alpha' }}")
export NEW_VERSION
```

.github/workflows/pre_publish.yml
Thoralf-M: Can't have the same `id` multiple times, if we want the PR part work for both we should just export the values to env variables

.github/workflows/pre_publish.yml
Thoralf-M: Also with the version below? I forgot to remove the 'alpha' part as this was just for testing
https://github.com/iotaledger/iota-rust-sdk/pull/426/commits/352b69e352f01bc4e2ab4bc810f9fcc98e3bc76f
And I combined them because it looked a bit wasted to do two lines for this :joy:

Alex6323: "ShellCheck (SC2155) warns because combining declaration and command substitution with export masks the exit status of bump_semver; export will return success even if bump_semver fails, so you’d never notice a failure. Split it so the command’s status is visible, e.g.:

NEW_VERSION=$(bump_semver "$CURRENT_VERSION" "${{ inputs.level || 'alpha' }}") || exit 1
export NEW_VERSION
"
(chatty 5.1 Codex Max)

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/426/commits/7ce6724d9af7399230f7e97c2a2d652e09e0ff92

Alex6323 (approved):

.github/workflows/pre_publish.yml
```diff
@@ -101,30 +121,50 @@ jobs:
           pr-template-file: ./.github/release-pr-template.ejs
 
       - name: Create Pull Request
-        if: inputs.package == 'kotlin-sdk' || inputs.package == 'python-sdk'
-        uses: peter-evans/create-pull-request@v7 # TODO pin
+        if: inputs.package == 'iota-kotlin-sdk' || inputs.package == 'iota-python-sdk'
```
Alex6323, line 121: the python part of the if guard should be removed I guess, but I can change that.

Thoralf-M: Won't you use the same part for the pull request?

Alex6323: this is using `bump-kotlin` step which doesn't run in my case?

Alex6323: we could instead both write the new version to a common github env and use that here.

Thoralf-M: Actually thought I did that already, but was just the VERSION_TAG then I guess

Thoralf-M: Changed in https://github.com/iotaledger/iota-rust-sdk/pull/426/commits/1a1254915072138ea2a464e8acae623b327a2b1d

.github/workflows/pre_publish.yml
```diff
@@ -72,13 +74,30 @@ jobs:
...
+        id: bump-kotlin
+        run: |
+          # Read current version from build.gradle.kts
+          CURRENT_VERSION=$(grep '^version = ' bindings/kotlin/build.gradle.kts | sed 's/version = "\(.*\)"/\1/')
+          echo "Current version: $CURRENT_VERSION"
+
+          # Bump the version using the input level
+          source ./.github/bump_version.sh
+          NEW_VERSION=$(bump_semver "$CURRENT_VERSION" "${{ inputs.level }}")
+          export NEW_VERSION
```
thibault-martinez: what's that for?

thibault-martinez: Ah, for gen-changelog.sh, confusing IMO, would be better if it took arguments instead of expecting env vars to be set

thibault-martinez: I can change that later in the base PR if you prefer

.github/workflows/pre_publish.yml
```diff
@@ -101,30 +120,50 @@ jobs:
           pr-template-file: ./.github/release-pr-template.ejs
 
       - name: Create Pull Request
-        if: inputs.package == 'kotlin-sdk' || inputs.package == 'python-sdk'
-        uses: peter-evans/create-pull-request@v7 # TODO pin
+        if: inputs.package == 'iota-kotlin-sdk' || inputs.package == 'iota-python-sdk'
+        id: create-pr
```
thibault-martinez: Can you change the IDs to `pr-create-rust` and `pr-create-bindings` or something? `pr-create` and `create-pr` is meh

.github/workflows/pre_publish.yml
Thoralf-M: `bump_semver` takes arguments though, it's not relying on env vars

.github/workflows/pre_publish.yml
Thoralf-M: Sure https://github.com/iotaledger/iota-rust-sdk/pull/426/commits/e378fdf9eee701d6f79f1f9aab77b67a5ec56b28

.github/workflows/pre_publish.yml
thibault-martinez: Yeah, what's your point? I think it's easier to read with arguments

Thoralf-M: > would be better if it took arguments instead of expecting env vars to be set

my point is that the function is taking arguments and no env vars need to be set
But I guess you want to provide the arguments to `gen-changelog.sh` directly instead of the function from it

thibault-martinez: I think you are confusing `bump_version.sh` and `gen-changelog.sh`

Thoralf-M: Ah yes because that was shown in the code section for your comment :sweat_smile:

Thoralf-M: Changed it https://github.com/iotaledger/iota-rust-sdk/pull/426/commits/96861d7aa44415327cc7b3dfe7776d3eb53a028f

.github/workflows/pre_publish.yml
```diff
@@ -72,13 +74,29 @@ jobs:
...
+        id: bump-kotlin
+        run: |
+          # Read current version from build.gradle.kts
+          CURRENT_VERSION=$(grep '^version = ' bindings/kotlin/build.gradle.kts | sed 's/version = "\(.*\)"/\1/')
+          echo "Current version: $CURRENT_VERSION"
+
+          # Bump the version using the input level
+          source ./.github/bump_version.sh
+          NEW_VERSION=$(bump_semver "$CURRENT_VERSION" "${{ inputs.level }}")
+          export NEW_VERSION
```
thibault-martinez: Not needed anymore

release.toml
```diff
@@ -1,6 +1,4 @@
 enable-all-features = true
-pre-release-hook = [
-  "../../.github/gen-changelog.sh",
-]
+pre-release-hook = ["../../.github/gen-changelog.sh . \"$NEW_VERSION\""]
```
thibault-martinez: So the rust job was exporting a `CRATE_ROOT` to `.`?

.github/workflows/pre_publish.yml
Thoralf-M: Good catch, removed https://github.com/iotaledger/iota-rust-sdk/pull/426/commits/2f5660033e21efc4c03f2602a51d81316f0c9183

release.toml
Thoralf-M: I changed that, but got messed up, fixed in https://github.com/iotaledger/iota-rust-sdk/pull/426/commits/d2b095959ed1a1f9d8309ff7fa46e349b09d82dd
The env var is set here https://github.com/crate-ci/cargo-release/blob/master/docs/reference.md#hook-environment-variables and I don't see that a path is changed, so probably shouldn't use `.`

thibault-martinez (approved):

PR #445, opened 2025-12-03
feat(transaction-builder): Add support for extensible async signer

### LINKED ISSUE #436: Enable signing operations through APIs that support the enclave principle
APIs that perform signature operations (e.g. `TransactionBuilder::execute`) rely on the assumption that their invoker can supply a whole keypair, but this assumption is not valid in cases where users store their crypto-material within a key management system that adhere to the secure enclave principle (e.g. Hashicorp Vault, AWS KMS).

To enable these users (as well as other IOTA products that adhere to the aforementioned principle) to consume the SDK, either the existing APIs are changed or new APIs are introduced.

One possible approach would be to create a new trait `AsyncIotaSigner` that behaves like `IotaSigner` but in an async manner. Providing an implementation of `AsyncIotaSigner` for all types that implement `IotaSigner`, and in turn implementing `IotaSigner` for all key pair types, the existing API would support this use case without having to break.

The approach of relying on the IOTA Keytool is not a viable option, as it fails to support the targets not supported by the Keytool itself like browsers or mobile devices.

DaughterOfMars:
## Description

Adds a new trait to the transaction builder crate, `Signer`, which defines one simple method `sign` to sign a transaction asynchronously. This is implemented for any type that implements `IotaSigner` and thus supports basic signing as well as passing the transaction to other services which hold the appropriate secrets.

Closes #436


crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -1148,7 +1149,7 @@ impl<C: ClientMethods, L> TransactionBuilder<C, L> {
     /// which case the transaction will be sent to the endpoint for execution.
     pub async fn execute(
         mut self,
-        keypair: &SimpleKeypair,
+        keypair: &impl Signer,
```
thibault-martinez: Can you rename `keypair` please?

thibault-martinez (commented):
What does this PR mean for the FFIs?

DaughterOfMars:
> What does this PR mean for the FFIs?

No change to that API

crates/iota-sdk-transaction-builder/src/builder/signer.rs
```diff
@@ -0,0 +1,31 @@
+//! Defines the [`Signer`] trait, which allows users to implement any
```
Thoralf-M: Missing copyright header

thibault-martinez:
> > What does this PR mean for the FFIs?
> 
> No change to that API

I know but my point was that maybe it should? Bindings won't have traits so the existing methods still take a `SimpleKeyPair`, what will happen when we want another method for a new signer? We should at least rename the existing ones to have `key_pair` in the name?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: done

crates/iota-sdk-transaction-builder/src/builder/signer.rs
DaughterOfMars: fixed

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -1184,19 +1185,15 @@ impl<C: ClientMethods, L> TransactionBuilder<C, L> {
     /// finalization.
     pub async fn execute_with_sponsor(
         mut self,
-        keypair: &SimpleKeypair,
-        sponsor_keypair: &SimpleKeypair,
+        keypair: &impl Signer,
+        sponsor_keypair: &impl Signer,
```
thibault-martinez: Any reason this wasn't changed in the FFIs?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -1184,19 +1185,15 @@ impl<C: ClientMethods, L> TransactionBuilder<C, L> {
     /// finalization.
     pub async fn execute_with_sponsor(
         mut self,
-        keypair: &SimpleKeypair,
-        sponsor_keypair: &SimpleKeypair,
+        keypair: &impl Signer,
```
thibault-martinez: rename to signer

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -876,7 +877,7 @@ impl<L> TransactionBuilder<(), L> {
     /// [`TransactionEffects`]
     pub async fn execute_with_gas_station(
         mut self,
-        keypair: &SimpleKeypair,
+        keypair: &impl Signer,
```
thibault-martinez: rename

crates/iota-sdk-transaction-builder/src/builder/gas_station.rs
```diff
@@ -353,7 +352,7 @@ impl GasStationData {
     pub(crate) async fn execute_txn(
         self,
         txn: &mut Transaction,
-        keypair: &SimpleKeypair,
+        keypair: &impl Signer,
```
thibault-martinez: rename

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: just forgot, thanks

crates/iota-sdk-transaction-builder/src/builder/gas_station.rs
DaughterOfMars: done

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: done

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: done

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: done

crates/iota-sdk-transaction-builder/src/builder/signer.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use iota_crypto::{IotaSigner, SignatureError};
+use iota_types::{Transaction, UserSignature};
+
+/// Defines a type which can sign a transaction asynchronously.
+///
+/// This trait can be implemented downstream to enable signing when using the
+/// [`TransactionBuilder`](crate::TransactionBuilder)
+/// [`execute`](crate::TransactionBuilder::execute) function.
+pub trait Signer {
```
thibault-martinez: I have thought about our discussion a bit more and concluded that I would be fine with this only handling transaction signatures if we rename it to `TransactionSigner`. This ends up in the global scope of the bindings and having something called `Signer` than handles only one of the signature possibilities is just not ok. I can do it if you want.

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/445/commits/f7305d1af41ab1ed428f691442c3e5c1a0f9480a

crates/iota-sdk-ffi/src/transaction_builder/signer.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/// Defines a type which can sign a transaction asynchronously.
+///
+/// This trait can be implemented downstream to enable signing when using the
+/// `TransactionBuilder::execute` function.
+#[uniffi::export(with_foreign)]
+#[async_trait::async_trait]
+pub trait TransactionSignerFn: Send + Sync + std::fmt::Debug {
+    /// Sign a transaction and return a BCS serialized `UserSignature`.
```
thibault-martinez: Why do you say BCS serialised? It just returns a `UserSignature`?

thibault-martinez (approved):

crates/iota-sdk-ffi/src/transaction_builder/signer.rs
DaughterOfMars: Oops, leftover from when it was a byte vec

PR #447, opened 2025-12-03
feat(ci): run rust examples

thibault-martinez:
Rust examples should be ran in the CI the same way bindings examples are.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,21 +37,6 @@ jobs:
 
   tests-with-network:
     runs-on: ubuntu-latest
-    services:
-      postgres: # we need this postgres instance for running a local network with indexer and graphql
-        image: postgres
-        env:
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: postgrespw
-          POSTGRES_DB: iota_indexer_v2
-          POSTGRES_HOST_AUTH_METHOD: trust
-        options: >-
-          --health-cmd pg_isready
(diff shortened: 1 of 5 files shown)


DaughterOfMars (commented):
If we merge this now I will have to remove the example from my PR. Is that what we want?

thibault-martinez:
> If we merge this now I will have to remove the example from my PR. Is that what we want?

We don't really know when we'll be able to merge the AA PR, it depends on the vm-lang team so let's merge this and put back the example later it's fine. You have an issue for examples anyway.

DaughterOfMars (approved):

.github/workflows/tests.yml
```diff
@@ -70,6 +70,32 @@ jobs:
           NETWORK: "local"
         run: make test-with-localnet
 
+  examples:
+    runs-on: ubuntu-latest
+    services:
+      postgres: # we need this postgres instance for running a local network with indexer and graphql
```
Thoralf-M: Can't we move this into the start-local-network action?

thibault-martinez: You can't have a service in a composite action but I'm trying with a step, let's see https://github.com/iotaledger/iota-rust-sdk/pull/447/commits/69485a83d317da42a874b74343a72c0f38176971

thibault-martinez: Passed, nice change 👍🏻

Thoralf-M (approved):

PR #459, opened 2025-12-04
chore(ci): Filter bindings CI file paths for PRs

### LINKED ISSUE #458: CI: filter bindings CI on FFI crate and bindings folders

DaughterOfMars:
## Description

Filter bindings CI file paths for PRs the same way we do for pushes to develop.

Closes #458

### DIFF
--- a/.github/workflows/bindings.yml
+++ b/.github/workflows/bindings.yml
@@ -14,7 +14,7 @@ jobs:
   diff:
     runs-on: ubuntu-latest
     outputs:
-      filters: ${{ steps.diff.outputs }}
+      isBindings: ${{ steps.diff.outputs.isBindings }}
     steps:
       - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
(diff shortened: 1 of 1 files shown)


.github/workflows/bindings.yml
```diff
@@ -9,6 +9,10 @@ on:
       - "crates/iota-sdk-ffi/**"
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+    paths:
+      - ".github/workflow/bindings.yml"
+      - "bindings/**"
+      - "crates/iota-sdk-ffi/**"
```
thibault-martinez: Is there no global path instead of duplicating it?

thibault-martinez (approved):

thibault-martinez:
Ah, they are required, any way to do the same monorepo trick?

.github/workflows/bindings.yml
DaughterOfMars: I don't think so but maybe I missed it

thibault-martinez (approved):

Issue #429, opened 2025-11-24
Step 3: investigate removal of the iota_sdk_types_conversions.rs file


thibault-martinez:

DaughterOfMars:
Four additional types that can be removed:
- [x] `CheckpointCommitment` https://github.com/iotaledger/iota/pull/11976
- [x] `EndOfEpochData` https://github.com/iotaledger/iota/pull/12027
- [x] `CheckpointSummary` https://github.com/iotaledger/iota/pull/12091
- [x] `CheckpointContents` https://github.com/iotaledger/iota/pull/12191

thibault-martinez:
Everything done except https://github.com/iotaledger/iota/issues/12253 and https://github.com/iotaledger/iota/issues/10724

PR #463, opened 2025-12-04
chore(types): Add some missing functionality needed by core

DaughterOfMars:
## Description

While investigating #429 I found some things that should probably be added to the SDK in order to integrate it with the core repo.

### DIFF
--- a/crates/iota-sdk-types/src/digest.rs
+++ b/crates/iota-sdk-types/src/digest.rs
@@ -88,6 +88,18 @@ impl Digest {
             .map_err(|_| DigestParseError)
             .map(Self)
     }
+
+    pub fn next_lexicographical(&self) -> Option<Self> {
+        let mut next_digest = *self;
+        let pos = next_digest.0.iter().rposition(|&byte| byte != 255)?;
+        next_digest.0[pos] += 1;
+        next_digest
+            .0
+            .iter_mut()
+            .skip(pos + 1)
+            .for_each(|byte| *byte = 0);
+        Some(next_digest)
+    }
 }
 
 impl std::str::FromStr for Digest {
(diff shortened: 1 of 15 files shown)


thibault-martinez (commented):
Should probably add the new methods to FFI?

crates/iota-sdk-types/src/digest.rs
```diff
@@ -88,6 +88,18 @@ impl Digest {
             .map_err(|_| DigestParseError)
             .map(Self)
     }
+
+    pub fn next_lexicographical(&self) -> Option<Self> {
```
thibault-martinez: Was there no docs wherever you took this? It's not trivial to understand...

DaughterOfMars: Yeah, none whatsoever

thibault-martinez (approved):

crates/iota-sdk-types/src/digest.rs
```diff
@@ -224,4 +237,31 @@ mod tests {
         let d = s.parse::<Digest>().unwrap();
         assert_eq!(digest, d);
     }
+
+    #[test]
+    fn test_lexical_order() {
```
thibault-martinez, line 242: Are these tests from the monorepo?

DaughterOfMars: no, I made it

Thoralf-M (approved):

crates/iota-sdk-types/src/digest.rs
thibault-martinez: There are some in the monorepo, let's maybe take them as well

DaughterOfMars: these are already covering those cases basically

thibault-martinez: Alrighty

PR #476, opened 2025-12-15
feat: make sure `iota-sdk` is wasm compatible

thibault-martinez:
Contrary to other crates, we are just wasm-building the library and not tests/examples. Making the examples wasm-compatible is not really interesting for the vast majority of people.

### DIFF
--- a/crates/iota-sdk-transaction-builder/Cargo.toml
+++ b/crates/iota-sdk-transaction-builder/Cargo.toml
@@ -23,7 +23,6 @@ iota-graphql-client.workspace = true
 iota-types = { workspace = true, features = ["serde", "hash"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-getrandom = { version = "0.2", features = ["js"] }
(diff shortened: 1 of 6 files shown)


crates/iota-sdk-transaction-builder/Cargo.toml
```diff
@@ -23,7 +23,6 @@ iota-graphql-client.workspace = true
 iota-types = { workspace = true, features = ["serde", "hash"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-getrandom = { version = "0.2", features = ["js"] }
```
Thoralf-M, line 26: So if someone only wants to use the transaction builder and not import it through the sdk crate, would it still work for wasm if we don't have this here?

Alex6323 (approved):

crates/iota-sdk-transaction-builder/Cargo.toml
thibault-martinez: Yes it still works, it was never needed in the first place

Thoralf-M (approved):

PR #482, opened 2025-12-22
chore(transaction-builder): rename `name` to `assign`

thibault-martinez:
Maybe replacing https://github.com/iotaledger/iota-rust-sdk/pull/324, I think `assign` is a better name than `name` as it's what its CLI PTB counterpart uses and what people are used to. Only changed the rust SDK as the bindings don't really use this syntax. It remains a `names` argument in the Python bindings that we can choose to keep like that or change to `assign` as well.


thibault-martinez:
Tbh I'm also not super happy about `res()`, maybe `assigned()`?

DaughterOfMars (approved):

Thoralf-M (approved):

DaughterOfMars (approved):

Alex6323 (approved):

PR #487, opened 2025-12-23
feat: make public enums non_exhaustive

thibault-martinez:
Based on https://github.com/MystenLabs/sui-rust-sdk/pull/142/commits/fd36eb1386be07c5ef9d9e31125be22d92aa2b96

### DIFF
--- a/crates/iota-sdk-ffi/src/types/transaction/v1.rs
+++ b/crates/iota-sdk-ffi/src/types/transaction/v1.rs
@@ -252,6 +252,7 @@ impl From<iota_sdk::types::UnchangedSharedKind> for UnchangedSharedKind {
                 Self::Cancelled { version }
             }
             iota_sdk::types::UnchangedSharedKind::PerEpochConfig => Self::PerEpochConfig,
+            _ => unimplemented!(),
         }
     }
 }
@@ -311,6 +312,7 @@ impl From<iota_sdk::types::ObjectIn> for ObjectIn {
                 digest: Arc::new(digest.into()),
                 owner: Arc::new(owner.into()),
             },
+            _ => unimplemented!(),
         }
     }
 }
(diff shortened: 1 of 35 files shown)


crates/iota-sdk-transaction-builder/src/builder/gas_station.rs
```diff
@@ -407,6 +407,7 @@ impl GasStationData {
                 inner_txn.gas_payment.objects = objects;
                 reservation_id
             }
+            _ => unimplemented!(),
```
Thoralf-M: Shouldn't we error instead of panic here since we return a Result?

crates/iota-sdk-ffi/src/types/transaction/v1.rs
```diff
@@ -252,6 +252,7 @@ impl From<iota_sdk::types::UnchangedSharedKind> for UnchangedSharedKind {
                 Self::Cancelled { version }
             }
             iota_sdk::types::UnchangedSharedKind::PerEpochConfig => Self::PerEpochConfig,
+            _ => unimplemented!(),
```
DaughterOfMars: Add a message to these to indicate that there is a new variant?

crates/iota-sdk-transaction-builder/src/builder/gas_station.rs
thibault-martinez: Idk, it's more of something for us dev-tools to act on than something consumers would be interested in. I can add some more context but I think I'd keep it a panic. @DaughterOfMars opinion?

DaughterOfMars: I also think panic is better here because there's no way to handle the error anyway

Thoralf-M (approved):

crates/iota-sdk-ffi/src/types/transaction/v1.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/487/commits/a55ac705bbc70b1ac7f7fcb9e8b0961d37fab8ee

DaughterOfMars (approved):

PR #503, opened 2026-01-07
fix: impl Error for IntentError

thibault-martinez:


crates/iota-sdk-types/src/crypto/intent.rs
```diff
@@ -7,11 +7,18 @@ use std::str::FromStr;
 
 pub const INTENT_PREFIX_LENGTH: usize = 3;
 
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
 pub enum IntentError {
+    #[error("Invalid bytes for Intent")]
     Bytes,
-    Intent,
+    #[error("Invalid String for Intent")]
```
Thoralf-M: Should it mention that it's supposed to be a hex string?

Thoralf-M (approved):

DaughterOfMars (approved):

Alex6323 (approved):

crates/iota-sdk-types/src/crypto/intent.rs
DaughterOfMars: also maybe no capital letters?

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/503/commits/7a4cc07b0a237323099282d8ebcbc9b946ff2fe8

PR #516, opened 2026-01-09
feat(iota-sdk-graphql-client): add move_view_call

### LINKED ISSUE #507: Support Move View Call

Thoralf-M:
Add move_view_call() to the client so one can get a JSON response for a single move call

IIP for it: https://github.com/iotaledger/IIPs/blob/main/iips/IIP-0005/iip-0005.md

Fixes #507


Thoralf-M:
Hash example fails until https://github.com/iotaledger/iota/issues/9732 is fixed

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1656,6 +1656,49 @@ impl Client {
...
+    ///   JSON format
+    ///
+    /// # Returns
+    /// A `MoveViewResult` containing either execution results (return values)
+    /// or an error.
+    pub async fn move_view_call(
+        &self,
+        function_name: String,
+        type_args: impl Into<Option<Vec<String>>>,
+        arguments: impl Into<Option<Vec<serde_json::Value>>>,
```
DaughterOfMars: Maybe this is preferable?
```suggestion
        function_name: Identifier,
        type_args: impl Into<Option<Vec<TypeTag>>>,
        arguments: impl Into<Option<Vec<serde_json::Value>>>,
```

DaughterOfMars: Also since the function name is fully qualified, could we instead accept it in three params? `package_id`, `module` and `function`?

crates/iota-sdk/examples/move_view_call.rs
```diff
@@ -0,0 +1,46 @@
...
+    let result = client
+        .move_view_call(
+            "0x5e7a300e640f645a4030aeb507c7be16909e6fa9711e7ca2d4397bbd967d5c50::auction::get_auction_metadata".to_string(),
+            None::<Vec<String>>,
+            Some(vec![
+                serde_json::json!("0x31deb8cbd320867089d52c37fed2d443520aac0fc5a957de1f64f9135b83f42b"),
+                serde_json::json!("auc.iota"),
+            ]),
+        )
+        .await?;
```
DaughterOfMars, line 22: It would be nice to re-use some of the API for the transaction builder to simplify this. We could also have a builder pattern for adding the call and type args to avoid thing like `None::<Vec<String>>`

Thoralf-M: I want to try just adding a move view method to the transaction builder

crates/iota-sdk-graphql-client/src/lib.rs
Thoralf-M: Yes but not sure if that's actually nicer to use than all together

crates/iota-sdk/examples/move_view_call.rs
DaughterOfMars: Does it make sense to use the transaction builder? Can you call multiple move view functions?

crates/iota-sdk-graphql-client/src/lib.rs
DaughterOfMars: It would be consistent with `move_call` though

crates/iota-sdk/examples/move_view_call.rs
Thoralf-M: At the moment it's limited to one, but still would be the exact same way to build the function call

DaughterOfMars: Feels like it would be better to have a different type then, and just mimic the behavior of the txn builder without all the complex stuff. `MoveViewCallBuilder` maybe?

crates/iota-sdk-graphql-client/src/lib.rs
Thoralf-M: move_view_call exists to make it easier to get data if you don't know the details, therefore I'm not sure if having the specific types here are better or not

crates/iota-sdk/examples/move_view_call.rs
Thoralf-M: Mimicking the tx builder sounds like it would be a lot of duplicated code, I added it to the tx builder now, but not sure if that's really good or if we should only have the client method

Makefile
```diff
@@ -183,13 +183,13 @@ kotlin-examples: ## Run all Kotlin bindings examples
 .PHONY: kotlin-examples-format-check
 kotlin-examples-format-check: ## Check format of all Kotlin bindings examples
 	cd bindings/kotlin; \
-	find examples -name "*.kt" -not -path "*/release/*" -exec ./gradlew KtfmtCheck --files {} \; || exit $$?; \
+	./gradlew KtfmtCheck || exit $$?; \
```
thibault-martinez, line 186: So in the end we are checking format of the release example and it's fine?

Thoralf-M: Seems so, I don't know why it complained earlier

thibault-martinez: What about python and go?

bindings/python/examples/move_view_call.py
```diff
@@ -0,0 +1,44 @@
...
+
+    result = await client.move_view_call("0x2::hash::blake2b256", None,
+                                         arguments)
+
+    if result.error is not None:
+        print("Error:", result.error)
+    elif result.results is not None:
+        print("Results:", result.results)
+    else:
+        print("No results")
```
thibault-martinez:
```suggestion
    if result.error is not None:
        print("Hash Error:", result.error)
    elif result.results is not None:
        print("Hash Results:", result.results)
    else:
        print("No hash results")
```
To be consistent with the 3 other examples

thibault-martinez (commented):
I think it would be best if one of the examples needed a type args

crates/iota-sdk/examples/move_view_call.rs
```diff
@@ -0,0 +1,46 @@
...
+use iota_sdk::graphql_client::{Client, error::Result};
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let client = Client::new_devnet();
+
+    let result = client
+        .move_view_call(
+            "0x5e7a300e640f645a4030aeb507c7be16909e6fa9711e7ca2d4397bbd967d5c50::auction::get_auction_metadata".to_string(),
+            None::<Vec<String>>,
```
thibault-martinez: The fully qualified None doesn't seem necessary

crates/iota-sdk-graphql-client/src/query_types/move_view_call.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// The result of executing a Move View Function.
+///
+/// Execution errors are captured in the `error` field, in which case the
+/// `results` field will be `None`. On success, the `results` field will contain
+/// the return values of the Move view function, and the `error` field will be
+/// `None`.
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema = "rpc", graphql_type = "MoveViewResult")]
+pub struct MoveViewResult {
```
thibault-martinez, line 35:
```suggestion
pub struct MoveViewResult {
```
Kinda triggers me it doesn't contain Call

Thoralf-M: Also like that in the graphql, want me to udpate it there as well?

crates/iota-sdk/examples/move_view_call.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/41f8568d8e6121228f8621b8d55ecd1afacde10d

Makefile
Thoralf-M: Also seems fine there https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/04cdb0ddcde2c06014c889eef63dc68defff2900

crates/iota-sdk-graphql-client/src/query_types/move_view_call.rs
thibault-martinez: Would it break the schema or something?

Thoralf-M: Would be a breaking change yes, but since it's only deployed on devnet yet we could still do that I think

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -1656,6 +1656,49 @@ impl Client {
...
+    /// * `type_args` - The type arguments of the Move function
+    /// * `arguments` - The arguments to be passed into the Move function, in
+    ///   JSON format
+    ///
+    /// # Returns
+    /// A `MoveViewResult` containing either execution results (return values)
+    /// or an error.
+    pub async fn move_view_call(
+        &self,
+        function_name: String,
```
thibault-martinez: - take an Into<String> or equivalent so it's easier to call (with a &str)
- I'd say `function_name` is slightly misleading as it's more than that

crates/iota-sdk/examples/move_view_call.rs
thibault-martinez: 2 cents: `None::<Vec<String>>` was not necessary in the first place

crates/iota-sdk-graphql-client/src/lib.rs
Thoralf-M: Also wasn't sure about the `function_name`, but didn't find a better alternative and this is also the name used in the grapql spec

Thoralf-M: For the Into: https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/227ae6af9883abbe68547bb5b1636a852104a453

crates/iota-sdk/examples/move_view_call.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use iota_sdk::graphql_client::{Client, error::Result};
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let client = Client::new_devnet();
+
+    let result = client
+        .move_view_call(
+            "0x5e7a300e640f645a4030aeb507c7be16909e6fa9711e7ca2d4397bbd967d5c50::auction::get_auction_metadata".to_string(),
```
thibault-martinez:
```suggestion
            "0x5e7a300e640f645a4030aeb507c7be16909e6fa9711e7ca2d4397bbd967d5c50::auction::get_auction_metadata",
```

crates/iota-sdk/examples/move_view_call.rs
```diff
@@ -0,0 +1,102 @@
...
+        println!("JSON Error: {error}");
+    } else if let Some(results) = result.results {
+        println!("JSON Results: {results:?}");
+    } else {
+        println!("No JSON results");
+    }
+
+    // ===========================================================================
+    // Example 3: Using move_view_call() with typed arguments (auction)
+    // ===========================================================================
```
thibault-martinez, line 55: Not adding examples 3 and 4 to the bindings? Also, would be nice to have at least one example with type arguments

Thoralf-M: In the bindings it's only the json function atm, wanted to get feedback from you first

crates/iota-sdk-graphql-client/src/move_view_call.rs
```diff
@@ -0,0 +1,352 @@
...
+
+impl MoveViewArg for &serde_json::Value {
+    fn to_json(&self) -> serde_json::Value {
+        (*self).clone()
+    }
+}
+
+// ===========================================================================
+// MoveViewArgList Trait and Implementations
+// ===========================================================================
```
thibault-martinez: I would either remove that or add it to `MoveViewArg` as well, otherwise it's inconsistent. I'd remove it.

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -2327,4 +2328,66 @@ mod tests {
...
+
+        assert!(
+            result.error.is_none(),
+            "Move view call returned error for {} network: {:?}",
+            client.rpc_server(),
+            result.error
+        );
+        assert!(
+            result.results.is_some(),
+            "Move view call returned no results for {} network",
```
thibault-martinez: Would the result be too annoying to match against?

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -2327,4 +2328,66 @@ mod tests {
...
             .unwrap();
     }
+
+    #[tokio::test]
+    async fn test_move_view_call() {
+        let client = test_client();
+
+        // Test blake2b256 hash function with typed arguments
+        let result = client
+            .move_view_call("0x2::hash::blake2b256", None, vec![0u8, 1, 2])
```
thibault-martinez: Even if not needed, maybe pass a type argument to test it?

crates/iota-sdk-ffi/src/types/graphql.rs
```diff
@@ -1182,6 +1182,30 @@ pub struct MoveEnum {
     pub variants: Option<Vec<MoveEnumVariant>>,
 }
 
+#[derive(uniffi::Record)]
```
thibault-martinez, line 1191: Copy over the docs

crates/iota-sdk-graphql-client/src/move_view_call.rs
```diff
@@ -0,0 +1,352 @@
...
+                    $(
+                        self.$n.to_json()
+                    ),+
+                ]
+            }
+        }
+    };
+}
+
+impl_move_view_args_tuple!((0, T0), (1, T1));
```
thibault-martinez: You didn't want to use `variadics_please`?

crates/iota-sdk-graphql-client/src/query_types/move_view_call.rs
```diff
@@ -0,0 +1,57 @@
...
+/// The function name should be fully qualified as
+/// `<package_id>::<module_name>::<function_name>`,
+/// e.g., `0x3::iota_system::get_total_iota_supply`.
+#[derive(cynic::QueryVariables, Debug)]
+pub struct MoveViewCallArgs {
+    /// The Move function fully qualified name as
+    /// `<package_id>::<module_name>::<function_name>`.
+    pub function_name: String,
+    /// The type arguments of the Move function.
+    pub type_args: Option<Vec<String>>,
```
thibault-martinez: I know it's the schema but the inconsistency type_args vs arguments is killing me

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -2327,4 +2328,66 @@ mod tests {
...
+    }
+
+    #[tokio::test]
+    async fn test_move_view_call_json() {
+        let client = test_client();
+
+        // Test blake2b256 hash function with JSON arguments
+        let result = client
+            .move_view_call_json(
+                "0x2::hash::blake2b256",
```
DaughterOfMars, line 2410: I'd still prefer three params here, because imagine you instead have already parsed the address, now you have to format it into a string just so it can be parsed again. Also, if you use the full address then this will be a very long string :c

crates/iota-sdk-graphql-client/src/move_view_call.rs
```diff
@@ -0,0 +1,352 @@
...
+}
+
+impl MoveViewArg for &Address {
+    fn to_json(&self) -> serde_json::Value {
+        serde_json::Value::String(self.to_string())
+    }
+}
+
+/// Wrapper for specifying shared (immutable) object parameters.
+pub struct Shared<T>(pub T);
```
DaughterOfMars: I don't think we need this, right? The whole point of it in the txn builder is because we have to specify the `Input` differently depending on whether it's shared/shared mut/etc. But here it's just an object ID, no?

crates/iota-sdk-graphql-client/src/lib.rs
thibault-martinez: But you could say the same for the opposite? If you have the string then you need to split it in 3

crates/iota-sdk-graphql-client/src/lib.rs
Thoralf-M: Would be an error here, need a function with generics

crates/iota-sdk-graphql-client/src/query_types/move_view_call.rs
Thoralf-M: What do you prefer? Changing to type_arguments or args?

crates/iota-sdk-graphql-client/src/lib.rs
thibault-martinez: Ah yeah, but would still be nice to test it though, with a function with generics

crates/iota-sdk-graphql-client/src/query_types/move_view_call.rs
thibault-martinez: I prefer `type_arguments` but let's not do it if it's breaking the schema, I was just venting

bindings/go/examples/move_view_call/main.go
```diff
@@ -0,0 +1,150 @@
...
+	fmt.Println("=== Example 1: MoveViewCall() with typed arguments (blake2b256) ===")
+	fmt.Println()
+
+	// Using typed arguments: an array of u8 values
+	hashArgs := []iota_sdk.MoveViewArg{
+		iota_sdk.MoveViewArgArray{
+			Value: []iota_sdk.MoveViewArg{
+				iota_sdk.MoveViewArgU8{Value: 0},
+				iota_sdk.MoveViewArgU8{Value: 1},
+				iota_sdk.MoveViewArgU8{Value: 2},
```
thibault-martinez: 😭

crates/iota-sdk-ffi/src/graphql/move_view_call.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// An unsigned 32-bit integer.
+    U32 { value: u32 },
+    /// An unsigned 64-bit integer.
+    U64 { value: u64 },
+    /// An unsigned 128-bit integer (as string to avoid precision loss).
+    U128 { value: String },
+    /// A string value.
+    Str { value: String },
+    /// An object ID.
+    Object { value: Arc<ObjectId> },
```
thibault-martinez: I'd call it ObjectId

crates/iota-sdk-ffi/src/graphql/move_view_call.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// An address.
+    Addr { value: Arc<Address> },
```
thibault-martinez: Address

crates/iota-sdk-ffi/src/graphql/move_view_call.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// A vector/array of arguments.
+    Array { value: Vec<MoveViewArg> },
+    /// A null/none value (for Option::None).
+    Null,
```
thibault-martinez: Why not Option like MoveArg?

crates/iota-sdk-ffi/src/graphql/move_view_call.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// An unsigned 128-bit integer (as string to avoid precision loss).
+    U128 { value: String },
+    /// A string value.
+    Str { value: String },
+    /// An object ID.
+    Object { value: Arc<ObjectId> },
+    /// An address.
+    Addr { value: Arc<Address> },
+    /// A vector/array of arguments.
+    Array { value: Vec<MoveViewArg> },
```
thibault-martinez: I'd keep consistent with MoveArg and call it Vector

bindings/go/examples/move_view_call/main.go
Thoralf-M: Well yeah would just use the JSON version instead

crates/iota-sdk-graphql-client/src/move_view_call.rs
```diff
@@ -0,0 +1,287 @@
...
+    /// * `arguments` - The arguments to be passed into the Move function, in
+    ///   JSON format
+    ///
+    /// # Returns
+    /// A `MoveViewResult` containing either execution results (return values)
+    /// or an error.
+    pub async fn move_view_call_json(
+        &self,
+        function_name: impl Into<String>,
+        type_args: impl Into<Option<Vec<String>>>,
```
thibault-martinez: Maybe we can at least change the name here :)

thibault-martinez: (and in the docs if ever you change it)

thibault-martinez: And same for the other function ofc

crates/iota-sdk-graphql-client/src/lib.rs
DaughterOfMars: This is why it would be nice to have a builder, to cover cases like this. I don't really like forcing users to build to canonical string themselves

crates/iota-sdk-graphql-client/src/query_types/move_view_call.rs
Thoralf-M: Well it's not breaking the schema if we rename it here, but it would not match the schema

crates/iota-sdk-ffi/src/graphql/move_view_call.rs
Thoralf-M: Was an Enum instead of an Object, but I changed it now

crates/iota-sdk-graphql-client/src/query_types/move_view_call.rs
thibault-martinez: Ah, then I guess it's fine to rename

bindings/go/examples/move_view_call/main.go
thibault-martinez: Way better now ❤️

crates/iota-sdk-graphql-client/src/move_view_call.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/5aabc63bc0979b4e8fa1136b3a24522b8a35d486

crates/iota-sdk-graphql-client/src/move_view_call.rs
Thoralf-M: Yes, removed

crates/iota-sdk-graphql-client/src/lib.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/eda9531139f865d239cdeef8e5aeab67323ec77d

crates/iota-sdk-graphql-client/src/move_view_call.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// Allow passing raw JSON values
+impl MoveViewArg for serde_json::Value {
+    fn to_json(&self) -> serde_json::Value {
+        self.clone()
+    }
+}
+
+impl MoveViewArg for &serde_json::Value {
```
DaughterOfMars: You can probably replace several of these with
```
impl<T: MoveViewArg> MoveViewArg for &T {...}
```
I couldn't do that with the PTB args because there is another generic impl blocking it. I'm not 100% sure this will work but possibly worth trying

crates/iota-sdk-graphql-client/src/move_view_call.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn to_json(&self) -> serde_json::Value {
+        (*self).clone()
+    }
+}
+
+/// A trait which defines a list of arguments for a Move View Function call.
+#[diagnostic::on_unimplemented(
+    message = "Provided value is not a valid list of Move view arguments.",
+    note = "Expected a tuple, vector, array, or slice of types that implement `MoveViewArg`."
```
DaughterOfMars: I don't think this is actually implemented for any of these except tuples

crates/iota-sdk-graphql-client/src/query_types/move_view_call.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/427ba9912752d6af299144c7727d50628b5f9081

crates/iota-sdk-graphql-client/src/move_view_call.rs
Thoralf-M: True, just copied it over, fixed in https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/fc52eb899cfbfcb3352dcdde812f50151a999be9

crates/iota-sdk-graphql-client/src/move_view_call.rs
Thoralf-M: Works :partying_face: https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/fc52eb899cfbfcb3352dcdde812f50151a999be9

crates/iota-sdk-graphql-client/src/move_view_call.rs
DaughterOfMars: Oh, I assumed you would implement it 😆

crates/iota-sdk-graphql-client/src/move_view_call.rs
DaughterOfMars: Nice, you can also do this with
```
impl<T: MoveViewArg> MoveViewArg for [T]
```

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -2327,4 +2328,111 @@ mod tests {
...
+            .as_array()
+            .expect("First result should be a JSON array")
+            .iter()
+            .map(|v| v.as_u64().expect("Each element should be a number") as u8)
+            .collect();
+        assert_eq!(actual_hash, expected_hash);
+
+        // Test option::some with type argument
+        let result = client
+            .move_view_call_json(
```
thibault-martinez: Move this to `test_move_view_call_json`?

crates/iota-sdk-graphql-client/src/lib.rs
```diff
@@ -2327,4 +2328,111 @@ mod tests {
...
+            "Move view call JSON returned error for {} network: {:?}",
+            client.rpc_server(),
+            result.error
+        );
+        assert!(
+            result.results.is_some(),
+            "Move view call JSON returned no results for {} network",
+            client.rpc_server()
+        );
+    }
```
thibault-martinez, line 2480: Maybe also do the hash comparison since you already have it

crates/iota-sdk-graphql-client/src/lib.rs
Thoralf-M: Added it to move_view_call(), but kept it was String for move_view_call_json() https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/626bf0f79d89c948d922d3017c02ebbfd8de3cd8

crates/iota-sdk-graphql-client/src/lib.rs
Thoralf-M: Thought the same, changed in https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/626bf0f79d89c948d922d3017c02ebbfd8de3cd8

crates/iota-sdk-graphql-client/src/lib.rs
Thoralf-M: Done in https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/626bf0f79d89c948d922d3017c02ebbfd8de3cd8

thibault-martinez (approved):

crates/iota-sdk-graphql-client/src/move_view_call.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/bb419e6dfaaf615e7d8c661f53fb9bd3a6afc64f

crates/iota-sdk-graphql-client/src/move_view_call.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/cf8141bf92df462aae9c90ded11c01475a360a18

crates/iota-sdk-graphql-client/src/move_view_call.rs
```diff
@@ -237,17 +237,38 @@ impl<T: MoveViewArg> MoveViewArg for &T {
...
 )]
 pub trait MoveViewArgList {
     /// Convert the arguments to a vector of JSON values.
     fn to_json_vec(self) -> Vec<serde_json::Value>;
 }
 
-// Single element implementation
-impl<T: MoveViewArg> MoveViewArgList for T {
+// Single element tuple implementation
+impl<T: MoveViewArg> MoveViewArgList for (T,) {
```
DaughterOfMars: This can be done with the `variadics_please` call. Also, I suspect that you did this because 
```impl<T: MoveViewArg> MoveViewArgList for T``` started failing? It may be that was the issue preventing me from implementing ```impl<T: MoveViewArg> MoveViewArg for &T``` If so, I would recommend reverting the change and allowing the list impl for all `MoveViewArg` types for easier UX

Thoralf-M: > Also, I suspect that you did this because
> impl<T: MoveViewArg> MoveViewArgList for T started failing?

Exactly

> and allowing the list impl for all MoveViewArg types for easier UX

You mean like it is now, or adding a new impl for something else?

DaughterOfMars: I would add the impls for the various list types and revert the generic impl ```impl<T: MoveViewArg> MoveViewArg for &T``` (so you have to manually impl for all those ref types again, sorry 😭 )

Thoralf-M: Please let me know if that's what you meant https://github.com/iotaledger/iota-rust-sdk/pull/516/commits/73dc158d5a9f838dfb41b37650827d567529565b

DaughterOfMars (approved):

PR #521, opened 2026-01-19
feat: Use Version struct over type def

DaughterOfMars:
## Description

Adds a `Version` struct like the core one which has various methods and constants for interacting with versions (A.K.A. sequence numbers).


thibault-martinez:
CI is failing

crates/iota-sdk-types/src/version.rs
```diff
@@ -0,0 +1,233 @@
...
+    derive_more::RemAssign,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[repr(transparent)]
+pub struct Version(pub u64);
+
+impl Version {
+    /// An inclusive lower limit on a valid sequence number.
```
thibault-martinez: You have lots of mentions of `sequence number` in the docs

DaughterOfMars: hmmm good point

bindings/kotlin/lib/iota_sdk/iota_sdk.kt
thibault-martinez, line 1: Not sure why you are generating a lib with the `ffi` suffix, we removed that

DaughterOfMars: I'm just calling `make kotlin; make python; make go`

thibault-martinez: Weird, doing that on your branch does not produce a lib with the suffix

thibault-martinez: I'll push it if you want

thibault-martinez (commented):
There is still a few places with `version: u64`, `version: Option<u64>`, etc... Maybe have another look just in case

bindings/kotlin/lib/iota_sdk/iota_sdk.kt
DaughterOfMars: deleting the kotlin build intermediates seems to have fixed it

thibault-martinez: You are still renaming it

crates/iota-sdk-ffi/src/types/version.rs
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2025 IOTA Stiftung
```
thibault-martinez:
```suggestion
// Copyright (c) 2026 IOTA Stiftung
```

crates/iota-sdk-ffi/src/types/crypto/move_authenticator.rs
```diff
@@ -45,7 +45,7 @@ impl MoveAuthenticator {
             call_args.into_iter().map(|v| v.0.clone()).collect(),
             type_args.into_iter().map(|v| v.0.clone()).collect(),
             **object_to_authenticate,
-            initial_shared_version,
+            iota_sdk::types::Version(initial_shared_version),
```
thibault-martinez: Just curious, why not change the parameter type? You did for other methods/ctors

crates/iota-sdk-types/src/version.rs
```diff
@@ -0,0 +1,233 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
```
thibault-martinez:
```suggestion
// Modifications Copyright (c) 2026 IOTA Stiftung
```

bindings/kotlin/lib/iota_sdk/iota_sdk.kt
thibault-martinez: Did it for you https://github.com/iotaledger/iota-rust-sdk/pull/521/commits/5e5f767ccb26fe85264ea03439e85ec68a9837b6

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -909,7 +909,7 @@ pub enum Input {
     Shared {
         object_id: ObjectId,
         #[cfg_attr(feature = "schemars", schemars(with = "crate::_schemars::U64"))]
```
Thoralf-M, line 911: Do we still need this schemars thing?

crates/iota-sdk-ffi/src/types/crypto/move_authenticator.rs
DaughterOfMars: just missed it :)

crates/iota-sdk-types/src/transaction/mod.rs
DaughterOfMars: nope!

DaughterOfMars: I moved them to the struct :)

crates/iota-sdk-transaction-builder/src/builder/gas_station.rs
```diff
@@ -193,32 +193,11 @@ struct GasObjectRef {
     /// The object id of this object.
     pub object_id: ObjectId,
     /// The version of this object.
-    #[serde(deserialize_with = "deserialize_readable_u64")]
-    pub version: Version,
+    pub version: u64,
```
thibault-martinez, line 197: How come it's not a Version anymore?

DaughterOfMars: It was never serialized like the version is, hence the custom deserialize function. I removed it on accident but realized now that it still needs to be there

crates/iota-sdk-types/src/version.rs
```diff
@@ -0,0 +1,241 @@
...
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[repr(transparent)]
+pub struct Version(
+    #[cfg_attr(feature = "serde", serde(with = "crate::_serde::ReadableDisplay"))]
+    #[cfg_attr(feature = "schemars", schemars(with = "crate::_schemars::U64"))]
+    pub u64,
+);
+
+impl Version {
```
thibault-martinez, line 54: Can you confirm that ALL of the constants, methods and additional Impl are required by the monorepo?

Thoralf-M (approved):

crates/iota-sdk-types/src/version.rs
DaughterOfMars: They all are, but `get_congested_version_suggested_gas_price` and the two associated consts are only used in one place, so it may be better to leave that out and reimplement it locally.

DaughterOfMars: Actually it's just the fn, not the consts. But they should be private like the core ones.

crates/iota-sdk-ffi/src/types/version.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    /// Returns the next version, or an error if overflow occurs.
+    pub fn next(&self) -> Result<Self> {
+        Ok(Self(self.0.next()?))
+    }
+
+    /// Returns the previous version, or an error if underflow occurs.
+    pub fn previous(&self) -> Result<Self> {
+        Ok(Self(self.0.previous()?))
+    }
```
thibault-martinez, line 121: What about increment and decrement?

thibault-martinez: I mean, I see it's mutable but can't we adapt it?

crates/iota-sdk-types/src/version.rs
```diff
@@ -0,0 +1,246 @@
...
+    const MIN_CONGESTED_FOR_GAS_PRICE_FEEDBACK: Self =
+        Self(Self::MAX_VALID_EXCL.0 + Self::CONGESTED_BASE_OFFSET_FOR_GAS_PRICE_FEEDBACK.0);
+
+    /// Create a new Version from a u64 value
+    pub const fn from_u64(value: u64) -> Self {
+        Self(value)
+    }
+
+    /// Get the underlying u64 value of this version
+    pub const fn as_u64(&self) -> u64 {
```
thibault-martinez, line 115: I would keep `value` for minimum monorepo changes. Fine to have an additional `as_u64` in the SDK though

DaughterOfMars: I do feel like it's important for future proofing to name things like this. Otherwise if we change the impl later the `value` method must break or we would have a weird `value_u128` or something

crates/iota-sdk-ffi/src/types/version.rs
DaughterOfMars: Whatchu mean, I'm a little confused?

thibault-martinez: You didn't add increment/decrement to the FFI, is there really no way to add them?

DaughterOfMars: Ohhhh, yeah that's because it can't use `&mut self`

thibault-martinez: Yeah but I was wondering if it made sense that instead it returns the incremented/decremented value but maybe not

crates/iota-sdk-types/src/version.rs
thibault-martinez: Alright

thibault-martinez (approved):

crates/iota-sdk-ffi/src/types/version.rs
DaughterOfMars: Yeah I mean that's just what the other methods do anyway

crates/iota-sdk-types/src/lib.rs
```diff
@@ -320,6 +322,13 @@ mod _serde {
     pub(crate) type ReadableBase64Encoded =
         ::serde_with::As<::serde_with::IfIsHumanReadable<Base64Encoded, ::serde_with::Bytes>>;
 
+    pub(crate) type ReadableDisplayOrDefault = ::serde_with::As<
```
thibault-martinez: What's the difference with `ReadableDisplay` then?

DaughterOfMars: This one just tries both when deserializing

Alex6323 (approved):

thibault-martinez (approved):

PR #524, opened 2026-01-21
feat(iota-sdk-types)!: use camelCase for GasCostSummary serialization

Thoralf-M:
Was like that in the monorepo and makes test fail without

### DIFF
--- a/crates/iota-sdk-types/src/gas.rs
+++ b/crates/iota-sdk-types/src/gas.rs
@@ -39,7 +39,11 @@
 ///                    u64 ; non-refundable-storage-fee
 /// ```
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "camelCase")
+)]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
 pub struct GasCostSummary {


thibault-martinez (approved):

DaughterOfMars:
Can we not? This seems weird to do

Thoralf-M:
> Can we not? This seems weird to do

That would be a breaking change in the JSON API then? https://github.com/iotaledger/iota/actions/runs/21204155249/job/60996999376?pr=9871

Alex6323 (approved):

DaughterOfMars:
> > Can we not? This seems weird to do
> 
> That would be a breaking change in the JSON API then? https://github.com/iotaledger/iota/actions/runs/21204155249/job/60996999376?pr=9871

I think we're going to break some things regardless

Thoralf-M:
> > > Can we not? This seems weird to do
> > 
> > 
> > That would be a breaking change in the JSON API then? [iotaledger/iota/actions/runs/21204155249/job/60996999376?pr=9871](https://github.com/iotaledger/iota/actions/runs/21204155249/job/60996999376?pr=9871)
> 
> I think we're going to break some things regardless

But here it's simple not to break and camelCase in JSON is not uncommon, so what's so weird about it?

DaughterOfMars:
> > > > Can we not? This seems weird to do
> > > 
> > > 
> > > That would be a breaking change in the JSON API then? [iotaledger/iota/actions/runs/21204155249/job/60996999376?pr=9871](https://github.com/iotaledger/iota/actions/runs/21204155249/job/60996999376?pr=9871)
> > 
> > 
> > I think we're going to break some things regardless
> 
> But here it's simple not to break and camelCase in JSON is not uncommon, so what's so weird about it?

It's weird that we're only doing this for one struct

thibault-martinez:
> > > Can we not? This seems weird to do
> > 
> > 
> > That would be a breaking change in the JSON API then? https://github.com/iotaledger/iota/actions/runs/21204155249/job/60996999376?pr=9871
> 
> I think we're going to break some things regardless

There is a big difference between breaking rust API and breaking RPC APIs. We can't break the latter.

Thoralf-M:
> It's weird that we're only doing this for one struct

Want me to change it for all?

DaughterOfMars:
> > It's weird that we're only doing this for one struct
> 
> Want me to change it for all?

I dunno. I don't like that there's no way to choose casing for each format. I feel like if possible I would impl camel for all when serializing json but not for everything else

Thoralf-M:
> I dunno. I don't like that there's no way to choose casing for each format. I feel like if possible I would impl camel for all when serializing json but not for everything else

But realistically to what else is it serialized where it makes a difference?

DaughterOfMars:
> > I dunno. I don't like that there's no way to choose casing for each format. I feel like if possible I would impl camel for all when serializing json but not for everything else
> 
> But realistically to what else is it serialized where it makes a difference?

TOML, for instance

Thoralf-M:
> TOML, for instance

In which case do you see the types being serialised to TOML?

DaughterOfMars:
> > TOML, for instance
> 
> In which case do you see the types being serialised to TOML?

It's not about whether we do it. This is an SDK so it's supposed to provide types for general use. I don't know what use cases someone else may have for serializing to TOML. Personally, I think serde should allow passing formatting configs to serializers to solve this problem, but unfortunately they do not.

Thoralf-M:
> > > TOML, for instance
> > 
> > 
> > In which case do you see the types being serialised to TOML?
> 
> It's not about whether we do it. This is an SDK so it's supposed to provide types for general use. I don't know what use cases someone else may have for serializing to TOML. Personally, I think serde should allow passing formatting configs to serializers to solve this problem, but unfortunately they do not.

So what? Since serde doesn't provide what you want, we have to choose one thing here. Users can always convert the casing afterwards, this would also be an option for the monorepo I guess, keeping the camel_case and add custom serialization there to convert to camelCase, but I think this could get quite annoying for us where with camelCase I don't see any extra difficulties

PR #526, opened 2026-01-21
feat(iota-sdk-types): add ConsensusCommitPrologueV2 transaction type

piotrm50:
## Summary

This PR adds `ConsensusCommitPrologueV2`, which extends the V1 version with an `additional_state_digest` field. This digest is used to detect forking bugs as early as possible by capturing any additional state computed by the consensus handler.

## Changes

- Added `ConsensusCommitPrologueV2` struct in `crates/iota-sdk-types/src/transaction/mod.rs`
- Added the new transaction variant to `TransactionKind` enum
- Updated serialization support in `crates/iota-sdk-types/src/transaction/serialization.rs`
- Exported `ConsensusCommitPrologueV2` from `crates/iota-sdk-types/src/lib.rs`
- Uses `Digest` type directly for `additional_state_digest` field for better compatibility

## Implementation Details

The implementation follows the same pattern as `ConsensusCommitPrologueV1`:
- All fields from V1 are preserved
- New `additional_state_digest: Digest` field added at the end
- Proper BCS serialization format documented
- Full serde and schemars support for both human-readable and binary formats

## Testing

- Code compiles successfully with `cargo check`
- All existing tests pass


crates/iota-sdk-types/src/digest.rs
```diff
@@ -225,6 +225,135 @@ impl std::error::Error for DigestParseError {}
...
+}
+
+impl_digest!(CheckpointDigest);
+impl_digest!(CheckpointContentsDigest);
+impl_digest!(TransactionDigest);
+impl_digest!(TransactionEffectsDigest);
+impl_digest!(TransactionEventsDigest);
+impl_digest!(ObjectDigest);
+impl_digest!(ConsensusCommitDigest);
+impl_digest!(EffectsAuxiliaryDataDigest);
```
Thoralf-M: Why add them if they aren't used anywhere?

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -209,6 +209,8 @@ pub enum TransactionKind {
     EndOfEpoch(Vec<EndOfEpochTransactionKind>),
     /// Randomness update
     RandomnessStateUpdate(RandomnessStateUpdate),
+    /// V2 consensus commit update with additional state digest
+    ConsensusCommitPrologueV2(ConsensusCommitPrologueV2),
```
DaughterOfMars, line 211: Is there a reason these aren't ordered after the previous version?

crates/iota-sdk-types/src/digest.rs
piotrm50: Ah, I started on an old revision before restructuring and I guess this was there back there and I must've rebased incorrectly. Removing it.

crates/iota-sdk-types/src/transaction/mod.rs
piotrm50: The tests were failing if I added them in the middle of a struct. There was a few instances of this, so I just reordered all of them to be added at the end. I learned the hard lesson of putting things in the middle of struct before in the monorepo :sweat_smile:

thibault-martinez: Can only add to the end of enums meant to be converted to BCS otherwise you shift the IDs

DaughterOfMars: oof

DaughterOfMars (approved):

thibault-martinez (approved):

PR #527, opened 2026-01-22
feat(iota-sdk-types): add ConsensusCommitPrologueV2 transaction type

piotrm50:
## Summary
Created this PR again to merge it again at a later point in time when the final source types in monorepo are established.

This PR adds `ConsensusCommitPrologueV2`, which extends the V1 version with an `additional_state_digest` field. This digest is used to detect forking bugs as early as possible by capturing any additional state computed by the consensus handler.

## Changes

- Added `ConsensusCommitPrologueV2` struct in `crates/iota-sdk-types/src/transaction/mod.rs`
- Added the new transaction variant to `TransactionKind` enum
- Updated serialization support in `crates/iota-sdk-types/src/transaction/serialization.rs`
- Exported `ConsensusCommitPrologueV2` from `crates/iota-sdk-types/src/lib.rs`
- Uses `Digest` type directly for `additional_state_digest` field for better compatibility

## Implementation Details

The implementation follows the same pattern as `ConsensusCommitPrologueV1`:
- All fields from V1 are preserved
- New `additional_state_digest: Digest` field added at the end
- Proper BCS serialization format documented
- Full serde and schemars support for both human-readable and binary formats

## Testing

- Code compiles successfully with `cargo check`
- All existing tests pass


piotrm50:
We decided to put this on hold for now until our congestion control implements similar features that we include digests for in the new prologue version.

PR #529, opened 2026-01-26
feat(types): enhance Address with system package constants and methods

DaughterOfMars:
## Description

- Add GENESIS_BRIDGE and STARDUST system package constants
- Rename from_u8 to from_u16 to support larger package IDs
- Add is_system_package() method
- Add const fn from_object_id conversion
- Rename into_inner/inner to into_bytes/bytes for consistency

### DIFF
--- a/crates/iota-sdk-types/src/address.rs
+++ b/crates/iota-sdk-types/src/address.rs
@@ -2,6 +2,8 @@
 // Modifications Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::ObjectId;
+
 /// Unique identifier for an Account on the IOTA blockchain.
 ///
 /// An `Address` is a 32-byte pseudonymous identifier used to uniquely identify
@@ -69,18 +71,33 @@ pub struct Address(
 impl Address {
     pub const LENGTH: usize = 32;
     pub const ZERO: Self = Self([0u8; Self::LENGTH]);
-    pub const STD: Self = Self::from_u8(1);
-    pub const FRAMEWORK: Self = Self::from_u8(2);
-    pub const SYSTEM: Self = Self::from_u8(3);
+    pub const STD: Self = Self::from_u16(1);
+    pub const FRAMEWORK: Self = Self::from_u16(2);
+    pub const SYSTEM: Self = Self::from_u16(3);
+    pub const GENESIS_BRIDGE: Self = Self::from_u16(0xb);
+    pub const STARDUST: Self = Self::from_u16(0x107a);
 
     pub const fn new(bytes: [u8; Self::LENGTH]) -> Self {
         Self(bytes)
     }
 
-    pub(crate) const fn from_u8(byte: u8) -> Self {
+    pub const fn from_u16(suffix: u16) -> Self {
         let mut address = Self::ZERO;
-        address.0[31] = byte;
+        let [hi, lo] = suffix.to_be_bytes();
+        address.0[Address::LENGTH - 2] = hi;
+        address.0[Address::LENGTH - 1] = lo;
         address
+    }
+
+    pub fn is_system_package(&self) -> bool {
+        [
+            Self::STD,
+            Self::FRAMEWORK,
+            Self::SYSTEM,
+            Self::GENESIS_BRIDGE,
+            Self::STARDUST,
+        ]
+        .contains(self)
     }
 
     #[cfg(feature = "rand")]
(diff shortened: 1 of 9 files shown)


crates/iota-sdk-types/src/object_id.rs
```diff
@@ -26,13 +26,13 @@ use super::Address;
...
-pub struct ObjectId(Address);
+pub struct ObjectId(pub(crate) Address);
 
 impl ObjectId {
     pub const LENGTH: usize = Address::LENGTH;
     pub const ZERO: Self = Self(Address::ZERO);
-    pub const SYSTEM: Self = Self(Address::from_u8(5));
-    pub const CLOCK: Self = Self(Address::from_u8(6));
+    pub const SYSTEM: Self = Self(Address::from_u16(5));
+    pub const CLOCK: Self = Self(Address::from_u16(6));
```
Thoralf-M, line 35: If you're on it, can you also add '0x403': 'IOTA Denylist Object', ?

crates/iota-sdk-types/src/lib.rs
```diff
@@ -186,6 +186,25 @@ pub use version::Version;
 #[cfg(all(test, feature = "serde", feature = "proptest"))]
 mod serialization_proptests;
 
+/// Returns the next array in byte-increasing order.
+pub const fn next_lexicographical_array<const N: usize>(array: &[u8; N]) -> [u8; N] {
```
thibault-martinez, line 190: Can `Digest::next_lexicographical` call this? Is it the exact same impl?

crates/iota-sdk-types/src/address.rs
```diff
@@ -94,19 +111,29 @@ impl Address {
         Self::new(buf)
     }
 
+    #[cfg(all(feature = "rand", not(target_arch = "wasm32")))]
+    #[cfg_attr(doc_cfg, doc(cfg(all(feature = "rand", not(target_arch = "wasm32")))))]
+    pub fn random() -> Self {
+        Self::generate(rand_core::OsRng)
```
thibault-martinez, line 118: Let's just remove `generate`? Why are we keeping it if we're moving it from ffi/bindings?

crates/iota-sdk-types/src/address.rs
```diff
@@ -69,20 +71,35 @@ pub struct Address(
...
+    pub const SYSTEM: Self = Self::from_u16(3);
+    pub const GENESIS_BRIDGE: Self = Self::from_u16(0xb);
+    pub const STARDUST: Self = Self::from_u16(0x107a);
 
     pub const fn new(bytes: [u8; Self::LENGTH]) -> Self {
         Self(bytes)
     }
 
-    pub(crate) const fn from_u8(byte: u8) -> Self {
+    pub const fn from_u16(suffix: u16) -> Self {
```
Thoralf-M, line 85: We should have doc comments for such methods
```suggestion
    /// Creates an `Address` from a `u16` suffix by setting the last two bytes.
    pub const fn from_u16(suffix: u16) -> Self {
```

crates/iota-sdk-types/src/object_id.rs
DaughterOfMars: I will when I do ObjectID :)

crates/iota-sdk-types/src/lib.rs
DaughterOfMars: Yes, all of the byte array types will eventually use this

crates/iota-sdk-types/src/address.rs
DaughterOfMars: It's needed for wasm

crates/iota-sdk-types/src/address.rs
DaughterOfMars: Yeah, it wasn't originall supposed to be a pub fn so I never wrote one. Thanks

crates/iota-sdk-types/src/address.rs
thibault-martinez: How come wasm need our Address to impl a generate method?

crates/iota-sdk-types/src/lib.rs
thibault-martinez: Why not already use this one for Digest? We shouldn't introduce code duplication. At least open an issue if you plan on doing something about it later.

Thoralf-M (approved):

crates/iota-sdk-types/src/address.rs
DaughterOfMars: Because the `random` fn cannot be used in wasm

crates/iota-sdk-types/src/lib.rs
DaughterOfMars: Alright, sure

crates/iota-sdk-types/src/object_id.rs
thibault-martinez, line 35: Can you explain this change please?

DaughterOfMars: It was needed to support the stardust address and object ID just reuses the method

thibault-martinez (approved):

PR #530, opened 2026-01-26
refactor: reverse dependency between graphql-client and transaction-builder

DaughterOfMars:
## Summary

This PR reverses the dependency relationship between `iota-sdk-graphql-client` and `iota-sdk-transaction-builder` crates.

### Code Migration
- Moved `ClientMethods` implementation for GraphQL `Client` from `transaction-builder` to `graphql-client` crate
- Moved `WaitForTx` enum definition to `transaction-builder` crate (re-exported from `graphql-client`)
- Moved integration tests that require a live GraphQL server to `graphql-client` crate

### Test Utilities
- Added `TestClient` stub implementation in `transaction-builder` for use in doc tests
- `TestClient` is gated behind the `test-client` feature and provides stub implementations of `ClientMethods`
- Updated doc examples to use `TestClient` with `no_run` attributes

## Benefits

- The `transaction-builder` crate is now more standalone and can be used without pulling in GraphQL dependencies
- Better separation of concerns: the trait definition (`ClientMethods`) lives in `transaction-builder`, while the concrete implementation for GraphQL lives in `graphql-client`
- Enables future alternative client implementations without modifying `transaction-builder`

## Testing

- `cargo check --workspace` passes
- `cargo nextest run --all-features --workspace --exclude iota-sdk-graphql-client` passes


crates/iota-sdk-transaction-builder/src/lib.rs
```diff
@@ -427,200 +338,4 @@ mod tests {
 
         tx.finish().unwrap();
     }
-
-    #[tokio::test]
-    async fn test_transfer_obj_execution() {
```
thibault-martinez, line 506: It feels very weird moving the TxBuilder tests in the client crate tbh. I don't think we would have a cycle if we import the client only in the dev-deps

crates/integration-tests/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Nothing to see here...
```
Thoralf-M, line 4: Why have this file then?

DaughterOfMars: It's required :c

Thoralf-M: Ah because we need something in the crate, but then another comment would be better

DaughterOfMars: It's just a test crate. Most people leave these files empty

crates/integration-tests/test_example_v1/Move.lock
Thoralf-M, line 1: Should we commit the lock files?

DaughterOfMars: Removed them

Thoralf-M (approved):

thibault-martinez (changes requested):
Will review when I get a chance.

crates/integration-tests/tests/graphql_ptbs.rs
```diff
@@ -0,0 +1,300 @@
...
+        .unwrap();
+    serde_json::from_str(&data).unwrap()
+}
+
+/// Generate a random private key and its corresponding address
+fn helper_address_pk() -> (Address, Ed25519PrivateKey) {
+    let pk = Ed25519PrivateKey::generate(rand::thread_rng());
+    let address = pk.public_key().derive_address();
+    (address, pk)
+}
```
Alex6323, line 45: CMIIAW, but `pk` is more often used for "public key" and `sk` for a secrect/private key? This always confuses me and others probably as well. So should we change the suffix to `_priv_key` or `_sk` or `_private_key`?

Alex6323: okay, just asked Claude, seems like `pk` is more often used **in general** to mean "private key", but it depends on the bubble. The ambiguity is real, though, and for the Rust bubble I am actually correct and `sk` (private key)/`pk` (public key) is preferred to disambiguate.

crates/integration-tests/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "integration-tests"
+version = "0.1.0"
```
Alex6323: Since it's not published should this also be "0.0.0"?

DaughterOfMars: sure, changed

crates/integration-tests/tests/graphql_ptbs.rs
DaughterOfMars: I don't think this is really important. It's just a test, and I just moved it anyway

crates/integration-tests/tests/graphql_ptbs.rs
```diff
@@ -0,0 +1,300 @@
...
+
+    (tx, address, pk, coins)
+}
+
+/// Wait for the transaction to be finalized and indexed. This queries the
+/// GraphQL server until it retrieves the requested transaction.
+async fn wait_for_tx(client: &Client, digest: Digest) {
+    while !client.is_tx_finalized(digest).await.unwrap() {
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+    }
```
Thoralf-M: Should limit the loop to some max number to prevent an infinite loop in case a tx never gets finalized

crates/iota-sdk-graphql-client/src/client_methods.rs
```diff
@@ -0,0 +1,132 @@
...
+        }))
+    }
+
+    async fn dry_run_tx(
+        &self,
+        tx: &Transaction,
+        skip_checks: bool,
+    ) -> Result<Self::DryRunResult, Self::Error> {
+        let Transaction::V1(tx) = &tx else {
+            unimplemented!("a new enum variant was added and needs to be handled")
```
Thoralf-M:
```suggestion
            unimplemented!("a new Transaction enum variant was added and needs to be handled")
```

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -1,16 +1,37 @@
...
+///
+/// Users should almost always use [`WaitForTx::Finalized`] (the default), as
+/// clients may interact with the indexer and not the fullnode directly.
+/// Using [`WaitForTx::IndexedOnNode`] only guarantees the transaction is
+/// indexed on the fullnode (meaning you can submit transactions that reference
+/// objects created by this transaction), but subsequent queries using the
+/// transaction ID can still fail until the transaction is indexed on the
+/// indexer.
+#[derive(Default)]
+pub enum WaitForTx {
```
Thoralf-M, line 22: Why was `#[non_exhaustive]` removed?

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -332,3 +261,113 @@ impl<T: ClientMethods> ClientMethods for std::sync::Arc<T> {
...
+    ///
+    /// This client is useful for testing scenarios where a real client
+    /// connection is not needed. All methods return default or empty
+    /// values.
+    #[derive(Clone, Copy, Debug, Default)]
+    pub struct TestClient;
+
+    /// Error type for [`TestClient`].
+    #[derive(Clone, Debug)]
+    pub struct TestClientError(pub String);
```
Thoralf-M, line 367: Any reason not to use `#[derive(thiserror::Error)]`?

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
DaughterOfMars: Users don't need to match on this so there's no reason for it really

Thoralf-M: Even if users don't need to match on it, adding a new variant will be breaking change

DaughterOfMars: Yeah but I don't think there should ever be a new variant, right?

thibault-martinez: famous last words

DaughterOfMars: alright, I added it although I think it's overkill

crates/iota-sdk-transaction-builder/src/lib.rs
```diff
@@ -14,16 +14,16 @@
...
 //! [with_client](TransactionBuilder::with_client) method, the resulting builder
 //! will use it to find and validate provided IDs.
 //!
 //! ### Example with Client Resolution
 //!
-//! ```
+//! ```no_run
 //! # use std::str::FromStr;
-//! use iota_graphql_client::Client;
+//! # use iota_sdk_transaction_builder::TestClient;
```
thibault-martinez, line 26: Why do we actually need a TestClient if we end up not running any of these doctests?

thibault-martinez (commented):
There is two mentions of the graphql client in the builder crate, please check if they need to be removed

crates/iota-sdk-transaction-builder/Cargo.toml
```diff
@@ -16,15 +20,14 @@ base64ct = { workspace = true, features = ["std"] }
 bcs.workspace = true
 derive_more = { workspace = true, features = ["from"] }
 primitive-types.workspace = true
-reqwest.workspace = true
+reqwest = { workspace = true, features = ["rustls-tls", "json"] }
```
thibault-martinez: rustls-tls needed?

crates/iota-sdk-graphql-client/src/client_methods.rs
```diff
@@ -0,0 +1,132 @@
...
+            TransactionMetadata {
+                gas_budget: (tx.gas_payment.budget > 0).then_some(tx.gas_payment.budget),
+                gas_objects: (!gas_objects.is_empty()).then_some(gas_objects),
+                gas_price: Some(tx.gas_payment.price),
+                gas_sponsor: Some(tx.gas_payment.owner),
+                sender: Some(tx.sender),
+            },
+        )
+        .await
+    }
```
thibault-martinez, line 118: Is this a conflict resolution mistake? Isn't it supposed to called the self.dry_run_tx ?

crates/iota-sdk-transaction-builder/src/lib.rs
DaughterOfMars: Well we should be able to run them

thibault-martinez: Then why are you making them all `no_run`?

thibault-martinez:
> There is two mentions of the graphql client in the builder crate, please check if they need to be removed

Don't forget about this please

crates/iota-sdk-transaction-builder/src/lib.rs
DaughterOfMars: tbh I don't remember 😆 maybe we can just remove it. I'll look into it

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -522,14 +522,14 @@ impl<C, L> TransactionBuilder<C, L> {
     ///
     /// # Example
     ///
-    /// ```rust
```
thibault-martinez, line 525: Was the rust part not needed to properly run?

Alex6323 (approved):

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: Pretty sure it's implicit

thibault-martinez: Alright

thibault-martinez (approved):

PR #536, opened 2026-01-27
fix(iota-sdk-transaction-builder): always wait for gas station tx to be indexed

Thoralf-M:
Otherwise trying to get the effects can result in this error:
Missing transaction for digest 2KZUg8faUSpgtp4MS4dXThbN9MoAn6HcDT7m4q9Fn4h3
Failed here https://github.com/iotaledger/iota-rust-sdk/actions/runs/21392276765/job/61581830799?pr=516


Alex6323 (approved):

DaughterOfMars (changes requested):
Let me investigate whether we can actually convert IotaTransactionBlockEffects to TransactionEffects before we do this, please

DaughterOfMars (approved):
:C

thibault-martinez (approved):

PR #538, opened 2026-01-28
feat(iota-types): Better Address parsing

DaughterOfMars:
## Description

Improves address parsing by allowing hex strings with or without the `0x` prefix, and improves the `AddressParseError`.

### DIFF
--- a/crates/iota-sdk-types/src/address.rs
+++ b/crates/iota-sdk-types/src/address.rs
@@ -161,7 +161,7 @@ impl Address {
     pub fn from_bytes<T: AsRef<[u8]>>(bytes: T) -> Result<Self, AddressParseError> {
         let bytes = bytes.as_ref();
         <[u8; Self::LENGTH]>::try_from(bytes)
-            .map_err(|_| AddressParseError::InvalidLength {
+            .map_err(|_| AddressParseError::InvalidByteLength {
                 actual: bytes.len(),
             })
             .map(Self)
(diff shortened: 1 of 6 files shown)


thibault-martinez (commented):
Can you add a little test for both please 🙏🏻 ?

DaughterOfMars:
> Can you add a little test for both please 🙏🏻 ?

image

crates/iota-sdk-types/src/address.rs
```diff
@@ -109,12 +109,11 @@ impl Address {
...
-
-        let hex = &hex[2..];
+        let hex = if hex.starts_with(b"0x") {
+            &hex[2..]
+        } else {
+            hex
+        };
 
         // If the string is too short we'll need to pad with 0's
         if hex.len() < Self::LENGTH * 2 {
```
thibault-martinez, line 122: Should it be an error if the length was uneven? Padding an incomplete byte feels weird

thibault-martinez (approved):

crates/iota-sdk-types/src/address.rs
DaughterOfMars: The input is a string, so I dunno if this really applies, but either way I feel like it's probably fine

thibault-martinez: You still need two chars to form one byte. But yeah probably fine.

Thoralf-M (approved):

crates/iota-sdk-types/src/address.rs
```diff
@@ -109,12 +109,11 @@ impl Address {
 
     pub fn from_hex<T: AsRef<[u8]>>(hex: T) -> Result<Self, AddressParseError> {
```
Thoralf-M, line 113: Might also want to add a comment mentioning the padding that happens if the string is shorter than full length

DaughterOfMars: done

PR #540, opened 2026-01-28
feat(types): enhance ObjectId with system constants and methods

DaughterOfMars:
## Description

- Add GENESIS_BRIDGE and STARDUST system package constants
- Add const fn from_address conversion
- Add from_bytes constructor
- Rename into_inner/inner to into_bytes/bytes for consistency


thibault-martinez:
Change the base branch
EDIT: I did

crates/iota-sdk-types/src/address.rs
```diff
@@ -473,7 +512,7 @@ mod tests {
     #[test]
     #[cfg(feature = "serde")]
     fn formats() {
```
Thoralf-M, line 541: Maybe add a test like this (tested this to match the monorepo behavior)
```Rust
    fn serde_json_from_str() {
        let json_hex = "\"0x1234567890abcdef1234567890abcdefca843279e3427144cead5e4d5999a3d0\"";
        serde_json::from_str::<Address>(json_hex).unwrap();
        let json_hex = "\"1234567890abcdef1234567890abcdefca843279e3427144cead5e4d5999a3d0\"";
        serde_json::from_str::<Address>(json_hex).unwrap();
        let json_hex = "\"0x123\"";
        serde_json::from_str::<Address>(json_hex).unwrap();
        let json_hex = "\"123\"";
        serde_json::from_str::<Address>(json_hex).unwrap_err();
    }
```

Thoralf-M (approved):

crates/iota-sdk-types/src/object_id.rs
```diff
@@ -31,30 +31,114 @@ pub struct ObjectId(pub(crate) Address);
...
+    pub const FRAMEWORK_PACKAGE: Self = Self(Address::FRAMEWORK);
+    pub const SYSTEM_PACKAGE: Self = Self(Address::SYSTEM);
+    pub const GENESIS_BRIDGE_PACKAGE: Self = Self(Address::GENESIS_BRIDGE);
+    pub const STARDUST_PACKAGE: Self = Self(Address::STARDUST);
+    pub const SYSTEM: Self = Self::from_u16(5);
+    pub const CLOCK: Self = Self::from_u16(6);
+    pub const AUTHENTICATOR_STATE: Self = Self::from_u16(7);
+    pub const RANDOMNESS_STATE: Self = Self::from_u16(8);
+    pub const GENESIS_IOTA_BRIDGE: Self = Self::from_u16(9);
+    pub const DENY_LIST: Self = Self::from_u16(0x403);
```
thibault-martinez, line 45: Did it not make sense to have this on Address as well?

thibault-martinez (approved):

crates/iota-sdk-types/src/object_id.rs
DaughterOfMars: These are objects so no

thibault-martinez: So what? They still have an address that is their ObjectId, they are defined as both in the monorepo, because they are both.

DaughterOfMars: I guess. I dunno 🤷‍♀️

PR #542, opened 2026-01-28
feat(transaction-builder): Add `TransactionSigner::address` and use it in `TransactionBuilder::execute` functions

### LINKED ISSUE #539: Make trait `TransactionSigner` expose an `address(&self) -> Address` method
### Summary

Add a new method `fn address(&self) -> Address` to trait `TransactionSigner` and rely on it to set `TransactionBuilder`'s sender and sponsor addresses, removing the need to pass an address to `TransactionBuilder::new`.

When building a transaction through the SDK's `TransactionBuilder` it isn't always possible to provide the actual sender's address to `TransactionBuilder::new`. One simple instance of this is Gas Station execution: the transaction's sender is not the user but the IOTA Gas Station itself, whose address isn't necessarily known to the user.

Sender and sponsor address can only be known for certain by their corresponding signers, but at the moment a signer has no way of providing this information before performing a signature.

This change would make `TransactionBuilder` simpler to use and less error prone, as it won't be possible to provide mismatched signatures.

UMR1352:
# Description
- Add method `fn address(&self) -> Address` to trait `TransactionSigner`
- Add `TransactionBuilder::new_with_sender`
- Add `MissingSender` variant to `transaction_builder::Error`
- Change `TransactionBuilder::new(sender: Address)` to `TransactionBuilder::new()`

# Related Issues
Closes #539

### DIFF
--- a/crates/iota-sdk-ffi/src/transaction_builder/builder.rs
+++ b/crates/iota-sdk-ffi/src/transaction_builder/builder.rs
@@ -53,8 +53,8 @@ impl TransactionBuilder {
 impl TransactionBuilder {
     /// Create a new transaction builder and initialize its elements to default.
     #[uniffi::constructor]
-    pub fn new(sender: &Address) -> Self {
-        Self(iota_sdk::transaction_builder::TransactionBuilder::new(**sender).into())
+    pub fn new() -> Self {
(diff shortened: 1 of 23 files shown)


thibault-martinez:
https://github.com/iotaledger/iota-rust-sdk/pull/549 should be enough

PR #549, opened 2026-01-30
feat(transaction-builder): Add `set_sender` method

### LINKED ISSUE #539: Make trait `TransactionSigner` expose an `address(&self) -> Address` method

DaughterOfMars:
## Description

Adds a `set_sender` method to the transaction builder to allow for special use-cases.

Closes #539

### DIFF
--- a/bindings/go/iota_sdk/iota_sdk.go
+++ b/bindings/go/iota_sdk/iota_sdk.go
@@ -5160,6 +5160,15 @@ func uniffiCheckChecksums() {
 	}
 	{
 	checksum := rustCall(func(_uniffiStatus *C.RustCallStatus) C.uint16_t {
+		return C.uniffi_iota_sdk_ffi_checksum_method_clienttransactionbuilder_set_sender()
+	})
+	if checksum != 20194 {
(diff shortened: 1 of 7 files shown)


UMR1352 (approved):

Thoralf-M (changes requested):
Also add it to the bindings

Thoralf-M (approved):

thibault-martinez (approved):

PR #550, opened 2026-02-02
feat(iota-sdk-types): derive AddAssign and SubAssign for GasCostSummary

Thoralf-M:
AddAssign existed in the monorepo and might be useful https://github.com/iotaledger/iota/pull/9871#discussion_r2712408284

### DIFF
--- a/crates/iota-sdk-types/src/gas.rs
+++ b/crates/iota-sdk-types/src/gas.rs
@@ -121,6 +121,12 @@ impl std::ops::AddAssign<&Self> for GasCostSummary {
     }
 }
 
+impl std::ops::AddAssign for GasCostSummary {
+    fn add_assign(&mut self, other: Self) {
+        *self += &other;
+    }
+}
+
 impl std::ops::SubAssign<&Self> for GasCostSummary {
     fn sub_assign(&mut self, other: &Self) {
         self.computation_cost -= other.computation_cost;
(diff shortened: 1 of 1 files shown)


DaughterOfMars (approved):

crates/iota-sdk-types/src/gas.rs
```diff
@@ -111,6 +111,26 @@ impl GasCostSummary {
     }
 }
 
+impl std::ops::AddAssign<&Self> for GasCostSummary {
```
DaughterOfMars, line 114: Shouldn't we support both refs and owned?

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/550/commits/9fdf7e3a349796f2a72a62cacb3a037c97feb6bf

crates/iota-sdk-types/src/gas.rs
```diff
@@ -121,6 +121,12 @@ impl std::ops::AddAssign<&Self> for GasCostSummary {
     }
 }
 
+impl std::ops::AddAssign for GasCostSummary {
```
DaughterOfMars: We can derive this, no?

Thoralf-M: Yes, but there seems to be no way to derive with a ref https://github.com/iotaledger/iota-rust-sdk/pull/550/commits/9c779b1971ec50731c7aba9cb0553857e38911a2

DaughterOfMars (approved):

thibault-martinez (approved):

PR #551, opened 2026-02-02
feat(types): enhance Digest with additional methods and constants

DaughterOfMars:
## Description

Improves the Digest to prepare for core integration

- Add GENESIS constant for genesis digest marker
- Add UpperHex formatting support
- Add const fn improvements (inner, as_bytes, from_bytes)
- Add is_alive, is_deleted, is_wrapped methods for ObjectDigest


crates/iota-sdk-types/src/digest.rs
```diff
@@ -229,31 +291,4 @@ mod tests {
         let d = s.parse::<Digest>().unwrap();
         assert_eq!(digest, d);
     }
-
-    #[test]
-    fn test_lexical_order() {
```
Thoralf-M, line 234: Why delete the test?

DaughterOfMars: it was just moved to the lib.rs

Thoralf-M: In lib.rs it's not actually testing the methods from Digest

DaughterOfMars: it's not actually testing that though...

Thoralf-M: Wdym? It's calling `next_lexicographical()` on a `Digest` here, in lib.rs it's never creating a `Digest`

DaughterOfMars: I guess, but that's just calling this method and these tests are about this method

crates/iota-sdk-types/src/object_id.rs
```diff
@@ -163,6 +163,15 @@ impl ObjectId {
         Self::new(crate::next_lexicographical_array(self.bytes()))
     }
 
+    /// Returns the next object id in byte-increasing order, or `None` if the
+    /// result would overflow.
+    pub const fn next_lexicographical_opt(&self) -> Option<Self> {
+        match crate::next_lexicographical_array_opt(self.bytes()) {
+            Some(val) => Some(Self::new(val)),
+            None => None,
+        }
```
Thoralf-M, line 172:
```
`std::option::Option::<T>::map` is not yet stable as a const fn
```
 sad

Thoralf-M (approved):

crates/iota-sdk-types/src/lib.rs
```diff
@@ -205,6 +205,27 @@ pub const fn next_lexicographical_array<const N: usize>(array: &[u8; N]) -> [u8;
...
+    let mut i = N;
+    let mut wrapped = true;
+
+    while i > 0 {
+        i -= 1;
+        let (new_byte, overflow) = next[i].overflowing_add(1);
+        next[i] = new_byte;
+
+        if !overflow {
+            wrapped = false;
```
thibault-martinez: Why not just return?

DaughterOfMars: good point

crates/iota-sdk-types/src/lib.rs
```diff
@@ -205,6 +205,25 @@ pub const fn next_lexicographical_array<const N: usize>(array: &[u8; N]) -> [u8;
...
+    let mut next = *array;
+    let mut i = N;
+
+    while i > 0 {
+        i -= 1;
+        let (new_byte, overflow) = next[i].overflowing_add(1);
+        next[i] = new_byte;
+
+        if !overflow {
+            return Some(next);
```
thibault-martinez, line 220: Shouldn't we be doing the opposite?

DaughterOfMars: No, this is the early return. If we get through all the bytes and all of them overflow, that means the whole thing overflowed

thibault-martinez: Ah, sure

crates/iota-sdk-types/src/digest.rs
```diff
@@ -84,15 +108,44 @@ impl Digest {
...
+    pub const fn next_lexicographical_opt(&self) -> Option<Self> {
+        match crate::next_lexicographical_array_opt(&self.0) {
+            Some(val) => Some(Self(val)),
+            None => None,
+        }
+    }
+
+    /// Returns whether the digest represents an object that is neither deleted
+    /// nor wrapped
+    pub fn is_alive(&self) -> bool {
```
thibault-martinez: Just thinking out loud: maybe we should name them `is_object_alive` etc, these came from the ObjectDigest, I don't know if it makes a lot of sense for other kinds of digest

DaughterOfMars: That makes sense I think

thibault-martinez (approved):

PR #552, opened 2026-02-02
feat(graphql): Add `request_and_wait_for_finalized` for better waiting on faucet funds

DaughterOfMars:
## Description

Adds a new function which combines `request_and_wait` and waiting for the sending transactions to be finalized.


Thoralf-M (changes requested):
Bindings examples should also use the new method

Thoralf-M (approved):

thibault-martinez (approved):

PR #555, opened 2026-02-03
feat(iota-sdk-types): derive Hash for Event and TransactionEvents

Thoralf-M:
Monorepo Event struct also derived Hash
Also changed GasCostSummary to make use of safe math and added is_system_epoch_info_event methods to the Event

### DIFF
--- a/crates/iota-sdk-types/src/events.rs
+++ b/crates/iota-sdk-types/src/events.rs
@@ -59,15 +59,9 @@ pub struct Event {
 }
 
 impl Event {
-    const IOTA_SYSTEM_INNER: Identifier = Identifier::from_static("iota_system_state_inner");
-    const SYSTEM_EPOCH_INFO_EVENT_V1: Identifier =
-        Identifier::from_static("SystemEpochInfoEventV1");
-    const SYSTEM_EPOCH_INFO_EVENT_V2: Identifier =
-        Identifier::from_static("SystemEpochInfoEventV2");
-
     fn is_system_epoch_info_event_type(&self, name: Identifier) -> bool {
         self.type_.address() == Address::SYSTEM
-            && *self.type_.module() == Self::IOTA_SYSTEM_INNER
+            && *self.type_.module() == Identifier::IOTA_SYSTEM_STATE_INNER_MODULE
             && *self.type_.name() == name
     }
 
@@ -76,10 +70,10 @@ impl Event {
     }
 
     pub fn is_system_epoch_info_event_v1(&self) -> bool {
-        self.is_system_epoch_info_event_type(Self::SYSTEM_EPOCH_INFO_EVENT_V1)
+        self.is_system_epoch_info_event_type(Identifier::SYSTEM_EPOCH_INFO_EVENT_V1)
     }
 
     pub fn is_system_epoch_info_event_v2(&self) -> bool {
-        self.is_system_epoch_info_event_type(Self::SYSTEM_EPOCH_INFO_EVENT_V2)
+        self.is_system_epoch_info_event_type(Identifier::SYSTEM_EPOCH_INFO_EVENT_V2)
     }
 }
--- a/crates/iota-sdk-types/src/events.rs
+++ b/crates/iota-sdk-types/src/events.rs
@@ -90,8 +90,8 @@ impl Event {
     ///
     /// This is not fully random but designed to produce valid, varied test
     /// data.
-    #[cfg(feature = "rand")]
-    #[cfg_attr(doc_cfg, doc(cfg(feature = "rand")))]
+    #[cfg(all(feature = "rand", not(target_arch = "wasm32")))]
+    #[cfg_attr(doc_cfg, doc(cfg(all(feature = "rand", not(target_arch = "wasm32")))))]
     pub fn random() -> Self {
         use rand_core::{OsRng, RngCore};
 
@@ -137,7 +137,7 @@ impl Event {
 }
 
 #[cfg(test)]
-#[cfg(feature = "rand")]
+#[cfg(all(feature = "rand", not(target_arch = "wasm32")))]
 mod tests {
     use super::*;
 
--- a/crates/iota-sdk-types/src/events.rs
+++ b/crates/iota-sdk-types/src/events.rs
@@ -58,97 +58,22 @@ pub struct Event {
     pub contents: Vec<u8>,
 }
 
-fn is_system_epoch_info_event_type(type_: &StructTag, name: &str) -> bool {
-    type_.address() == Address::SYSTEM
-        && type_.module().as_str() == "iota_system_state_inner"
-        && type_.name().as_str() == name
-}
-
 impl Event {
+    fn is_system_epoch_info_event_type(&self, name: &str) -> bool {
+        self.type_.address() == Address::SYSTEM
+            && self.type_.module().as_str() == "iota_system_state_inner"
+            && self.type_.name().as_str() == name
+    }
+
     pub fn is_system_epoch_info_event(&self) -> bool {
         self.is_system_epoch_info_event_v1() || self.is_system_epoch_info_event_v2()
     }
 
     pub fn is_system_epoch_info_event_v1(&self) -> bool {
-        is_system_epoch_info_event_type(&self.type_, "SystemEpochInfoEventV1")
+        self.is_system_epoch_info_event_type("SystemEpochInfoEventV1")
     }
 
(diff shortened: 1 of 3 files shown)


thibault-martinez (approved):

crates/iota-sdk-types/src/events.rs
```diff
@@ -13,7 +13,7 @@ use super::{Address, Identifier, ObjectId, StructTag, TypeTag};
 /// ```text
 /// transaction-events = vector event
 /// ```
-#[derive(Eq, PartialEq, Clone, Debug)]
+#[derive(Eq, PartialEq, Clone, Debug, Hash)]
```
Alex6323, line 16: do we also want to expose it in the FFI?

Thoralf-M: In the FFI we just have a Record

crates/iota-sdk-types/src/events.rs
```diff
@@ -76,6 +76,21 @@ impl Event {
     pub fn is_system_epoch_info_event_v2(&self) -> bool {
         is_system_epoch_info_event_type(&self.type_, "SystemEpochInfoEventV2")
     }
+
+    pub fn random_for_testing() -> Self {
```
DaughterOfMars: Could we instead feature or target gate this and just call it `random`?

DaughterOfMars: Or just make a function in `iota-types` or something, because that's where it will be used

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/555/commits/c56f72d5d989dca06a2cfc6e8ff3bc71dee67d6e wouldn't put it in iota-types as we want to basically get rid of it and this might be interesting for other users as well

DaughterOfMars: I don't know how interesting it is and also it can be easily implemented for those who want it

Thoralf-M: It's used in 3 different crates in the monorepo, so there seems to be some demand for it, would you rather have each crate implement an own random fn for their tests?

DaughterOfMars: Possibly, yes. It's not a complicated method really, and it's for testing only.

DaughterOfMars: The thing is each crate may have its own needs for these randomized events. One random event for one crate may not be what the other wants. Trying to make this generic to all use-cases is kind of pointless

crates/iota-sdk-types/src/events.rs
```diff
@@ -58,6 +58,101 @@ pub struct Event {
     pub contents: Vec<u8>,
 }
 
+fn is_system_epoch_info_event_type(type_: &StructTag, name: &str) -> bool {
+    type_.address() == Address::SYSTEM
+        && type_.module().as_str() == "iota_system_state_inner"
+        && type_.name().as_str() == name
+}
```
Alex6323: Can we just make it a private method on `Event` instead of a free function?

crates/iota-sdk-types/src/gas.rs
```diff
@@ -101,33 +101,61 @@ impl GasCostSummary {
...
+            .unwrap()
     }
 
     /// The net gas usage, which is the total gas used minus the storage rebate.
     /// A positive number means used gas; negative number means refund.
     pub fn net_gas_usage(&self) -> i64 {
-        self.gas_used() as i64 - self.storage_rebate as i64
+        (self.gas_used() as i64)
+            .checked_sub(self.storage_rebate as i64)
+            .unwrap()
```
Alex6323: not sure what this change is exactly for tbh

crates/iota-sdk-types/src/events.rs
```diff
@@ -58,6 +58,101 @@ pub struct Event {
...
+
+#[cfg(test)]
+#[cfg(all(feature = "rand", not(target_arch = "wasm32")))]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_event_random() {
+        for _ in 0..100 {
```
Alex6323: isn't it a bit pointless to check 100 times that it created "something". I think just checking it once would suffice, because why wouldn't it create something in the othe 99 cases?

crates/iota-sdk-types/src/events.rs
Thoralf-M: Some general random Event might still be useful, but I agree that one often might want to have specific event data, therefore removed it now

crates/iota-sdk-types/src/events.rs
Thoralf-M: Sure https://github.com/iotaledger/iota-rust-sdk/pull/555/commits/7d4b70cc4a3c21cd0f9c781a38e7b5981f9b744e

crates/iota-sdk-types/src/events.rs
Thoralf-M: I did this to make sure that the random values are actually valid, running something random only one can easily create valid data, but multiple times could create something invalid (this actually kinda happened, there was an assert initially that a vec is empty and it worked the first time, but later failed as the vec was filled randomly with nothing or some bytes)

But I removed the random method completely now anyways

crates/iota-sdk-types/src/events.rs
```diff
@@ -58,6 +58,26 @@ pub struct Event {
...
+            && self.type_.module().as_str() == "iota_system_state_inner"
+            && self.type_.name().as_str() == name
+    }
+
+    pub fn is_system_epoch_info_event(&self) -> bool {
+        self.is_system_epoch_info_event_v1() || self.is_system_epoch_info_event_v2()
+    }
+
+    pub fn is_system_epoch_info_event_v1(&self) -> bool {
+        self.is_system_epoch_info_event_type("SystemEpochInfoEventV1")
```
DaughterOfMars: Can we make these consts?
```
const SYSTEM_EPOCH_INFO_EVENT_V1: &IdentifierRef = IdentifierRef::const_new("SystemEpochInfoEventV1");
const SYSTEM_EPOCH_INFO_EVENT_V2: &IdentifierRef = IdentifierRef::const_new("SystemEpochInfoEventV2");
```

Thoralf-M: Yes https://github.com/iotaledger/iota-rust-sdk/pull/555/commits/ea282ff907a42577edfd8aecf7ccdb370f148eff

Alex6323 (approved):

crates/iota-sdk-types/src/gas.rs
```diff
@@ -101,33 +101,61 @@ impl GasCostSummary {
...
     fn add_assign(&mut self, other: &Self) {
-        self.computation_cost += other.computation_cost;
-        self.computation_cost_burned += other.computation_cost_burned;
-        self.storage_cost += other.storage_cost;
-        self.storage_rebate += other.storage_rebate;
-        self.non_refundable_storage_fee += other.non_refundable_storage_fee;
+        self.computation_cost = self
+            .computation_cost
+            .checked_add(other.computation_cost)
+            .unwrap();
```
Alex6323, line 123: note however, that `derive_more::AddAssign` etc does not do Safe Math, so we now have different behavior for `Self` and `&Self` RHS.

DaughterOfMars: mmmm fair point

Thoralf-M: Good point, changed https://github.com/iotaledger/iota-rust-sdk/pull/555/commits/9d8dcd93251f1fd627961e706978ae1bd688a21b

Alex6323 (approved):

DaughterOfMars (approved):

crates/iota-sdk-types/src/events.rs
```diff
@@ -57,3 +57,29 @@ pub struct Event {
     #[cfg_attr(feature = "schemars", schemars(with = "crate::_schemars::Base64"))]
     pub contents: Vec<u8>,
 }
+
+impl Event {
+    const IOTA_SYSTEM_INNER: Identifier = Identifier::from_static("iota_system_state_inner");
```
thibault-martinez: Rename to IOTA_SYSTEM_STATE_INNER

Thoralf-M: Just removed it as we have it on the Identifier already https://github.com/iotaledger/iota-rust-sdk/pull/555/commits/62a0c8f22d9f0e409ca6af1501b2817b784e619a

PR #558, opened 2026-02-04
feat: ExecutionStatus changes for monorepo

thibault-martinez:


crates/iota-sdk-types/src/execution_status.rs
```diff
@@ -354,6 +449,55 @@ pub struct MoveLocation {
...
+// pub struct CongestedObjects(pub Vec<ObjectID>);
+
+// impl fmt::Display for CongestedObjects {
+//     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
+//         for obj in &self.0 {
+//             write!(f, "{obj}, ")?;
+//         }
+//         Ok(())
+//     }
+// }
```
Thoralf-M: Came to a conclusion after updating the iota repo?

thibault-martinez: I turned them into functions

crates/iota-sdk-types/src/execution_status.rs
```diff
@@ -39,6 +41,28 @@ pub enum ExecutionStatus {
...
+            ExecutionStatus::Failure { .. } => {
+                panic!("Unable to unwrap() on {self:?}");
+            }
+        }
+    }
+
+    pub fn unwrap_err(self) -> (ExecutionError, Option<u64>) {
+        match self {
+            ExecutionStatus::Success => {
+                panic!("Unable to unwrap() on {self:?}");
```
Thoralf-M:
```suggestion
                panic!("Unable to unwrap_err() on {self:?}");
```

Thoralf-M (approved):

crates/iota-sdk-types/src/digest.rs
```diff
@@ -374,6 +398,33 @@ mod tests {
         );
     }
 
+    #[test]
+    fn partial_eq_array_u8() {
+        let digest = Digest::new([1u8; 32]);
+        let matching: [u8; 32] = [1u8; 32];
+        let non_matching: [u8; 32] = [2u8; 32];
```
Alex6323: nit: the type can be inferred I guess; same in the `Vec` test

crates/iota-sdk-types/src/execution_status.rs
```diff
@@ -39,6 +41,28 @@ pub enum ExecutionStatus {
...
+        }
+    }
+
+    pub fn unwrap_err(self) -> (ExecutionError, Option<u64>) {
+        match self {
+            ExecutionStatus::Success => {
+                panic!("Unable to unwrap_err() on {self:?}");
+            }
+            ExecutionStatus::Failure { error, command } => (error, command),
+        }
```
Alex6323: I would prefer to use `Self` if possible instead of its long name.

crates/iota-sdk-types/src/execution_status.rs
```diff
@@ -152,130 +211,199 @@ impl ExecutionStatus {
 #[non_exhaustive]
 pub enum ExecutionError {
     /// Insufficient Gas
+    #[error("Insufficient Gas.")]
     InsufficientGas,
     /// Invalid Gas Object.
+    #[error("Invalid Gas Object. Possibly not address-owned or possibly not an IOTA coin.")]
     InvalidGasObject,
     /// Invariant Violation
+    #[error("INVARIANT VIOLATION.")]
```
Alex6323: sure about the screaming uppercase? And this is the most cryptic errors of all, but here we can't provide some hints as to what might have gone wrong?

crates/iota-sdk-types/src/execution_status.rs
```diff
@@ -152,130 +211,199 @@ impl ExecutionStatus {
...
     InvariantViolation,
     /// Attempted to used feature that is not supported yet
+    #[error("Attempted to used feature that is not supported yet")]
```
Alex6323: typo: `use` instead of `used`, also in the comment

Alex6323: also because I think you care (haha), but punctuation is a bit inconsistent.

crates/iota-sdk-types/src/digest.rs
thibault-martinez: done

crates/iota-sdk-types/src/execution_status.rs
thibault-martinez: done

crates/iota-sdk-types/src/execution_status.rs
thibault-martinez: I wouldn't mess with this, it's propagated through the infra. If protocol/infra wants to change it, they will.

crates/iota-sdk-types/src/execution_status.rs
thibault-martinez: done

Alex6323 (approved):

PR #563, opened 2026-02-09
chore!: remove `BalanceChange` type

Alex6323:
If I am not mistaken, this type is completely useless in the SDK, because:

- it's not exported in the FFI
- is incompatible with the graphql schema spec (which allows different owner variants)
- only compatible with the REST API version, but the SDK doesn't provide a client for that.

Therefore this PR removes it.

### DIFF
--- a/crates/iota-sdk-types/src/lib.rs
+++ b/crates/iota-sdk-types/src/lib.rs
@@ -486,32 +486,6 @@ mod _schemars {
         }
     }
 
-    #[expect(dead_code)]
-    pub(crate) struct I128;
-
-    impl JsonSchema for I128 {
-        fn schema_name() -> String {
-            "i128".to_owned()
-        }
-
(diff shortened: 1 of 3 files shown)


crates/iota-sdk-types/src/lib.rs
```diff
@@ -486,6 +486,7 @@ mod _schemars {
         }
     }
 
+    #[expect(dead_code)]
     pub(crate) struct I128;
```
Thoralf-M: Any point in keeping it if it's not used?

Alex6323: no, we can remove it now and re-introduce it in case we actually need it at some point again.

Thoralf-M (approved):

DaughterOfMars (approved):

thibault-martinez (approved):

Issue #515, opened 2026-01-09
[Bounty] Add move package example


Thoralf-M:
Please read the [README](https://github.com/orgs/iotaledger/projects/97?pane=info) first.

### Summary

Given a move package id, fetch data that is interesting for it and print it, for example:
- package versions, the version of the provided package and other existing versions
- modules and their functions
- dependencies
- types defined in this package and example objects of this type if available in the network

### Estimated complexity

Easy

### Estimated duration

Few hours

### Details & acceptance criteria

- [ ] Add example to rust
- [ ] Add example to all bindings

### Pull request required?

Yes

David-code-tang:
Claiming this bounty. I will add a new move package example in Rust and corresponding examples across bindings (Go/Python/Kotlin), focused on package metadata exploration (versions/modules/dependencies/types and sample objects).

David-code-tang:
Done: opened the implementation PR for this bounty.\n\n- PR: https://github.com/iotaledger/iota-rust-sdk/pull/570\n\nIt adds the move package example in Rust and all bindings (Python/Go/Kotlin), covering package versions, latest package, dependencies, modules/functions, and typed object lookup for package structs.

bitdeckai:
Hi, I would like to claim this bounty task.

Plan:
- add Move package example in Rust
- add equivalent example to existing bindings where required by the issue
- include short verification notes

ETA: 1 day.

MrLawrenceKwan:
I’d like to work on this bounty under the project rules. Could you assign this issue to me (MrLawrenceKwan) before I submit PR?\n\nI have implementation ready and can re-submit immediately once assigned.

EasonC13:
Hi, we're from GiveRep Labs. We have extensive Sui Move development experience which translates well to IOTA Move packages. We'd like to take on this bounty and implement the move package example in Rust and all bindings. Ready to start once assigned.

thibault-martinez:
> Hi, we're from GiveRep Labs. We have extensive Sui Move development experience which translates well to IOTA Move packages. We'd like to take on this bounty and implement the move package example in Rust and all bindings. Ready to start once assigned.

If no one else expresses interest, I'll assign you, for now you already have a task to work on 👍🏻

abhiraj-mengade:
I would like to take this up @thibault-martinez

chloejw-18:
[...]

PR #570, opened 2026-02-12
[Bounty] Add move package example across SDK bindings

David-code-tang:
## Summary
- add a new `move_package` example in Rust
- add matching `move_package` examples in Python, Go, and Kotlin bindings
- example shows package versions, latest version details, dependencies, modules/functions, and typed object lookup for package structs

## Notes
- This is examples-only work for issue #515.
- Local toolchain is not available in this environment (no cargo/go), so I could only run Python syntax check.

### DIFF
--- a/bindings/go/examples/move_package/main.go
+++ b/bindings/go/examples/move_package/main.go
@@ -1,0 +1,102 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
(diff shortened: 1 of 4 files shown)


thibault-martinez:
Thanks @David-code-tang for your interest in this but the Bounty Program will start on the 16th of February at first only for select community members. It will then be open for anyone with remaining tasks on the 2nd of March. Please feel free to reopen your PR then or make it a non-bounty PR.

Issue #501, opened 2026-01-06
[Bounty] Add an SDK Example for multisig address


Thoralf-M:
Please read the [README](https://github.com/orgs/iotaledger/projects/97?pane=info) first.

### Summary

Add an example that create a multisig address with multiple keys, then gets funds to it and signs a transaction using multiple keys from this address. Similar to https://github.com/Thoralf-M/iota-examples/blob/main/wallet/transactions/multisig.ts

### Estimated complexity

Easy

### Estimated duration

1 hour

### Details & acceptance criteria

- [ ] Add example to rust
- [ ] Add example to all bindings

### Pull request required?

Yes

David-code-tang:
Claiming this bounty. I will add a multisig address example in Rust and matching examples for all bindings (Python/Go/Kotlin), including funding and multi-key signing flow.

David-code-tang:
Done: opened implementation PR for this bounty.\n\n- PR: https://github.com/iotaledger/iota-rust-sdk/pull/571\n\nIt adds a multisig address example in Rust and all bindings (Python/Go/Kotlin), including committee creation, multisig address derivation, faucet funding, multi-signature aggregation, and transaction execution.

shaibuafeez:
Interested in building this.

thibault-martinez:
> Interested in building this.

Assigned you! Thanks :)

shaibuafeez:
Always happy to help 🫡

justinleeyang:
I can take bounty #501. I will share a scoped implementation plan and first PR shortly.

justinleeyang:
Quick follow-up on #501: I am still actively working on this. Please let me know if you want me to narrow scope to a smaller first PR.

justinleeyang:
No maintainer response yet on #501, I will switch to another open bounty for now and can resume this issue anytime when feedback arrives.

PR #571, opened 2026-02-12
[Bounty] Add multisig address example across SDK bindings

David-code-tang:
## Summary
- add a multisig address example in Rust
- add matching examples in Python, Go, and Kotlin bindings
- flow includes: create 2-of-2 multisig committee, derive multisig address, fund via faucet, build tx, collect two signatures, aggregate multisig signature, and execute tx

## Notes
- This addresses issue #501.
- Local environment does not have cargo/go toolchains, so compile checks were not run here.

### DIFF
--- a/bindings/go/examples/multisig_address/main.go
+++ b/bindings/go/examples/multisig_address/main.go
@@ -1,0 +1,103 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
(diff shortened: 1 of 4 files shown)


thibault-martinez:
Thanks @David-code-tang for your interest in this but the Bounty Program will start on the 16th of February at first only for select community members. It will then be open for anyone with remaining tasks on the 2nd of March. Please feel free to reopen your PR then or make it a non-bounty PR.

Issue #514, opened 2026-01-08
[Bounty] Add IOTA Names example


thibault-martinez:
Please read the [README](https://github.com/orgs/iotaledger/projects/97?pane=info) first.

### Summary

Add an example showcasing all major IOTA Names operations.

### Estimated complexity

Easy

### Estimated duration

2 hours

### Details & acceptance criteria

- [ ] Add example to rust
- [ ] Add example to all bindings
- [ ] (Optional: change dev_inspect example so it doesn't use IOTA names)

### Pull request required?

Yes

Drscq:
Claiming this bounty. I’ll add a dedicated IOTA Names example across Rust, Python, Go, and Kotlin, covering lookup, registrations, and default name retrieval.

Drscq:
Submitted PR: https://github.com/iotaledger/iota-rust-sdk/pull/565\n\nThis adds the new  example in Rust and matching examples in Python, Go, and Kotlin, covering lookup, default-name retrieval, and registrations with pagination.

Drscq:
Follow-up: the new example file is crates/iota-sdk/examples/iota_names.rs.\n\nPR link: https://github.com/iotaledger/iota-rust-sdk/pull/565

Drscq:
Active submission PR: https://github.com/iotaledger/iota-rust-sdk/pull/566\n\n(Previous PR #565 has been closed as superseded.)

MrLawrenceKwan:
I’d like to work on this bounty under the project rules. Could you assign this issue to me (MrLawrenceKwan) before I submit PR?\n\nI have implementation ready and can re-submit immediately once assigned.

EasonC13:
Hi, we're from GiveRep Labs. Our product [iota.gives](https://iota.gives/) (built with TypeScript) already integrates IOTA Names, so we're very familiar with the operations involved. We also have development experience across other languages and would be happy to contribute examples in Rust and all bindings. Can also refactor the dev_inspect example if needed. Ready to start once assigned.

thibault-martinez:
> Hi, we're from GiveRep Labs. Our product [iota.gives](https://iota.gives/) (built with TypeScript) already integrates IOTA Names, so we're very familiar with the operations involved. We also have development experience across other languages and would be happy to contribute examples in Rust and all bindings. Can also refactor the dev_inspect example if needed. Ready to start once assigned.

Thank you for your interest, you are now assigned :)

justinleeyang:
[...]

PR #576, opened 2026-02-14
feat(examples): add IOTA Names examples for Go and Python (#514)

hechibing:
## Summary
- add a new Go example at `bindings/go/examples/iota_names/main.go`
- add a new Python example at `bindings/python/examples/iota_names.py`
- demonstrate the three GraphQL IOTA Names APIs:
  - `iota_names_lookup`
  - `iota_names_default_name`
  - `iota_names_registrations`
- include optional-value handling and printed output for resolved names and registrations

## Validation
- `python -m py_compile bindings/python/examples/iota_names.py`
- `go build -buildvcs=false ./examples/iota_names` (with cgo linker flags to `target/release`)

### DIFF
--- a/bindings/go/examples/iota_names/main.go
+++ b/bindings/go/examples/iota_names/main.go
@@ -1,0 +1,64 @@
+// Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"fmt"
+	"log"
+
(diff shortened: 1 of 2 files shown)


thibault-martinez:
Thanks for your interest but as per its [README](https://github.com/orgs/iotaledger/projects/97?pane=info), the Bounty is currently not open to the general public.

Issue #512, opened 2026-01-08
[Bounty] Improve `iota-sdk-graphql-client` structure and uniformity


DaughterOfMars:
Please read the [README](https://github.com/orgs/iotaledger/projects/97?pane=info) first.

### Summary

Some areas of the code in this crate can be organized better.

### Estimated complexity

Easy

### Estimated duration

1 day

### Details & acceptance criteria

- [ ] Split up big files such as `lib.rs`
- [ ] Refactor query structures so that they are consistent and descriptive
- [ ] Move code as needed to better organize the modules

### Pull request required?

Yes

hk089660:
Hi! I’d like to work on this bounty (#512) and open a PR.

Plan: split up large files (e.g. lib.rs) and reorganize/refactor the GraphQL query modules so naming/structure is consistent and descriptive.

Quick question before I start — what is the payout amount and payout method for this bounty once the PR is merged?

I prefer async text-based communication; I’ll post progress updates here. Also, I’ll make sure my commits are signed as requested. Thanks!

bitdeckai:
Hi maintainers, I'd like to take this bounty (#512).

Plan:
1) split oversized GraphQL client modules into clearer, feature-scoped files;
2) standardize naming conventions and query/response structure across modules;
3) keep behavior unchanged, with focused tests/docs updates where structure changed.

I can start immediately and share a first PR in small reviewable commits.

thibault-martinez:
> Hi maintainers, I'd like to take this bounty ([#512](https://github.com/iotaledger/iota-rust-sdk/issues/512)).
> 
> Plan:
> 
> 1. split oversized GraphQL client modules into clearer, feature-scoped files;
> 2. standardize naming conventions and query/response structure across modules;
> 3. keep behavior unchanged, with focused tests/docs updates where structure changed.
> 
> I can start immediately and share a first PR in small reviewable commits.

Hi @bitdeckai, thanks for your interest but as per the [README](https://github.com/orgs/iotaledger/projects/97?pane=info), the bounty program is currently only available to select community members.

bitdeckai:
Update: I opened a first PR for this bounty scope: https://github.com/iotaledger/iota-rust-sdk/pull/971

It focuses on structural uniformity in iota-sdk-graphql-client by splitting shared query types and scalars into dedicated modules while keeping public API re-exports consistent.

SeventhOdyssey71:
Interested in taking this task. @thibault-martinez

[...]

PR #578, opened 2026-02-14
refactor(graphql-client): reorganize faucet and output types

hechibing:
Refs #512.


thibault-martinez:
Thanks for your interest but as per its [README](https://github.com/orgs/iotaledger/projects/97?pane=info), the Bounty is currently not open to the general public.

PR #580, opened 2026-02-16
feat: update `TransactionEffects`

Alex6323:
## Summary

Continues the SDK replacement series. Brings `TransactionEffects` in line with the monorepo replacement: drops the inherent accessors so callers match on the enum (mirroring how `Transaction` already works), adds `Default` impls, moves `IdOperation` up to the `effects` module, introduces `InputSharedObject` and `ObjectChange`, and pins public schema names on internal serde helpers across the crate.

- **Drop `status()` / `epoch()` / `gas_summary()` accessors** on `TransactionEffects` (and the matching ones on `TransactionEffectsV1`) in [crates/iota-sdk-types/src/effects/mod.rs](crates/iota-sdk-types/src/effects/mod.rs) and [crates/iota-sdk-types/src/effects/v1.rs](crates/iota-sdk-types/src/effects/v1.rs). Callers now match on the variant or go through `as_v1()`.
- **`Default`** impls for `TransactionEffects` and `TransactionEffectsV1` in [crates/iota-sdk-types/src/effects/mod.rs](crates/iota-sdk-types/src/effects/mod.rs) and [crates/iota-sdk-types/src/effects/v1.rs](crates/iota-sdk-types/src/effects/v1.rs).
- **Move `IdOperation`** from `effects::v1` up to `effects` in [crates/iota-sdk-types/src/effects/mod.rs](crates/iota-sdk-types/src/effects/mod.rs) so it sits next to the other shared effect types; re-exports adjusted in [crates/iota-sdk-types/src/lib.rs](crates/iota-sdk-types/src/lib.rs).
- **`InputSharedObject`** new enum (`Mutate`, `ReadOnly`, `ReadDeleted`, `MutateDeleted`, `Cancelled`) with `object_ref()` and `id_and_version()` helpers, in [crates/iota-sdk-types/src/effects/mod.rs](crates/iota-sdk-types/src/effects/mod.rs).
- **`ObjectChange`** new struct keyed by `ObjectId` with input/output `Version`/`Digest` and an `IdOperation`, in [crates/iota-sdk-types/src/effects/mod.rs](crates/iota-sdk-types/src/effects/mod.rs).
- **`ChangedObject` schema rename** to `EffectsObjectChange` with `rename_all = "camelCase"` to match the monorepo wire format, in [crates/iota-sdk-types/src/effects/v1.rs](crates/iota-sdk-types/src/effects/v1.rs).
- **Pin public schema names** by adding `#[serde(rename = "PublicName")]` to internal `Readable*` / `Binary*` serde helper structs and enums, so generated schemas keep the public type name. Affected types: `TransactionEffects`, `TransactionEffectsV1`, `UnchangedSharedKind`, `ObjectIn`, `ObjectOut` ([effects/mod.rs](crates/iota-sdk-types/src/effects/mod.rs), [effects/v1.rs](crates/iota-sdk-types/src/effects/v1.rs)); `CheckpointSummary`, `CheckpointContents`, `CheckpointCommitment` ([checkpoint.rs](crates/iota-sdk-types/src/checkpoint.rs)); `MultisigAggregatedSignature` ([crypto/multisig.rs](crates/iota-sdk-types/src/crypto/multisig.rs)); `UserSignature` ([crypto/signature.rs](crates/iota-sdk-types/src/crypto/signature.rs)); `ExecutionError`, `CommandArgumentError`, `PackageUpgradeError` ([execution_status.rs](crates/iota-sdk-types/src/execution_status.rs)); `Package`, `MoveStruct`, `Object` ([object.rs](crates/iota-sdk-types/src/object.rs)); `EndOfEpochTransactionKind`, `Input`, `Command`, `SignedTransaction` ([transaction/serialization.rs](crates/iota-sdk-types/src/transaction/serialization.rs)).
- **Callers updated** to match on `TransactionEffects::V1` (or go through `as_v1()`) instead of using the removed accessors, in [iota-sdk-transaction-builder](crates/iota-sdk-transaction-builder/src/lib.rs), [iota-sdk-transaction-builder/client_methods](crates/iota-sdk-transaction-builder/src/builder/client_methods.rs), and the `abstract_account`, `multisig`, `polling-indexer`, `publish_upgrade`, `sign_send_iota`, and `transaction_signer_callback` examples under [crates/iota-sdk/examples](crates/iota-sdk/examples).

### DIFF
--- a/crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
+++ b/crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
@@ -225,7 +225,7 @@ impl ClientMethods for iota_graphql_client::Client {
     async fn estimate_tx_budget(&self, tx: &Transaction) -> Result<Option<u64>, Self::Error> {
         let res = self.dry_run_tx(tx, true).await?;
         Ok(res.effects.map(|effects| match effects {
-            TransactionEffects::V1(v1) => v1.gas_used.gas_used(),
+            TransactionEffects::V1(v1) => v1.gas_cost_summary.gas_used(),
             _ => unimplemented!(
                 "a new TransactionEffects enum variant was added and needs to be handled"
             ),
--- a/crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
+++ b/crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
@@ -224,7 +224,12 @@ impl ClientMethods for iota_graphql_client::Client {
 
     async fn estimate_tx_budget(&self, tx: &Transaction) -> Result<Option<u64>, Self::Error> {
         let res = self.dry_run_tx(tx, true).await?;
-        Ok(res.effects.map(|e| e.gas_summary().gas_used()))
+        Ok(res.effects.map(|effects| match effects {
+            TransactionEffects::V1(v1) => v1.gas_used.gas_used(),
+            _ => unimplemented!(
+                "a new TransactionEffects enum variant was added and needs to be handled"
+            ),
+        }))
     }
 
     async fn dry_run_tx(
--- a/crates/iota-sdk-transaction-builder/src/lib.rs
+++ b/crates/iota-sdk-transaction-builder/src/lib.rs
@@ -384,14 +384,19 @@ mod tests {
         }
     }
 
-    /// Wait for the transaction to be finalized and indexed, and check the
-    /// effects' to ensure the transaction was successfully executed.
-    async fn check_effects_status_success(effects: Result<TransactionEffects, Error>) {
+    /// Check the effects to ensure the transaction was successfully executed.
+    fn check_effects_status_success(effects: Result<TransactionEffects, Error>) {
         assert!(effects.is_ok(), "Execution failed. Effects: {effects:?}");
+
         // check that it succeeded
-        let status = effects.unwrap();
-        let expected_status = ExecutionStatus::Success;
-        assert_eq!(&expected_status, status.status());
+        match effects.unwrap() {
+            TransactionEffects::V1(v1) => {
+                assert_eq!(ExecutionStatus::Success, v1.status);
+            }
+            _ => unimplemented!(
+                "a new TransactionEffects enum variant was added and needs to be handled"
+            ),
+        };
     }
 
     #[tokio::test]
@@ -508,7 +513,7 @@ mod tests {
         tx.transfer_objects(recipient, [coin]);
 
         let effects = tx.execute(&pk, WaitForTx::Finalized).await;
-        check_effects_status_success(effects).await;
+        check_effects_status_success(effects);
 
         // check that recipient has 1 coin
         let recipient_coins = client
@@ -529,7 +534,7 @@ mod tests {
             .arguments([Some(1u64)]);
 
         let effects = tx.execute(&pk, WaitForTx::Finalized).await;
-        check_effects_status_success(effects).await;
+        check_effects_status_success(effects);
     }
 
     #[tokio::test]
@@ -544,7 +549,7 @@ mod tests {
         tx.transfer_objects(recipient, [assigned("coin")]);
 
         let effects = tx.execute(&pk, WaitForTx::Finalized).await;
-        check_effects_status_success(effects).await;
+        check_effects_status_success(effects);
 
         // check that recipient has 1 coin
         let recipient_coins = client
@@ -563,11 +568,15 @@ mod tests {
         // transfer 1 IOTA
         tx.split_coins(coin, [1_000_000_000u64]);
 
-        let effects = tx.execute(&pk, WaitForTx::Finalized).await.unwrap();
-
-        let expected_status = ExecutionStatus::Success;
-        // The tx failed, so we expect Failure instead of Success
-        assert_ne!(&expected_status, effects.status());
+        match tx.execute(&pk, WaitForTx::Finalized).await.unwrap() {
+            TransactionEffects::V1(v1) => {
+                // The tx failed, so we expect Failure instead of Success
+                assert_ne!(ExecutionStatus::Success, v1.status);
+            }
+            _ => unimplemented!(
+                "a new TransactionEffects enum variant was added and needs to be handled"
+            ),
+        };
     }
 
     #[tokio::test]
@@ -586,7 +595,7 @@ mod tests {
         let client = tx.get_client().clone();
 
         let effects = tx.execute(&pk, WaitForTx::Finalized).await;
-        check_effects_status_success(effects).await;
+        check_effects_status_success(effects);
 
         // check that there are two coins
         let coins_after = client
@@ -603,7 +612,7 @@ mod tests {
         tx.make_move_vec([1u64]);
 
         let effects = tx.execute(&pk, WaitForTx::Finalized).await;
-        check_effects_status_success(effects).await;
+        check_effects_status_success(effects);
     }
 
     #[tokio::test]
@@ -616,7 +625,7 @@ mod tests {
             .transfer_objects(address, [assigned("cap")]);
 
         let effects = tx.execute(&pk, WaitForTx::Finalized).await;
-        check_effects_status_success(effects).await;
+        check_effects_status_success(effects);
     }
 
     #[tokio::test]
@@ -652,7 +661,7 @@ mod tests {
                 _ => unimplemented!("a new enum variant was added and needs to be handled"),
             }
         }
-        check_effects_status_success(effects).await;
+        check_effects_status_success(effects);
 
         let client = Client::new_localnet();
         let mut tx = TransactionBuilder::new(address).with_client(&client);
@@ -690,6 +699,6 @@ mod tests {
         tx.gas([coins.last().unwrap().id]);
 
         let effects = tx.execute(&pk, WaitForTx::Finalized).await;
-        check_effects_status_success(effects).await;
+        check_effects_status_success(effects);
     }
 }
--- a/crates/iota-sdk-types/src/effects/mod.rs
+++ b/crates/iota-sdk-types/src/effects/mod.rs
@@ -9,8 +9,6 @@ pub use v1::{
     UnchangedSharedObject,
 };
 
-use crate::{Digest, ObjectId, ObjectReference, Version};
-
 /// The output or effects of executing a transaction
 ///
 /// # BCS
@@ -141,40 +139,6 @@ mod serialization {
     }
 }
 
-/// Description of a shared object that was used as input to a transaction
-///
-/// Captures how each shared object was accessed during execution: whether it
-/// was mutated, read-only, deleted after mutable or read-only access, or
-/// cancelled.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum InputSharedObject {
-    Mutate(ObjectReference),
-    ReadOnly(ObjectReference),
-    ReadDeleted(ObjectId, Version),
-    MutateDeleted(ObjectId, Version),
-    Cancelled(ObjectId, Version),
-}
-
-impl InputSharedObject {
-    pub fn id_and_version(&self) -> (ObjectId, Version) {
-        let (object_id, version, ..) = self.object_ref().into_parts();
-        (object_id, version)
-    }
-
-    pub fn object_ref(&self) -> ObjectReference {
-        match self {
-            InputSharedObject::Mutate(oref) | InputSharedObject::ReadOnly(oref) => *oref,
-            InputSharedObject::ReadDeleted(id, version)
-            | InputSharedObject::MutateDeleted(id, version) => {
-                ObjectReference::new(*id, *version, Digest::OBJECT_DELETED)
-            }
-            InputSharedObject::Cancelled(id, version) => {
-                ObjectReference::new(*id, *version, Digest::OBJECT_CANCELLED)
-            }
-        }
-    }
-}
-
 /// Defines what happened to an ObjectId during execution
 ///
 /// # BCS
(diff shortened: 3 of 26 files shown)


crates/iota-sdk-types/src/effects/mod.rs
```diff
@@ -162,3 +152,132 @@ mod serialization {
         }
     }
 }
+
+#[derive(Eq, PartialEq, Clone, Debug)]
```
Alex6323: reminder: reorder derives

crates/iota-sdk-types/src/effects/mod.rs
```diff
@@ -162,3 +152,132 @@ mod serialization {
...
+    None,
+    Created,
+    Deleted,
+}
+
+impl IdOperation {
+    crate::def_is!(None, Created, Deleted);
+}
+
+#[derive(Clone)]
```
Thoralf-M: Derive more things and add a comment

Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/67a623f7886a1e7f0fe3ff05ce2d146454c99f51

crates/iota-sdk-types/src/effects/mod.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/67a623f7886a1e7f0fe3ff05ce2d146454c99f51

crates/iota-sdk-types/src/effects/mod.rs
```diff
@@ -162,3 +152,150 @@ mod serialization {
...
+    /// Return the status of the transaction.
+    fn status(&self) -> &ExecutionStatus;
+    fn into_status(self) -> ExecutionStatus;
+    /// Return the epoch in which this transaction was executed.
+    fn epoch(&self) -> EpochId;
+    fn modified_at_versions(&self) -> Vec<(ObjectId, Version)>;
+    /// The version assigned to all output objects (apart from packages).
+    fn lamport_version(&self) -> Version;
+    fn old_object_metadata(&self) -> Vec<(ObjectReference, Owner)>;
+    fn input_shared_objects(&self) -> Vec<InputSharedObject>;
```
Thoralf-M: Add the comments from the iota repo for the remaining methods?
There are some missing for old_object_metadata, input_shared_objects, unchanged_shared_objects...

Alex6323: 5661e11d6ebf00d7ad510ac5557bc8a16a4e9f7e

Thoralf-M (approved):

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -224,7 +224,13 @@ impl ClientMethods for iota_graphql_client::Client {
 
     async fn estimate_tx_budget(&self, tx: &Transaction) -> Result<Option<u64>, Self::Error> {
         let res = self.dry_run_tx(tx, true).await?;
-        Ok(res.effects.map(|e| e.gas_cost_summary().gas_used()))
+        // Ok(res.effects.map(|effects| effects.gas_cost_summary().gas_used()))
```
DaughterOfMars: left the comment

Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/6cc65b98e25f5ce782ab8c8c2d0ba2af2d26e1b7

crates/iota-sdk-transaction-builder/src/lib.rs
```diff
@@ -387,9 +387,9 @@ mod tests {
     async fn check_effects_status_success(effects: Result<TransactionEffects, Error>) {
         assert!(effects.is_ok(), "Execution failed. Effects: {effects:?}");
         // check that it succeeded
-        let status = effects.unwrap();
+        let effects = effects.unwrap().into_v1();
```
DaughterOfMars: We should probably have a match for places like this so we'll know to update them if we add new versions

DaughterOfMars (approved):

crates/iota-sdk-transaction-builder/src/lib.rs
```diff
@@ -386,10 +386,16 @@ mod tests {
     /// effects' to ensure the transaction was successfully executed.
```
Thoralf-M: Was already before, but do you mind updating the documentation here? This fn is not waiting for the tx to be finalized and indexed (also there is no need to have it async since it doesn't do that)

Alex6323: changed

crates/iota-sdk/examples/sign_send_iota.rs
```diff
@@ -39,11 +39,16 @@ async fn main() -> Result<()> {
 
     let signature = private_key.sign_transaction(&tx)?;
 
-    let effects = client.execute_tx(&[signature], &tx, None).await?;
-
-    println!("Digest: {}", effects.digest());
-    println!("Transaction status: {:?}", effects.status());
-    println!("Effects: {effects:#?}");
+    match client.execute_tx(&[signature], &tx, None).await? {
```
thibault-martinez: Let's not match in examples (like most of them actually), use as_v1

crates/iota-sdk/examples/transaction_signer_callback.rs
```diff
@@ -41,11 +41,16 @@ async fn main() -> Result<()> {
     builder.send_iota(recipient_address, amount);
 
     let signer = AsyncSigner(private_key);
-    let effects = builder.execute(&signer, WaitForTx::Finalized).await?;
-
-    println!("Digest: {}", effects.digest());
-    println!("Transaction status: {:?}", effects.status());
-    println!("Effects: {effects:#?}");
+    match builder.execute(&signer, WaitForTx::Finalized).await? {
```
thibault-martinez: Let's not match in examples (like most of them actually), use as_v1

crates/iota-sdk/examples/abstract_account.rs
```diff
@@ -38,11 +38,17 @@ async fn main() -> Result<()> {
         .finish(&client)
         .await?;
 
-    let effects = builder
+    match builder
```
thibault-martinez: What's the point of this match?

crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -68,20 +67,21 @@ pub struct TransactionEffectsV1 {
...
-    }
-
-    /// The epoch when this transaction was executed.
-    pub fn epoch(&self) -> EpochId {
-        self.epoch
-    }
-
-    /// The gas used in this transaction.
-    pub fn gas_summary(&self) -> &GasCostSummary {
-        &self.gas_used
```
thibault-martinez, line 84: Commenting here because I can't on the struct. I know it was like that already but idk if I like `gas_used`, the type is called `GasCostSummary` and we keep calling it `gas_cost_summary` in the monorepo

crates/iota-sdk-types/src/effects/mod.rs
```diff
@@ -156,3 +146,82 @@ mod serialization {
...
+    Mutate(ObjectReference),
+    ReadOnly(ObjectReference),
+    ReadDeleted(ObjectId, Version),
+    MutateDeleted(ObjectId, Version),
+    Cancelled(ObjectId, Version),
+}
+
+impl InputSharedObject {
+    pub fn id_and_version(&self) -> (ObjectId, Version) {
+        let (object_id, version, ..) = self.object_ref().into_parts();
```
roman1e2f5p8s: for the last 3 variants, this constructs a digest via `object_ref()` only to immediately discard it. why not a direct match here?

roman1e2f5p8s (approved):

crates/iota-sdk/examples/abstract_account.rs
Alex6323: I am not sure I fully agree on your comments, because IMHO we should look at the examples from the perspective of what we suggest users to do in their consumer code (but I think we had a similar discussion where we disagreed). In my opinion these examples should show best practices in general, and specifically here we want users to take into consideration, that there will be future versions V2, V3, ...., so they should already prepare their code with a match statement giving them a proper signal as to what they have to do when a new version lands. `as_v1()` will just panic with a confusing error.

Alex6323: Of course the downside is that the examples are less concise because of the match.

crates/iota-sdk/examples/sign_send_iota.rs
Alex6323: Fine, but I think we should not encourage the usage of `as_v1` to consumers, and those examples are often expected to show best practice of how the API is supposed to be used. Those examples are most likely used as a starting point by copy&pasting. Or what we also could do is at least add a comment there to tell them to use a match instead in their code and show a useful panic message in case a new version arrives.

crates/iota-sdk/examples/abstract_account.rs
thibault-martinez: That would be true if the example was focusing on TransactionEffects, it is not, so it's completely irrelevant. The point of examples is that they are straight to the point showcasing a very specific feature, there is no reason to make it more complex than it should be.

crates/iota-sdk/examples/sign_send_iota.rs
thibault-martinez: We should not encourage the use of `unwrap` either, but we're still doing it. Because properly handling the error is irrelevant to the feature being showcased. Same thing.

crates/iota-sdk-types/src/effects/v1.rs
Alex6323: Right, then let's rename everything to `gas_cost_summary`, the field and the method. Okay?

crates/iota-sdk-types/src/effects/mod.rs
Alex6323: I see. You're right. I would still argue against it: We still want `object_ref()` to exist on this type, and this is purely to adhere to the DRY principle, because what you suggest is an additional match, which means we then have two places we need to update in case. Constructing that type is negligible, no? I doesn't look like there's heap allocation involved, so it should be very cheap to do.

crates/iota-sdk/examples/sign_send_iota.rs
Alex6323: `unwrap()` is a clear signal, `as_v1()` is not!

Alex6323: I also see your point though, so it's fine: https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/9c951d8ab066521dc8928458cd653edcaf8b896a

crates/iota-sdk/examples/transaction_signer_callback.rs
Alex6323: changed in https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/9c951d8ab066521dc8928458cd653edcaf8b896a

crates/iota-sdk/examples/abstract_account.rs
Alex6323: I don't think it's even possible to have a pure `TransactionEffects` example without showing how to sign and send a tx, but I don't think I can convince you, so I made the change: https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/9c951d8ab066521dc8928458cd653edcaf8b896a

crates/iota-sdk/examples/sign_send_iota.rs
thibault-martinez: `as_v1_opt.unwrap` if you really care that much

crates/iota-sdk-types/src/effects/v1.rs
Alex6323: I renamed. For the readable serialization I added a serde rename as to not break stuff, for the binary serialization that was irrelevant because of bcs: https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/9c951d8ab066521dc8928458cd653edcaf8b896a

crates/iota-sdk/examples/abstract_account.rs
thibault-martinez: The match is still here

crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -68,20 +67,21 @@ pub struct TransactionEffectsV1 {
...
+impl Default for TransactionEffectsV1 {
```
thibault-martinez: Let's remove this, this is error-prone and shouldn't really be needed

crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -95,7 +95,12 @@ impl TransactionEffectsV1 {
 /// changed-object = object-id object-in object-out id-operation
 /// ```
 #[derive(Eq, PartialEq, Clone, Debug)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename = "EffectsObjectChange"),
+    serde(rename_all = "camelCase")
```
thibault-martinez: We don't use camelCase

crates/iota-sdk/examples/abstract_account.rs
Alex6323: f***

crates/iota-sdk-types/src/effects/v1.rs
Alex6323: okay

crates/iota-sdk/examples/abstract_account.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/36dfa952c9e41d473a321b8acfd1a730e2066ca8

crates/iota-sdk-types/src/effects/v1.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/36dfa952c9e41d473a321b8acfd1a730e2066ca8

crates/iota-sdk-types/src/effects/v1.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/580/commits/36dfa952c9e41d473a321b8acfd1a730e2066ca8

crates/iota-sdk-types/src/effects/mod.rs
Alex6323: Unrelated to your comment, but this type has been moved back to the core repo.

thibault-martinez (approved):

PR #581, opened 2026-02-16
feat: `Owner` improvements

DaughterOfMars:
## Description

Improve the Owner enum for the core implementation.

### DIFF
--- a/crates/iota-sdk-types/src/object.rs
+++ b/crates/iota-sdk-types/src/object.rs
@@ -113,6 +113,16 @@ impl Owner {
     crate::def_is!(Immutable);
 
     crate::def_is_as_into_opt!(Address, Object(ObjectId), Shared(Version));
+
+    /// Returns an address if this object is owned by an address or
+    /// object, and None if it is shared or immutable.
+    pub fn address(&self) -> Option<&Address> {
+        Some(match self {
+            Self::Address(addr) => addr,
+            Self::Object(obj_id) => obj_id.as_address(),
+            _ => return None,
+        })
+    }
 }
 
 impl std::fmt::Display for Owner {
--- a/crates/iota-sdk-types/src/object.rs
+++ b/crates/iota-sdk-types/src/object.rs
@@ -118,10 +118,22 @@ impl Owner {
     /// object, and None if it is shared or immutable.
     pub fn address(&self) -> Option<&Address> {
         Some(match self {
-            Self::Address(addr) => addr,
-            Self::Object(obj_id) => obj_id.as_address(),
+            Self::Address(address) => address,
+            Self::Object(object_id) => object_id.as_address(),
             _ => return None,
         })
+    }
+}
+
+impl PartialEq<Address> for Owner {
+    fn eq(&self, other: &Address) -> bool {
+        self.as_address_opt() == Some(other)
+    }
+}
+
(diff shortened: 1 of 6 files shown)


crates/iota-sdk-types/src/object.rs
```diff
@@ -121,6 +121,28 @@ impl Owner {
     crate::def_is!(Immutable);
 
     crate::def_is_as_into_opt!(Address, Object(ObjectId), Shared(Version));
+
+    /// Returns an address if this object is owned by an address or
+    /// object, and None if it is shared or immutable.
+    pub fn address(&self) -> Option<&Address> {
+        Some(match self {
+            Self::Address(addr) => addr,
+            Self::Object(obj_id) => obj_id.as_address(),
```
thibault-martinez:
```suggestion
            Self::Object(object_id) => object_id.as_address(),
```

crates/iota-sdk-types/src/object.rs
```diff
@@ -121,6 +121,28 @@ impl Owner {
...
+            _ => return None,
+        })
+    }
+}
+
+impl PartialEq<Address> for Owner {
+    fn eq(&self, other: &Address) -> bool {
+        self.as_address_opt() == Some(other)
```
thibault-martinez, line 122: This feels a tiny bit confusing to me as we now also have `address()` that returns if it's Address or Object but I guess it's fine

DaughterOfMars: Yeah, I can see that. We could possibly call it `get_address` maybe?

crates/iota-sdk-types/src/object.rs
```diff
@@ -121,6 +121,28 @@ impl Owner {
     crate::def_is!(Immutable);
 
     crate::def_is_as_into_opt!(Address, Object(ObjectId), Shared(Version));
+
+    /// Returns an address if this object is owned by an address or
+    /// object, and None if it is shared or immutable.
+    pub fn address(&self) -> Option<&Address> {
+        Some(match self {
+            Self::Address(address) => address,
+            Self::Object(object_id) => object_id.as_address(),
```
Thoralf-M, line 114: Monorepo also has `get_address_owner_address()` which doesn't return the address if it's an object, don't want to have such a function here as well?

DaughterOfMars: It's called `as_address`

Thoralf-M (approved):

crates/iota-sdk-types/src/object.rs
Alex6323: I also find it confusing, any prefix like `get_`, `unwrap_`, `extract_` would be helpful I think.

thibault-martinez: I don't know if changing the name would do anything. We have a function that returns a Address that can be from the Address or Object variant, but comparing with an Address is only with the Address variant. It's just conceptually a bit confusing, not sure we can do anything about it.

Alex6323: Maybe we should add a comment to remind ourselves about this weirdness. But I am fine either way.

Alex6323 (approved):

thibault-martinez (approved):

PR #967, opened 2026-02-17
feat: improvements to Argument

thibault-martinez:


DaughterOfMars (approved):

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -1229,7 +1229,7 @@ pub struct Upgrade {
 /// argument-result         = %x02 u16
 /// argument-nested-result  = %x03 u16 u16
 /// ```
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
```
Alex6323, line 1231: Hmmm, the order is a bit uncommon now with `Clone`, `Copy` being last. I don't particularly care, but maybe it was accidental. I think typically we have `Clone`, Copy`, `Debug` first, and then the rest.

thibault-martinez: I copy pasted it from somewhere else, it's very inconsistent across the codebase. Do you mind if I just open an issue to add a script to sort them all and deal with this later?

Thoralf-M (approved):

crates/iota-sdk-types/src/transaction/mod.rs
DaughterOfMars: I don't see the point really

thibault-martinez: There is no point, it's purely cosmetic. I would probably do it at some point, but not really here where everything else is already inconsistent.

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/issues/974

PR #968, opened 2026-02-17
feat: Command improvements

thibault-martinez:


crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -1229,7 +1229,7 @@ pub struct Upgrade {
 /// argument-result         = %x02 u16
 /// argument-nested-result  = %x03 u16 u16
 /// ```
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
```
Alex6323, line 1706: style: not sure why `Clone`, `Copy` come last in this particular case?

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -1192,6 +1194,68 @@ impl Command {
...
+            Command::Upgrade(cmd) => Some(&cmd.modules),
+            Command::MoveCall(_)
+            | Command::TransferObjects(_)
+            | Command::SplitCoins(_)
+            | Command::MergeCoins(_)
+            | Command::MakeMoveVector(_) => None,
+        }
+    }
+
+    pub fn is_input_arg_used(&self, input_arg: u16) -> bool {
```
Thoralf-M: Add some comments like
    /// If this is a `Publish` or `Upgrade` command, return the modules to
    /// publish. Returns `None` for all other command types.


    /// Returns `true` if the command uses an input argument at the given index.

DaughterOfMars: This is another one of those methods I think should be defined locally, unless this is somehow used everywhere

crates/iota-sdk-types/src/transaction/mod.rs
thibault-martinez: No specific reason, it's generally inconsistent. We have an issue to order them.

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -1333,6 +1333,234 @@ impl Command {
         MakeMoveVector,
         Upgrade,
     );
+
+    pub fn new_move_call(
```
Alex6323, line 1338: you still probbly want to add docs to those public methods, right?

crates/iota-sdk-types/src/transaction/mod.rs
thibault-martinez: I've moved it back to the monorepo

crates/iota-sdk-types/src/transaction/mod.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/968/changes/4990e3c352d5a86c9dcfb0889f49be52fde0f9e3

DaughterOfMars (approved):

Alex6323 (approved):

Thoralf-M (approved):

PR #973, opened 2026-02-18
feat: MovePackage changes

thibault-martinez:


crates/iota-sdk-types/src/move_package.rs
```diff
@@ -90,6 +92,351 @@ impl MovePackageData {
...
+// ///
+// /// Directly modifying any field is undefined behavior. The fields are only
+// /// public for read-only access.
+// #[derive(Eq, PartialEq, Debug, Clone, Deserialize, Serialize, Hash,
+// JsonSchema)] pub struct UpgradeInfo {
+//     /// `Storage ID`/`Package ID` of the referred package.
+//     pub upgraded_id: ObjectID,
+//     /// The version of the package at `upgraded_id`.
+//     pub upgraded_version: SequenceNumber,
+// }
```
Thoralf-M: Why commented?

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -90,6 +92,319 @@ impl MovePackageData {
     }
 }
 
+/// Upgraded package info for the linkage table
```
Thoralf-M: For what linkage table? In the monorepo the comment is `[MovePackage]'s linkage_table` which is a bit more useful

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -90,6 +92,311 @@ impl MovePackageData {
...
+    #[serde(alias = "struct_name")]
+    pub datatype_name: Identifier,
+    /// ID of the package, where the given type first appeared.
+    pub package: ObjectId,
+}
+
+// // serde_bytes::ByteBuf is an analog of Vec<u8> with built-in fast
+// // serialization.
+// #[serde_as]
+// #[derive(Eq, PartialEq, Debug, Clone, Deserialize, Serialize, Hash)]
```
Alex6323: can we just call this field `type_name`? not sure what is gained with the `data` prefix tbh

thibault-martinez: I'd rather not do unnecessary changes

thibault-martinez: Plus maybe they chose this to explicit the fact it can only be a data type (struct or enum)

crates/iota-sdk-types/src/move_package.rs
thibault-martinez: Done

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -90,6 +92,270 @@ impl MovePackageData {
...
+    pub fn id(&self) -> ObjectId {
+        self.id
+    }
+
+    pub fn version(&self) -> Version {
+        self.version
+    }
+
+    pub fn decrement_version(&mut self) {
+        self.version.decrement().unwrap();
```
Thoralf-M: Unwrapping sounds bad here

Thoralf-M (approved):

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -90,6 +96,270 @@ impl MovePackageData {
...
+            self.linkage_table
+                .values()
+                .map(|UpgradeInfo { upgraded_id, .. }| upgraded_id),
+        )
+    }
+
+    /// It is important that this function is shared across both the calculation
+    /// of the digest for the package, and the calculation of the digest
+    /// on-chain.
+    pub fn compute_digest_for_modules_and_deps<'a>(
```
DaughterOfMars, line 252: This feels like it should be private. Can't we just use `digest` instead by creating a MovePackage from the inner values?

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -90,6 +96,270 @@ impl MovePackageData {
...
+        digest.finalize()
+    }
+
+    /// Retrieve the module from this package with the given [ObjectId] and
+    /// [Identifier].
+    ///
+    /// [ObjectId] is expected to contain the `Storage ID` of this package.
+    /// In case the `Storage ID` doesn't match or the module name is not
+    /// present in this package the function returns None.
+    pub fn get_module(&self, package: &ObjectId, name: &Identifier) -> Option<&Vec<u8>> {
```
DaughterOfMars: This fn feels like it is doing too much. Can we instead do the package ID check at the caller and just lookup using the ident?

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -90,6 +96,270 @@ impl MovePackageData {
...
+                     ..
+                 }| module_name.len() + datatype_name.len() + ObjectId::LENGTH,
+            )
+            .sum::<usize>();
+
+        let linkage_table_size = self.linkage_table.len()
+            * (ObjectId::LENGTH
+                + (
+                    ObjectId::LENGTH + 8
+                    // Version
```
DaughterOfMars: Weird comment?

thibault-martinez: Not weird, it's explaining what the 8 is, but fmt is not helping

crates/iota-sdk-types/src/move_package.rs
thibault-martinez: No, that would require the caller to provide lots of dummy values with is really not convenient

crates/iota-sdk-types/src/move_package.rs
thibault-martinez: It used to take a ModuleId, I had to split it in two. I think it's fine that it's checking the package as well, if we remove it people will forget the check

crates/iota-sdk-types/src/move_package.rs
DaughterOfMars: It's a strange fn to have on this type though. At least maybe it should be a free fn somewhere?

crates/iota-sdk-types/src/move_package.rs
DaughterOfMars: but from the perspective of this type, it's just an accessor

crates/iota-sdk-types/src/move_package.rs
DaughterOfMars: Should we have a const then?

thibault-martinez: What about `std::mem::size_of::<Version>()`?

Alex6323 (approved):

crates/iota-sdk-types/src/move_package.rs
DaughterOfMars: that works

crates/iota-sdk-types/src/move_package.rs
thibault-martinez: I don't think it's strange but I'll move it back to the monorepo if you want

thibault-martinez: Nah actually I'll keep it, digest use it, would have to duplicate code, I think it's fine

crates/iota-sdk-types/src/move_package.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/973/commits/89830853763f7f60545c69c14b206ca4c676d8ba

crates/iota-sdk-types/src/move_package.rs
thibault-martinez: I'll try but if it looks like a mess I'll revert

thibault-martinez: Was ok https://github.com/iotaledger/iota/pull/10364/changes/b206d24d47e99cb32ad58f3854aaf4bef66cc230

DaughterOfMars: Nice :)

crates/iota-sdk-types/src/move_package.rs
DaughterOfMars: Can we at least move it to `hash.rs` or something?

thibault-martinez: No I think it's where it should be here, it's tailored to MovePackage

DaughterOfMars: eh, ok

DaughterOfMars (approved):

PR #977, opened 2026-02-19
feat: Update `AuthenticatorStateUpdateV1` for core

DaughterOfMars:
## Description

Update `AuthenticatorStateUpdateV1` to prepare for core replacement.


thibault-martinez (approved):

Thoralf-M (approved):

crates/iota-sdk-types/src/crypto/zklogin.rs
thibault-martinez, line 1: Maybe comment the impls for now as we don't support ZKLogin yet

DaughterOfMars: Well we need them for the core, unless you'd rather we just keep using fastcrypto types, in which case we'd need to convert between those and our types for the AuthenticatorStateUpdate type

Alex6323 (approved):

crates/iota-sdk-types/src/crypto/zklogin.rs
thibault-martinez: You ended up removing them?

DaughterOfMars: Yeah, I decided to just keep using fastcrypto

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -494,6 +494,54 @@ pub struct ActiveJwk {
     pub epoch: u64,
 }
 
+impl PartialOrd for ActiveJwk {
```
thibault-martinez, line 498:
```suggestion
// This must match the sort order defined by jwk_lt in authenticator_state.move
impl PartialOrd for ActiveJwk {
```

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -494,6 +494,54 @@ pub struct ActiveJwk {
...
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+// This must match the sort order defined by jwk_lt in authenticator_state.move
+impl Ord for ActiveJwk {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        fn string_bytes_ord(a: &str, b: &str) -> std::cmp::Ordering {
```
thibault-martinez, line 507: Is it me or does this look like a very standard Ord impl? There's really no such a thing by default?

DaughterOfMars: I think the default is just a little bit different, but I'm not 100% sure tbh

thibault-martinez: Alright

PR #978, opened 2026-02-19
feat: add polling indexer example using graphql client

### LINKED ISSUE #511: [Bounty] Create indexer using SDK
Please read the [README](https://github.com/orgs/iotaledger/projects/97?pane=info) first.

### Summary

Create an IOTA indexer using the SDK (no other IOTA crates like `iota-data-ingestion-core`).

### Estimated complexity

Medium

### Estimated duration

2 days

**Summary**

Provide an example showing how to build a custom indexer on top of `iota_sdk::graphql_client::Client`. The indexer should demonstrate polling-based checkpoint ingestion, transaction/event filtering, and progress tracking using the existing Client API.

**Motivation**
Users building dApps or analytics services often need to index on-chain data into a custom store/database. The SDK already exposes the necessary building blocks — `checkpoints`, `transaction_blocks`, `events`, `TransactionsFilter` (with `after_checkpoint` / `before_checkpoint`), `EventFilter`, and cursor-based `PagedStream` — but there is no end-to-end example tying them together into an indexer loop.

**Proposed Approach (proposing an alternative is allowed)**

1. **Checkpoint-range polling loop** — sequentially walk checkpoints from a persisted watermark using `TransactionsFilter::after_checkpoint` / `before_checkpoint` to sync a specific range, fetch associated transaction blocks + effects, and advance the watermark.
2. **Event / package filtering** — narrow ingestion to a specific Move package or event type via `EventFilter::event_type` / `TransactionsFilter::function`.
3. **Optional event-driven trigger** — while the SDK currently has no WebSocket subscription support, an event polling shortcut (e.g., `events_stream` filtered by package) could be used to trigger a checkpoint-based re-sync rather than serve as the primary data source, since push-based delivery may miss events under load or on reconnect.
4. **Progress tracking** — persist the last successfully processed checkpoint sequence number so the indexer can resume from where it left off.

### Pull request required?

No

SeventhOdyssey71:
This PR implements a custom IOTA indexer example using `iota_sdk::graphql_client::Client` (SDK from this repository), aligned with bounty #511 acceptance criteria and explicitly avoiding `iota-data-ingestion-core`.

The example demonstrates:

- Polling-based checkpoint ingestion  
- Transaction and event filtering  
- PostgreSQL-backed database storage  

---

## What Changed

- Reworked `crates/examples/iota-indexer` into an SDK-client polling indexer.
- Added an indexer loop in `crates/examples/iota-indexer/src/indexer.rs` that:
  - Polls the latest checkpoint sequence
  - Processes checkpoints sequentially
  - Supports bounded (`start` / `end`) and continuous modes
  - Uses `TransactionsFilter::before_checkpoint` for checkpoint-scoped querying
- Implemented filtering support for:
  - `--event-type`
  - `--event-module`
  - `--event-package-id`
- Ensures safe resume behavior on restart.
- Updated database schema and initialization in  
  `crates/examples/iota-indexer/src/db/mod.rs` with:
  - Idempotent writes
  - Query-optimized indexes
- Updated CLI configuration in  
  `crates/examples/iota-indexer/src/config.rs`
- Updated documentation:
  - `crates/examples/iota-indexer/README.md`
  - `docs/indexer-plan.md`


crates/examples/iota-indexer/src/config.rs
```diff
@@ -0,0 +1,136 @@
...
+
+fn default_graphql_url(network: Network) -> &'static str {
+    match network {
+        Network::Mainnet => "https://graphql.mainnet.iota.cafe",
+        Network::Testnet => "https://graphql.testnet.iota.cafe",
+        Network::Devnet => "https://graphql.devnet.iota.cafe",
+        Network::Localnet => "http://localhost:9125/graphql",
+        Network::Custom => "https://graphql.testnet.iota.cafe",
+    }
+}
```
Alex6323: It would be better if that would be a method on `Network` that only requires `&self` instead of `self`. The `Custom` variant should hold a URL.

crates/examples/iota-indexer/src/indexer.rs
```diff
@@ -0,0 +1,319 @@
...
+                Err(err) => {
+                    warn!(checkpoint = next_checkpoint, error = %err, "checkpoint processing failed; retrying");
+                    tokio::time::sleep(Duration::from_millis(self.config.poll_interval_ms)).await;
+                }
+            }
+        }
+    }
+
+    async fn process_checkpoint(&self, sequence: u64) -> anyhow::Result<bool> {
+        let Some(checkpoint) = self.client.checkpoint(None, Some(sequence)).await? else {
```
Alex6323, line 302: nit: you can just pass `sequence` without creating the `Option` yourself.

Alex6323 (commented):
PR looks good to me! But can you move it to the `iota-sdk/examples`?

crates/examples/iota-indexer/README.md
```diff
@@ -0,0 +1,90 @@
+# iota-indexer
```
Alex6323: I would suggest to call the project/example `polling-indexer` instead to make the implementation goals clearer. I also feel like in the future we might want to distinguish it from an also possible `streaming-indexer`.

SeventhOdyssey71: alright, would do this! it makes a lot of sense to make it all well structured.

SeventhOdyssey71:
Thanks for the review.

  Addressed all comments in the latest push:

  - Moved the example from crates/examples/iota-indexer to crates/iota-sdk/examples/polling-indexer
  - Renamed the example/package/docs/CLI references from iota-indexer to polling-indexer to make the polling behavior explicit
  - Applied the nit in process_checkpoint by passing sequence directly to checkpoint(...) (instead of wrapping it in Some(...))

  Also ran cargo check -p polling-indexer to confirm it still builds.

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -344,7 +344,7 @@ impl<C, L> TransactionBuilder<C, L> {
     ///
     /// # Example
     ///
-    /// ```
+    /// ```no_run
```
DaughterOfMars: Why did you add `no_run` to all of these examples?

crates/iota-sdk/examples/polling-indexer/README.md
```diff
@@ -0,0 +1,91 @@
...
+On restart, indexing resumes from the stored value.
+
+## Acceptance-Criteria Mapping
+
+- Uses SDK client from this repo: `iota_sdk::graphql_client::Client`
+- Polling checkpoint ingestion loop implemented
+- Uses transaction range filters with `after_checkpoint` / `before_checkpoint`
+- Filters transactions/events
+- Persists progress and resumes from watermark
+- Stores indexed data in database
```
DaughterOfMars: I don't think this section is needed

docs/indexer-plan.md
DaughterOfMars, line 1: You can delete this

crates/iota-sdk/examples/polling-indexer/README.md
SeventhOdyssey71: That was meant for just my /in-work purpose. I'd remove that.

.gitignore
```diff
@@ -1,10 +1,12 @@
...
 .idea
+.DS_Store
 *.dylib
 *.so
 *.dll
 __pycache__
 build
 .vscode
 .kotlin
+help/
```
DaughterOfMars: What is the reason for these new ignores?

Makefile
```diff
@@ -224,7 +224,7 @@ example:
 
 .PHONY: examples
 examples: ## Run all Rust examples
-	@for example in $$(find crates/iota-sdk/examples -name "*.rs" -not -path "*/release/*" -exec basename {} .rs \;); do \
+	@for example in $$(find crates/iota-sdk/examples -maxdepth 1 -name "*.rs" -exec basename {} .rs \;); do \
```
DaughterOfMars: Why?

.gitignore
SeventhOdyssey71: help/ is basically a folder with some things i noted while building, as well as things I found to be important.

i'd standardize this properly now. DS_Store is macOS Finder metadata that was being generated in the local helper workspace.

docs/indexer-plan.md
SeventhOdyssey71: alright, I'd do that.

crates/iota-sdk/examples/polling-indexer/src/config.rs
DaughterOfMars, line 1: I would divide the CLI and config into two files

crates/iota-sdk/examples/polling-indexer/src/db/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    .execute(pool)
+    .await?;
+    sqlx::query("CREATE INDEX IF NOT EXISTS idx_events_checkpoint ON events(checkpoint_seq);")
+        .execute(pool)
+        .await?;
+
+    Ok(())
+}
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
```
DaughterOfMars: 🤔

SeventhOdyssey71: just copying the standard made in other code files.. should this be taken out?

DaughterOfMars: It's at the bottom of the file 😆

SeventhOdyssey71: my bad, would fix that too.

crates/iota-sdk/examples/polling-indexer/README.md
```diff
@@ -0,0 +1,82 @@
...
+  -e POSTGRES_DB=polling_indexer \
+  -p 5432:5432 -d postgres:16
+```
+
+## Run the Indexer
+
+```bash
+cargo run -p polling-indexer -- \
+  --network testnet \
+  --db-url postgres://localhost:5432/polling_indexer \
```
Thoralf-M: The url needs to have the credentials added?
`postgres://postgres:postgres@localhost:5432/polling_indexer`

crates/iota-sdk/examples/polling-indexer/README.md
```diff
@@ -0,0 +1,82 @@
+# polling-indexer
+
+Polling-based custom indexer example built on top of `iota_sdk::graphql_client::Client`.
+
+This example demonstrates the bounty-required flow:
+
```
Thoralf-M: Shouldn't mention bounty-required here, just explain what it does

.gitignore
Thoralf-M, line 1: Add `.polling_indexer_progress.json` here? This gets generated when running the indexer

crates/iota-sdk/examples/polling-indexer/README.md
SeventhOdyssey71: alright, I'd fix that.

crates/iota-sdk/examples/polling-indexer/README.md
```diff
@@ -0,0 +1,82 @@
...
+```bash
+--event-type 0x...::module::EventName
+--event-module module_name
+--event-package-id 0x...package
+```
+
+## Progress Tracking
+
+A JSON progress file stores the `next_checkpoint` watermark.
+On restart, indexing resumes from the stored value.
```
Thoralf-M: Should probably mention the default filename here

crates/iota-sdk/examples/polling-indexer/README.md
```diff
@@ -0,0 +1,82 @@
...
+- transaction ingestion using `TransactionsFilter::after_checkpoint` / `before_checkpoint`,
+- event ingestion + filtering,
+- storage into PostgreSQL.
+
+## What It Indexes
+
+- `checkpoints` table: checkpoint summary metadata + raw JSON
+- `transactions` table: tx digest, sender, kind, status + raw JSON
+- `events` table: package/module/type data + raw JSON
+
```
Thoralf-M, line 18: Can you add some example for how to read this data then?

crates/iota-sdk/examples/polling-indexer/src/cli.rs
```diff
@@ -0,0 +1,122 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{fmt, str::FromStr};
+
+use clap::Parser;
+
+#[derive(Debug, Clone, Parser)]
+#[command(name = "polling-indexer")]
+#[command(about = "Custom polling indexer using iota_sdk::graphql_client::Client")]
+pub struct Cli {
```
Thoralf-M, line 12: There are quite many config options, how about adding support to read the config from a file?

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
```diff
@@ -0,0 +1,322 @@
...
+        }
+    }
+
+    pub async fn run(&self) -> anyhow::Result<()> {
+        let state = progress::load(&self.progress_path).await?;
+        let mut next_checkpoint = state.next_checkpoint;
+
+        if let Some(start) = self.config.start_checkpoint {
+            next_checkpoint = next_checkpoint.max(start);
+        }
```
Thoralf-M: If the provided start_checkpoint is larger than `state.next_checkpoint + 1` it would be good to print a warning that checkpoints are skipped

crates/iota-sdk/examples/polling-indexer/src/config.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::cli::{Cli, Network};
+
+const DEFAULT_DB_URL: &str = "postgres://localhost:5432/polling_indexer";
+const DEFAULT_PROGRESS_FILE: &str = ".polling_indexer_progress.json";
+const DEFAULT_PAGE_SIZE: i32 = 50;
+const DEFAULT_POLL_INTERVAL_MS: u64 = 2000;
+const DEFAULT_INCLUDE_FAILED_TXS: bool = true;
+
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(default, deny_unknown_fields)]
```
Thoralf-M, line 19: Can you add some comments for the fields, best with some example?
For the `event_module` for example I think it might be confusing, as it's compared against the `sending_module` from the Event, this is not necessarily the module in which the Event type was defined, but where its emitted, perhaps rename it to `event_sending_module`

SeventhOdyssey71: Yeah, sure thing! I'd add comments for these.

crates/iota-sdk-transaction-builder/src/lib.rs
```diff
@@ -21,7 +21,7 @@
 //!
 //! ### Example with Client Resolution
 //!
-//! ```
+//! ```no_run
```
DaughterOfMars: Still one left

SeventhOdyssey71: that's been fixed now.

docs/indexer-plan.md
Thoralf-M: Still here

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
```diff
@@ -0,0 +1,330 @@
...
+                Err(err) => {
+                    warn!(checkpoint = next_checkpoint, error = %err, "checkpoint processing failed; retrying");
+                    tokio::time::sleep(Duration::from_millis(self.config.poll_interval_ms)).await;
+                }
+            }
+        }
+    }
+
+    async fn process_checkpoint(&self, sequence: u64) -> anyhow::Result<bool> {
+        let Some(checkpoint) = self.client.checkpoint(None, sequence).await? else {
```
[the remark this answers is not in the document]
Thoralf-M: Accepts `impl Into<Option<u64>>` so it compiles fine

SeventhOdyssey71:
I addressed all of the flagged issues and pushed the fixes (latest commit: `42de51f`).

  Changes made:
  - Removed `db_url` from startup logging to avoid leaking PostgreSQL credentials.
  - Fixed the resume warning off-by-one logic (`start_checkpoint > stored_next_checkpoint` now warns correctly).
  - Refactored checkpoint processing to avoid holding a DB transaction open across network calls (writes remain idempotent via `UPSERT` / `ON CONFLICT DO NOTHING`).
  - Reworked event de-duplication to avoid `UNIQUE(transaction_digest, raw_json)` on `JSONB`; now uses a unique index on `(transaction_digest, md5(raw_json::text))`.
  - Added a small schema migration step to drop the old event unique constraint if it already exists locally.
  - Updated README SQL examples to match the actual schema column names.
  - Hardened progress-file writes for Windows by removing the destination file before `rename`.

  Note on the `Client::checkpoint` comment:
  - The original call compiled because the API accepts `impl Into<Option<u64>>`, but I changed it to `Some(sequence)` for clarity and to avoid ambiguity in review.

  Validation:
  - `cargo check -p polling-indexer`

crates/iota-sdk/examples/polling-indexer/src/progress.rs
```diff
@@ -42,6 +42,13 @@ pub async fn store(path: &Path, next_checkpoint: u64) -> anyhow::Result<()> {
 
     let tmp_path = path.with_extension("tmp");
     fs::write(&tmp_path, serde_json::to_vec_pretty(&state)?).await?;
+    // `rename` does not replace an existing file on all platforms (notably
```
DaughterOfMars: Why are you tracking progress with a file anyway? Why not use a table?

SeventhOdyssey71: I used a local progress file on purpose for this example to keep the setup simple and aligned with the bounty/reference pattern. Postgres already stores the indexed checkpoints, transactions, and events, the file is just a small `next_checkpoint` cursor so it can restart cleanly.

a DB-backed progress table makes more sense for production, so i'd do that.

SeventhOdyssey71:
made a push for updates.. please confirm if there's still any issues. Thanks for the review.

crates/iota-sdk/examples/polling-indexer/src/db/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+        "#,
+    )
+    .execute(pool)
+    .await?;
+
+    sqlx::query(
+        "ALTER TABLE events DROP CONSTRAINT IF EXISTS events_transaction_digest_raw_json_key;",
+    )
+    .execute(pool)
+    .await?;
```
DaughterOfMars: I'm guessing this was to update the table after removing that constraint? So we shouldn't need it...

crates/iota-sdk/examples/polling-indexer/src/cli.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub event_package_id: Option<String>,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Network {
+    Mainnet,
+    Testnet,
+    Devnet,
+    Localnet,
+    Custom(String),
```
DaughterOfMars: Really no need for this

crates/iota-sdk/examples/polling-indexer/src/config.rs
```diff
@@ -0,0 +1,164 @@
...
+    #[serde(alias = "progress_file")]
+    progress_key: Option<String>,
+    /// First checkpoint to index (inclusive).
+    start_checkpoint: Option<u64>,
+    /// Last checkpoint to index (inclusive). Omit for continuous mode.
+    end_checkpoint: Option<u64>,
+    /// GraphQL page size per request (must be > 0).
+    page_size: Option<i32>,
+    /// Polling interval in milliseconds when waiting for new checkpoints.
+    poll_interval_ms: Option<u64>,
```
DaughterOfMars, line 42: Can you make this a Duration, please

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
```diff
@@ -0,0 +1,321 @@
...
+                    warn!(
+                        checkpoint = next_checkpoint,
+                        "checkpoint not indexed yet; retrying"
+                    );
+                    tokio::time::sleep(Duration::from_millis(self.config.poll_interval_ms)).await;
+                }
+                Err(err) => {
+                    warn!(checkpoint = next_checkpoint, error = %err, "checkpoint processing failed; retrying");
+                    tokio::time::sleep(Duration::from_millis(self.config.poll_interval_ms)).await;
+                }
```
DaughterOfMars, line 248: There's currently no logic for retry attempts or backoff or anything. Would be nice to have something like that.

crates/iota-sdk/examples/polling-indexer/src/config.rs
SeventhOdyssey71: alright, would specify.

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
```diff
@@ -0,0 +1,321 @@
...
+        _ => None,
+    }
+}
+
+fn tx_kind_str(tx: &SignedTransaction) -> String {
+    match &tx.transaction {
+        Transaction::V1(v1) => format!("{:?}", v1.kind),
+        _ => "unknown".to_owned(),
+    }
+}
```
DaughterOfMars, line 516: Not sure this is doing what was intended. The debug impl will print the entire structure, not just the variant name.

crates/iota-sdk/examples/polling-indexer/src/db/progress.rs
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use serde::{Deserialize, Serialize};
+use sqlx::{PgPool, Row};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ProgressState {
```
DaughterOfMars, line 10: Shouldn't this file be in the `db` module now?

crates/iota-sdk/examples/polling-indexer/src/config.rs
```diff
@@ -17,8 +17,8 @@ const DEFAULT_INCLUDE_FAILED_TXS: bool = true;
 #[derive(Debug, Clone, Default, Deserialize)]
 #[serde(default, deny_unknown_fields)]
 struct FileConfig {
-    /// Network name or custom endpoint selector.
-    /// Examples: "testnet", "mainnet", "custom:https://your.graphql.endpoint/graphql"
+    /// Network preset name.
+    /// Examples: "testnet", "mainnet", "localnet"
     network: Option<String>,
```
DaughterOfMars: We should instead use the `Network` type here

crates/iota-sdk/examples/polling-indexer/src/db/mod.rs
DaughterOfMars: Still there

crates/iota-sdk/examples/polling-indexer/src/config.rs
SeventhOdyssey71: I changed `FileConfig.network` to use the existing `Network` type directly, and added serde deserialization on `Network` so JSON config still accepts the same string values
  (`"testnet"`, `"mainnet"`, etc.).

.gitignore
DaughterOfMars: Does it still get generated now that it's using the database table?

Makefile
```diff
@@ -276,7 +276,7 @@ example:
 
 .PHONY: examples
 examples: ## Run all Rust examples
-	@for example in $$(find crates/iota-sdk/examples -name "*.rs" -not -path "*/release/*" -exec basename {} .rs \;); do \
+	@for example in $$(find crates/iota-sdk/examples -maxdepth 1 -type f -name "*.rs" -exec basename {} .rs \;); do \
```
DaughterOfMars, line 330: If the intention was to run this new example with this command, I don't think this will work.

.gitignore
Thoralf-M: No, just looking at this now again, resolving

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
```diff
@@ -0,0 +1,357 @@
...
+        if let Some(start) = self.config.start_checkpoint {
+            if start > state.next_checkpoint {
+                warn!(
+                    configured_start_checkpoint = start,
+                    stored_next_checkpoint = state.next_checkpoint,
+                    skipped_checkpoints = start - state.next_checkpoint,
+                    "configured start_checkpoint skips checkpoints after stored progress"
+                );
+            }
+            next_checkpoint = next_checkpoint.max(start);
```
Thoralf-M: With .max() it's impossible to sync older checkpoints again if someone skipped some by accident
```
cargo run -p polling-indexer -- \
  --network testnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --progress-key polling-indexer \
  --start-checkpoint 20
```

and then  later 
```
cargo run -p polling-indexer -- \
  --network testnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --progress-key polling-indexer \
  --start-checkpoint 10
```
will just continue with the checkpoints >20, making it impossible to sync the ones before

SeventhOdyssey71: fixed. next_checkpoint now uses the configured --start-checkpoint value directly instead of .max(), so re-syncing older checkpoint ranges works as expected. A warning is logged when the configured value differs from stored progress.

Makefile
SeventhOdyssey71: I've added a clarifying comment to that effect and a dedicated make polling-indexer target (cargo run -p polling-indexer -- $(ARGS)) for manual runs.

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
Thoralf-M: Starting to sync with a lower start checkpoint will also resync all the known checkpoints above, they should be skipped if already available in the db

Makefile
```diff
@@ -276,10 +276,17 @@ example:
...
+	@# NOTE: -maxdepth 1 -type f excludes multi-crate examples like polling-indexer
+	@# that require external services (e.g. PostgreSQL). Run those separately.
 	@for example in $$(find crates/iota-sdk/examples -maxdepth 1 -type f -name "*.rs" -exec basename {} .rs \;); do \
 		$(MAKE) example "$$example" || exit $$?; \
 	done
 
+.PHONY: polling-indexer
+polling-indexer: ## Run the polling-indexer example (requires PostgreSQL)
+	@printf "\nRunning polling-indexer example\n"
+	@cargo run -p polling-indexer -- $(ARGS)
```
DaughterOfMars: it would be nice to run this in the CI

SeventhOdyssey71: sure thing, added that.

SeventhOdyssey71:
start_checkpoint logic @Thoralf-M: Replaced .max() with direct use of the configured --start-checkpoint value, so re-syncing older checkpoint ranges is now possible. Already-indexed checkpoints are skipped via a DB lookup to avoid redundant work.

Makefile/CI @DaughterOfMars: Added a clarifying comment to the examples target explaining why polling-indexer is excluded from auto-discovery, added a dedicated make polling-indexer target, and added a CI step in tests.yml that runs the polling-indexer for checkpoints 0–5 against the localnet PostgreSQL.

 Cleanup: Removed the help/ gitignore entry.

DaughterOfMars (approved):

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
```diff
@@ -0,0 +1,369 @@
...
+
+    async fn checkpoint_exists(&self, sequence: u64) -> anyhow::Result<bool> {
+        let row = sqlx::query("SELECT 1 FROM checkpoints WHERE sequence_number = $1")
+            .bind(sequence as i64)
+            .fetch_optional(&self.pool)
+            .await?;
+        Ok(row.is_some())
+    }
+
+    async fn process_checkpoint(&self, sequence: u64) -> anyhow::Result<bool> {
```
Thoralf-M, line 296: Sorry for noticing that just now, but processing the checkpoints one by one is a huge bottleneck
There are new checkpoints every ~250ms, requesting each individually might never be able to sync faster than new checkpoints are coming is if there is some network delay (since txs are also queried for each checkpoint, it's more than a single request for each checkpoint...)

Also another thing, if one wants to filter by a package id like this (IOTA-Names on mainnet)
```
cargo run -p polling-indexer -- \
  --network mainnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --tx-function 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83 \
  --start-checkpoint 96139267 # package was published here, so no point in checking earlier checkpoints
```
I would argue in this case, one isn't interested in any checkpoints, but just the transactions(or events with the other filter). One could just request transactions with a much larger checkpoint range (like 100k-1million even) and sync the data in minutes instead of hours/days

SeventhOdyssey71: Just addressed this. When any transaction or event filter is active, the indexer now uses a batch query mode that queries transactions across 100k-checkpoint ranges in a single paginated stream instead of fetching each checkpoint individually. This lets filtered workloads (like tracking a specific package) sync in minutes instead of hours/days.

The per-checkpoint path remains for unfiltered full indexing. Schema was updated to make checkpoint_seq nullable since TransactionDataEffects doesn't include the checkpoint sequence number.

crates/iota-sdk/examples/polling-indexer/README.md
```diff
@@ -0,0 +1,147 @@
...
+```bash
+--tx-function 0x2::iota_system::request_add_stake
+--tx-sender 0x...address
+--include-failed-txs false
+```
+
+Event-level:
+
+```bash
+--event-type 0x...::module::EventName
```
Thoralf-M, line 108: Why does running
```
cargo run -p polling-indexer -- \
  --network mainnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --event-type 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83::registry::NameRecordAddedEvent \
  --start-checkpoint 96139267
```
not store any events, but unrelated transactions instead that emitted an event (not the one I provided)?

```
polling_indexer=# SELECT checkpoint_seq, transaction_digest, package_id, module, event_name
FROM events
ORDER BY checkpoint_seq DESC, id DESC
LIMIT 20;
 checkpoint_seq | transaction_digest | package_id | module | event_name
----------------+--------------------+------------+--------+------------
(0 rows)

polling_indexer=# SELECT checkpoint_seq, transaction_digest, sender, success, kind
FROM transactions
ORDER BY checkpoint_seq DESC, id DESC
LIMIT 20;
 checkpoint_seq |              transaction_digest              |                               sender                               | success |             kind
----------------+----------------------------------------------+--------------------------------------------------------------------+---------+------------------------------
                | 4vnM7EzxdGCVFd97FvbNngwkZvu1Xts43fJAc7YBMP9L | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | 2KCk7BxkbgT9pcjpivfc4enTRAEWBa2vnKaFLnZ8yjLq | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | Ds8SAZAMGbHQ2G8zmSD81Hv2iE98oLa7FPFFoPPwCB98 | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | randomness_state_update
                | G3XcUCYGUntesnDjZ5qQof8i9ATSvuCgfMt6xdDVBCWJ | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | FcpaNp2rXvPHcn6uMWTxTQL3NX1Jxd6PKKZhEmMverC5 | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | 2Tg9HazvRpj7UNVAY9zbhimtuWdKKaBgF8McS1MP3P8f | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | BNYroR5tdQHV6ksercfkxmrSg4WQGcHzPoV69jSYonJi | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | G7XuSjowkLDMR9kTmprg58AL4VLJMSwsWL8aPhAWtXJn | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | HwJc8iJ7HwVCiwUYfoNwGAug6ERpXGabmxCNTD9HoJHb | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | 4iqibZrXX8U4Gun6PruoLZ7nZxSbiBECmihx2kPNs7RY | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | DvSYmr3uzQQArGk3j14UKhAST25cyAp9L7kbJa3NEnra | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | ATgB1tGjjNciTajmMWVPqz3pZCEtFqjsCUMHnWpVbMHk | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | 8tMk5n61UjMS1rRWNDdJE4iBykU92LbzF7VKgs4rSSZg | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | HxHCJ4mtixHtW2SA4Q7gWWApX9LttwvaShqevjK7g2qg | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | 8DLZH5N54KaucHo59KJDKUzfuQTNWWkcpiujgGX9Aq3S | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | randomness_state_update
                | 9habMRTSN6Q9aBeBFqATniebDAJyYSTVLs4ckNN23mbo | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | 2mXZDAsVWqwWkSosRePeNJ4Mb27tPU3ngkynZXnw6WBY | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | 5VpfUQNYTSgxWkS3qU81yGgJq1sjUZajwp5hytJvd8iz | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | 49YfUfmWxUi4JMsVBSUeMZjeswqdr9mNjmtyzQjpmBnA | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
                | VDDRRBzR9kdWKVTvAYsuD8VMhMPYB1xWtvEsAzJQoTy  | 0x0000000000000000000000000000000000000000000000000000000000000000 | t       | consensus_commit_prologue_v1
```

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
```diff
@@ -0,0 +1,531 @@
...
+use tracing::{debug, info, warn};
+
+use crate::{
+    config::{AppConfig, FilterConfig},
+    db::progress,
+};
+
+/// Default number of checkpoints to cover in a single batch query when filters
+/// are active. Large ranges let the GraphQL server do the heavy lifting.
+const DEFAULT_BATCH_RANGE: u64 = 100_000;
```
Thoralf-M: No way to overwrite this in the config?

crates/iota-sdk/examples/polling-indexer/README.md
SeventhOdyssey71: process_batch now queries events for each transaction before deciding whether to store it. When only event filters are set (no --tx-function/--tx-sender), transactions without matching events are skipped entirely. This prevents unrelated system transactions from being stored.

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
SeventhOdyssey71: Added --batch-range CLI flag and batch_range JSON config field (default: 100,000). This can be tuned depending on the use case, smaller ranges for denser checkpoints, larger for sparse filtered queries.

crates/iota-sdk/examples/polling-indexer/README.md
Thoralf-M: If there is only an --event-type filter it should as for transactions not go through every checkpoint and all txs, but just filter by this event type, this is much faster.
Please actually try this
```
cargo run -p polling-indexer -- \
  --network mainnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --event-type 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83::registry::NameRecordAddedEvent \
  --start-checkpoint 96139267
```
should be able to sync all the events of this type in <2 minutes

The option to go through every checkpoint should probably only be used if there is no filtering at all

Also make sure that the db values are actually set, in the output I had above, checkpoint_seq was never set for any entry

SeventhOdyssey71: Just resolved this. Now, when only event filters are set, the indexer now queries events directly via the events API (client.events(EventFilter { event_type })) with cursor-based pagination instead of iterating through all transactions. No checkpoint/transaction iteration at all.

Thoralf-M:
Running
```
cargo run -p polling-indexer -- \
  --network mainnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --tx-function 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83 \                               
  --start-checkpoint 9613926
```
still has no checkpoint_seq for the txs
```
polling_indexer=# SELECT checkpoint_seq, transaction_digest, sender, success, kind
FROM transactions
ORDER BY checkpoint_seq DESC, id DESC
LIMIT 20;

 checkpoint_seq |              transaction_digest              |                               sender                               | success |           kind
----------------+----------------------------------------------+--------------------------------------------------------------------+---------+--------------------------
                | C7rcpajZddZpioTCVaCLkUWvuZjh7npKz9LPPXCouvsd | 0x182f8940ce9ed90ec5be805f085ebe1d015e09e0dbd5f297e26f5da24a8b6c3e | t       | programmable_transaction
                | 2fKESfgSDnSTUJLM4mWr1znR5rN2WLoRW75yHixTkTo6 | 0x182f8940ce9ed90ec5be805f085ebe1d015e09e0dbd5f297e26f5da24a8b6c3e | t       | programmable_transaction
```
same for events
```
cargo run -p polling-indexer -- \
  --network mainnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --event-type 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83::registry::NameRecordAddedEvent \
  --start-checkpoint 96139267
```

```
polling_indexer=# SELECT checkpoint_seq, transaction_digest, package_id, module, event_name
FROM events
ORDER BY checkpoint_seq DESC, id DESC
LIMIT 20;
 checkpoint_seq |              transaction_digest              |                             package_id                             |   module   |      event_name
----------------+----------------------------------------------+--------------------------------------------------------------------+------------+-----------------------
                | 59WwWyK9YVt37CKJknJciFHifkQxqezmcea5BXJch5YY | 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83 | controller | UserDataSetEvent
                | 9wrMSFJJcNWUJW4c2s4tQGqvPAeV1t3iNxqJ6gR3g29t | 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83 | payment    | NameRecordAddedEvent
```

SeventhOdyssey71:
Screenshot 2026-03-05 at 17 39 14

fixed the missing `checkpoint_seq` row, it now populates in there

Thoralf-M:
> fixed the missing `checkpoint_seq` row, it now populates in there

For txs yes, but still not for events
```
cargo run -p polling-indexer -- \
  --network mainnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --event-type 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83::registry::NameRecordAddedEvent \
  --start-checkpoint 96139267
```
there it's also missing the tx digest now
```
 checkpoint_seq | transaction_digest |                             package_id                             |  module  |      event_name
----------------+--------------------+--------------------------------------------------------------------+----------+----------------------
                |                    | 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83 | payment  | NameRecordAddedEvent
                |                    | 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83 | payment  | NameRecordAddedEvent
```

SeventhOdyssey71:
Fixed — when --event-type is set, the package address is now automatically derived from the event type string (e.g. 0x6d2c...::registry::NameRecordAddedEvent → package  0x6d2c...) and used as a transaction filter in batch mode.

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
```diff
@@ -0,0 +1,730 @@
...
+                    .as_ref()
+                    .map(|m| m.package.address.to_string());
+                let module = event.sending_module.as_ref().map(|m| m.name.clone());
+                let sender = event.sender.as_ref().map(|s| s.address.to_string());
+                let event_type = event.type_.repr.clone();
+                let event_name = event_type
+                    .rsplit("::")
+                    .next()
+                    .unwrap_or(event_type.as_str())
+                    .to_owned();
```
Thoralf-M: This doesn't work with generics like `0x2::display::DisplayCreated<0x107a::nft::Nft>` -> `Nft>`:
```
polling_indexer=# SELECT checkpoint_seq, transaction_digest, package_id, module, event_name
FROM events
ORDER BY checkpoint_seq DESC, id DESC
LIMIT 20;
checkpoint_seq | transaction_digest | package_id | module | event_name
----------------+----------------------------------------------+--------------------------------------------------------------------+--------+------------
0 | 63X49x2QuuYNduExZWoJjfXut3s3WDWZ7Tr7nXJu32ZT | 0x000000000000000000000000000000000000000000000000000000000000107a | nft | Nft>
```

crates/iota-sdk/examples/polling-indexer/src/db/progress.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub next_checkpoint: u64,
+    pub last_cursor: Option<String>,
```
Thoralf-M: AFAIK the cursor in graphql is only valid for ~30 minutes, so for txs and event filter should probably store the checkpoint of the last received tx/event and provide that when trying again or switch to sync by a checkpoint range once the current live checkpoints are reached and then store the last checkpoint index/next_checkpoint

SeventhOdyssey71: Cursors are indeed short-lived and unreliable for persistence. Removed last_cursor from `ProgressState` and the store_cursor() function entirely. Both modes now track progress purely by checkpoint:                                                                                                                                  
                  
- Batch filtered mode stores the next range_start checkpoint after each batch completes, and uses after_checkpoint/before_checkpoint in TransactionsFilter to resume.

- Per-checkpoint mode stores next_checkpoint as before.

Cursors are still used within a single paginated loop (e.g. tx_cursor in process_batch), but only as ephemeral in-memory variables, and never persisted to the DB.

crates/iota-sdk/examples/polling-indexer/src/indexer.rs
```diff
@@ -0,0 +1,593 @@
...
+
+        info!(range_start, range_end, tx_count, "batch complete");
+        Ok(())
+    }
+
+    // ------------------------------------------------------------------
+    // Per-checkpoint mode: full indexing without filters
+    // ------------------------------------------------------------------
+
+    async fn run_per_checkpoint(&self) -> anyhow::Result<()> {
```
Thoralf-M: Functionality seems fine now :tada: But the code doesn't look very well structured and could use some refactoring, here are some ideas:
1. Duplicated polling loop skeleton

run_filtered() and run_per_checkpoint() share the same outer structure — load progress, loop, fetch latest checkpoint, compare bounds, handle failures with backoff, persist progress. The only difference is the "process one unit" step. This should be extracted into a single generic loop that accepts a processing closure/trait:

2. Duplicated retry/backoff logic

Both loops manually manage consecutive_failures and call retry_delay(). This belongs in a small dedicated struct:

3. Validation deferred to runtime instead of config construction

The guard at the top of run() — checking that event filters have a derivable package — should fail at config-construction time in [TryFrom<Cli>](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for [AppConfig](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), not silently pass through parsing and only blow up when the indexer starts.

4. only_event_filters is a local variable that belongs on FilterConfig

process_batch computes only_event_filters inline. Since it's derived entirely from FilterConfig, it should be a method there: filters.is_event_only().

5. Extra per-transaction round-trip in process_batch

lookup_tx_checkpoint is called once per transaction, making a separate GraphQL query each time. This is both a performance problem and a structural smell — the checkpoint info should ideally come from the transaction query itself, or at minimum be batched.

SeventhOdyssey71: Took a while, but this was what i was able to put together: 

1. Unified polling loop — run_filtered() and run_per_checkpoint() collapsed into a single run() method. A `ProcessResult` enum (Advance(u64) / NotReady) handles the
  tri-state from per-checkpoint mode. The shared skeleton (load progress, fetch latest, bounds check, backoff, progress store) is written once, with a simple if filtered {
  ... } else { ... } dispatch inside the loop.
  2. `RetryState` struct — Replaces the duplicated consecutive_failures + retry_delay() pattern. Three methods: reset(), failures(), next_delay().
  3. Validation at config time — The event-filter guard moved from Indexer::run() into TryFrom<Cli> for AppConfig, so invalid configs fail at startup before any network/DB
  connections.
  4. FilterConfig::is_event_only() — Replaces the inline has_event_filters() && !has_tx_filters() computation in process_batch.
  5. Batched checkpoint lookups — lookup_tx_checkpoint (1 query per tx) replaced by batch_lookup_tx_checkpoints, which builds a single aliased GraphQL query for all
  transactions in a page, reducing N round-trips to 1 per page.

Thoralf-M (approved):
Thanks for the PR!
Working well now, I tried these 3 things with checkpoints, tx filter and event filter each time also with stopping and restarting the indexer (it always resumed where it should):
```
# start postgres
docker run --name polling-indexer-pg \
  -e POSTGRES_PASSWORD=postgres \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_DB=polling_indexer \
  -p 5432:5432 -d postgres:16
# run indexer with checkpoints
cargo run -p polling-indexer -- \
  --network testnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --progress-key polling-indexer \
  --start-checkpoint 0 \
  --end-checkpoint 50
# check database 
docker exec polling-indexer-pg psql "postgres://postgres:postgres@localhost:5432/polling_indexer" -c "SELECT sequence_number, timestamp_ms, digest, raw_json FROM checkpoints ORDER BY sequence_number DESC LIMIT 10;"
# cleanup postgres container
docker rm -f polling-indexer-pg


# start postgres
docker run --name polling-indexer-pg \
  -e POSTGRES_PASSWORD=postgres \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_DB=polling_indexer \
  -p 5432:5432 -d postgres:16
# run indexer with events filter
cargo run -p polling-indexer -- \
  --network mainnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --event-type 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83::registry::NameRecordAddedEvent \
  --start-checkpoint 96139267
# check database 
docker exec polling-indexer-pg psql "postgres://postgres:postgres@localhost:5432/polling_indexer" -c "SELECT checkpoint_seq, transaction_digest, package_id, module, event_name, raw_json FROM events ORDER BY checkpoint_seq DESC, id DESC LIMIT 20;"
# cleanup postgres container
docker rm -f polling-indexer-pg


# start postgres
docker run --name polling-indexer-pg \
  -e POSTGRES_PASSWORD=postgres \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_DB=polling_indexer \
  -p 5432:5432 -d postgres:16
# run indexer with events filter
cargo run -p polling-indexer -- \
  --network mainnet \
  --db-url postgres://postgres:postgres@localhost:5432/polling_indexer \
  --tx-function 0x6d2c743607ef275bd6934fe5c2a7e5179cca6fbd2049cfa79de2310b74f3cf83 \
  --start-checkpoint 96139267
# check database 
docker exec polling-indexer-pg psql "postgres://postgres:postgres@localhost:5432/polling_indexer" -c "SELECT checkpoint_seq, transaction_digest, sender, success, kind, raw_json FROM transactions ORDER BY checkpoint_seq DESC, id DESC LIMIT 20;"
# cleanup postgres container
docker rm -f polling-indexer-pg
```

Alex6323 (approved):

SeventhOdyssey71:
Any updates on this yet @Thoralf-M @Alex6323 @DaughterOfMars

.gitignore
thibault-martinez: Where is this folder? I don't see it. Is it generated at run time by the polling example?

Makefile
```diff
@@ -325,10 +325,17 @@ example:
 
 .PHONY: examples
 examples: ## Run all Rust examples
-	@for example in $$(find crates/iota-sdk/examples -name "*.rs" -not -path "*/release/*" -exec basename {} .rs \;); do \
+	@# NOTE: -maxdepth 1 -type f excludes multi-crate examples like polling-indexer
```
thibault-martinez: What do you mean by multi-crate?

.gitignore
SeventhOdyssey71: It's a folder I was using to work in my local, I removed it, but forgot to remove from .gitignore, I'd fix that now.

Makefile
SeventhOdyssey71: yeah, that's misleading. This makes more sense right?  '@# NOTE: -maxdepth 1 -type f excludes package-based examples like polling-indexer'

Makefile
```diff
@@ -325,10 +325,17 @@ example:
...
 examples: ## Run all Rust examples
-	@for example in $$(find crates/iota-sdk/examples -name "*.rs" -not -path "*/release/*" -exec basename {} .rs \;); do \
+	@# NOTE: -maxdepth 1 -type f excludes package-based examples like polling-indexer
+	@# that require external services (e.g. PostgreSQL). Run those separately.
+	@for example in $$(find crates/iota-sdk/examples -maxdepth 1 -type f -name "*.rs" -exec basename {} .rs \;); do \
 		$(MAKE) example "$$example" || exit $$?; \
 	done
 
+.PHONY: polling-indexer
+polling-indexer: ## Run the polling-indexer example (requires PostgreSQL)
```
thibault-martinez: You don't reference this rule in the README nor the CI. I would either start using it or remove it.

crates/iota-sdk/examples/polling-indexer/src/cli.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    #[arg(
+        long,
+        help = "PostgreSQL connection URL (default: postgres://localhost:5432/polling_indexer)"
+    )]
+    pub db_url: Option<String>,
+
+    #[arg(
+        long,
+        alias = "progress-file",
```
thibault-martinez: I find this alias a bit confusing. Is it a key or a file?

crates/iota-sdk/examples/polling-indexer/src/config.rs
```diff
@@ -0,0 +1,224 @@
...
+    network: Option<Network>,
+    /// Explicit GraphQL URL override.
+    /// Example: "https://graphql.testnet.iota.cafe"
+    graphql_url: Option<String>,
+    /// PostgreSQL connection string.
+    /// Example: "postgres://postgres:postgres@localhost:5432/polling_indexer"
+    db_url: Option<String>,
+    /// Progress row key in the indexer_progress table.
+    /// Example: "polling-indexer"
+    #[serde(alias = "progress_file")]
```
thibault-martinez: same question

crates/iota-sdk/examples/polling-indexer/src/config.rs
```diff
@@ -0,0 +1,224 @@
...
+            (None, Some(network)) => network,
+            (None, None) => Network::Testnet,
+        };
+
+        let page_size = value
+            .page_size
+            .or(file_config.page_size)
+            .unwrap_or(DEFAULT_PAGE_SIZE);
+        if page_size <= 0 {
+            anyhow::bail!("page_size must be > 0");
```
thibault-martinez, line 150: Why is it a `i32` then?

SeventhOdyssey71: The SDK's `PaginationFilter::limit` is `Option<i32>`, so we match that type to avoid a cast. The `<= 0` check catches invalid values at config time.

thibault-martinez: Alright

thibault-martinez (approved):
Thank you! We will get in touch for the reward :)

PR #979, opened 2026-02-19
feat: API documentation generation for Python, Go, and Kotlin

Ginowine:
This PR adds a generate_docs.sh script to automate API documentation generation across the binding ecosystems.

Key Improvements:

- Python: Implemented a self-bootstrapping virtual environment and used direct FFI inspection to capture the full API reference (bypassing empty wrapper issues).
- Kotlin: Integrated Dokka GFM and added a workaround for Gradle 9.2 configuration cache compatibility.
- Go: Corrected pathing to target the iota_sdk source package for reliable go doc output.


Thoralf-M (commented):
How will this be used? Since right now the docs are nowhere added

generate_docs.sh
```diff
@@ -0,0 +1,43 @@
...
+echo "Extracting Python API reference..."
+python3 -c "import iota_sdk.iota_sdk_ffi; help(iota_sdk.iota_sdk_ffi)" > python_api.md
+
+deactivate
+cd ../..
+
+echo "Step 2: Kotlin Docs"
+cd bindings/kotlin
+# Disable config cache due to Dokka/Gradle 9.2 incompatibility
+./gradlew dokkaGfm --no-configuration-cache
```
Alex6323: I see that for python and go the docs are written to .md files. I guess, dokkaGfm produces `kotlin_api.md`?

Ginowine: By default, `dokkaGfm` produces a structured directory (to handle Kotlin's package/nested class hierarchy) with an index.md as the entry point. If you run this script, and go into the Kotlin/dokka/gfm/iota-sdk/iota_sdk, you will find different folders containing the needed .md files.

Ginowine:
> How will this be used? Since right now the docs are nowhere added

The script just help to extracts the docs directly from the source code.

I do not know if we should call the script from upload-docs.yml to generate the .md files, which will then be sent to an AWS S3 bucket, from where it can then be pulled into our docs website.

@Dr-Electron tried explaining the pipeline, but I will need more to understand the next steps, and what I should do next.

generate_docs.sh
```diff
@@ -0,0 +1,43 @@
...
+echo "Installing/Updating build tools..."
+pip install --upgrade pip
+pip install maturin
+
+# Build native bindings so symbols are discoverable
+cd src && maturin develop && cd ..
+
+# Target the FFI layer specifically for full API coverage
+echo "Extracting Python API reference..."
+python3 -c "import iota_sdk.iota_sdk_ffi; help(iota_sdk.iota_sdk_ffi)" > python_api.md
```
Dr-Electron: Not sure if it makes a difference but in the past we used pydoc-markdown.
Any reason to go for this solution now?

Dr-Electron:
Yeah I would add this to an action. Maybe for now with a workflow dispatch and an input on which language we want to push?

One we know that this works we can push on releases like we did in the past 👍

Dr-Electron:
> How will this be used? Since right now the docs are nowhere added

We are currently working on a PR in the docs 👍

generate_docs.sh
Ginowine: Oh no @Dr-Electron, no specific reason. I just used it based on my research; I didn't know PyDoc was the preferred one.

Thoralf-M:
We got swift and csharp bindings in the meantime :trollface: 
Want to add them in this PR?

Dr-Electron:
Added a upload action:

```
act release -e release-event.json --secret-file release-secrets -W .github/workflows/upload-docs.yml -P ubuntu-latest=catthehacker/ubuntu:full-latest
```


You don't need the secrets if you don't want to upload docs, so you can ignore it or remove it.
For release event use one of the release tags in the following format
`release-event.json`
```
{
  "action": "published",
  "ref": "refs/tag/iota-go-sdk-v0.0.1-alpha.1",
  "release": {
    "tag_name": "iota-go-sdk-v0.0.1-alpha.1"
  }
}
```

Dr-Electron:
The biggest issue is documentation generation tho as for Kotlin and Go we just generate way to big single files 😓 
Not sure if anyone knows better tools

PR #983, opened 2026-02-23
feat: update `ConsensusCommitPrologueV1`

Alex6323:
Key changes:
* Added a custom JsonSchema impl that keeps backwards compatibility with `openrpc.json` spec in the monorepo
* Replaced the `serde` derives with a custom impl that is distinguishes between JSON and BCS

### DIFF
--- a/crates/iota-sdk-types/src/transaction/mod.rs
+++ b/crates/iota-sdk-types/src/transaction/mod.rs
@@ -710,7 +710,7 @@ impl Ord for ActiveJwk {
 #[cfg_attr(
     feature = "schemars",
     derive(schemars::JsonSchema),
-    schemars(tag = "kind", rename_all = "snake_case")
+    schemars(tag = "kind", rename_all = "camelCase")
 )]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
 #[non_exhaustive]
--- a/crates/iota-sdk-types/src/transaction/mod.rs
+++ b/crates/iota-sdk-types/src/transaction/mod.rs
@@ -707,17 +707,12 @@ impl Ord for ActiveJwk {
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-#[cfg_attr(
-    feature = "schemars",
-    derive(schemars::JsonSchema),
-)]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
 #[non_exhaustive]
 pub enum ConsensusDeterminedVersionAssignments {
     /// Cancelled transaction version assignment.
     CancelledTransactions {
         #[cfg_attr(feature = "proptest", any(proptest::collection::size_range(0..=2).lift()))]
-        #[cfg_attr(feature = "schemars", schemars(rename = "CancelledTransactions"))]
         cancelled_transactions: Vec<CancelledTransaction>,
     },
 }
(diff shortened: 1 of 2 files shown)


Thoralf-M (approved):

DaughterOfMars (approved):

DaughterOfMars (approved):

thibault-martinez:
CI failing

Alex6323 (commented):
> CI failing 

on it

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -760,26 +749,26 @@ pub struct CancelledTransaction {
...
+/// The BCS serialized form for this type is defined by the
+/// following ABNF:
 ///
 /// ```text
 /// version-assignment = object-id u64
 /// ```
-#[derive(Clone, Debug, PartialEq, Eq)]
-#[cfg_attr(
-    feature = "serde",
-    derive(serde::Serialize, serde::Deserialize),
```
Thoralf-M, line 771: Why can't we derive serde anymore?

Alex6323: Because serialization is different for human readable serialization (JSON) and binary serialization (BCS).

Thoralf-M: How did it work before?

Alex6323: custom serde as well (for main type that holds this). Remember, this type we added. The schemas weren't updated I think, otherwise we would have caught it earlier. This PR fixes it though and makes sure that `openrpc.json` and `iota.yaml` specs are generated in a non-breaking way against `develop`.

crates/iota-sdk-types/src/transaction/serialization.rs
```diff
@@ -309,36 +309,35 @@ mod end_of_epoch {
...
+        CancelledTransactions(Vec<CancelledTransaction>),
     }
 
     #[derive(serde::Serialize)]
+    #[serde(rename = "ConsensusDeterminedVersionAssignments")]
     enum BinaryConsensusDeterminedVersionAssignmentsRef<'a> {
-        CancelledTransactions {
-            cancelled_transactions: &'a Vec<CancelledTransaction>,
-        },
+        CancelledTransactions(&'a Vec<CancelledTransaction>),
```
thibault-martinez, line 334: I'm just curious but such a change was not really needed, right?

Alex6323: Well, the generated `iota.yaml` was different than the one currently on `develop` without this change, which tbf only means that the tracer saw a difference. Probably doesn't really change it on the wire though (asked Claude about it), but still, since this type is just there for serialization I decided to change it such that the generated spec stays exactly the same.

Alex6323: `iota.yaml` on `develop`:

```
ConsensusDeterminedVersionAssignments:
  ENUM:
    0:
      CancelledTransactions:
        NEWTYPE:
          SEQ:
            TUPLE:
              - TYPENAME: TransactionDigest
              - SEQ:
                  TUPLE:
                    - TYPENAME: ObjectID
                    - TYPENAME: SequenceNumber
```

so it expects it to be a `NEWTYPE` wrapper instead of a `STRUCT`

Alex6323: but yeah, I guess no difference in practice

thibault-martinez: Yeah I guess any sort of wrapping doesn't really influence the BCS in the end, was just wondering if it was motivated by something else

thibault-martinez (approved):

PR #997, opened 2026-03-02
chore(examples+tests): switch to testnet

Alex6323:
Switches all tests and examples to use addresses, package/object IDs, transaction digests and serialized transactions from the testnet to make them robust against devnet wipes.


bindings/go/examples/dynamic_fields/main.go
```diff
@@ -13,7 +13,7 @@ import (
 func main() {
 	client := iota_sdk.GraphQlClientNewDevnet()
```
thibault-martinez: We're still using devnet in most examples?

Alex6323: only did the most annoying parts yesterday ... on it!

bindings/go/examples/release/main.go
```diff
@@ -7,11 +7,11 @@ import (
 	"fmt"
 	"log"
 
-	"github.com/iotaledger/iota-sdk-go"
+	"github.com/iotaledger/iota-rust-sdk/bindings/go/iota_sdk"
```
thibault-martinez: Please revert

Alex6323: I am not even sure why this happend tbh

crates/iota-sdk-types/src/iota_names/config.rs
thibault-martinez: These probably need to change as well

bindings/go/examples/move_view_call/main.go
```diff
@@ -5,140 +5,8 @@ package main
...
-	}
-
-	if auctionJsonResult.Error != nil {
-		fmt.Println("Auction JSON Error:", *auctionJsonResult.Error)
-	} else if auctionJsonResult.Results != nil {
-		fmt.Println("Auction JSON Results:", *auctionJsonResult.Results)
-	} else {
-		fmt.Println("No auction JSON results")
-	}
+	fmt.Println("Temporarily disabled until Move View calls are supported on testnet.")
```
thibault-martinez: Can we not keep the code commented so we don't have to write it again later?

Alex6323: If we comment the code in Rust then the formatter will add linebreaks in between IDs and strings, so uncommenting again leaves an invalid format requiring us to manually fix this. Can't really tell for the bindings if that's an issue there as well for some. Renaming just seemed the simplest solution to me except for Go.

bindings/go/examples/release/main.go
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/997/commits/3a44a966766f620514e2f5bfd9790bb8df6ee154

bindings/go/examples/move_view_call/main.go
thibault-martinez: I'd still keep it at least for go, fixing format manually is less annoying than writing it again or trying to find it in the git history?

crates/iota-sdk-types/src/iota_names/config.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/997/commits/028bd8417d263690f5102f75e11cd884843e02ce

bindings/go/examples/move_view_call/main.go
Alex6323: it's still there as `main.go.disabled`.

thibault-martinez: Or maybe that's fine to keep only this example on Devnet for now

thibault-martinez: > it's still there as main.go.disabled.

Oh my bad, didn't see that

thibault-martinez (approved):

bindings/go/examples/move_view_call/main.go
Thoralf-M: I would also keep some example with devnet as long as it's not available in testnet

Alex6323: the Rust one?

Alex6323: Not really sure it's worth teaching something that's not officially supported yet and might still change.

DaughterOfMars (approved):

bindings/go/examples/move_view_call/main.go
thibault-martinez: I mean, it is officially supported, just not on all networks. By having people trying it early we may be able to find issues before it becomes an actual problem.

Alex6323: Okay. Should I re-enable the Rust one or all of them?

Thoralf-M: Yes, someone from the community actually found already something that was not aligned, which we then changed
How about just having the blake2b256 part for devnet now and the iota names part commented? And then once it's available on testnet, we update it

Thoralf-M: Should be the same for all languages

Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/997/commits/9b83c1ce06757e7d83ccaf475a0b39cccbfc2452

Thoralf-M (approved):

PR #1005, opened 2026-03-06
feat(types): Skip validation of identifiers in `MoveCall` for BCS

DaughterOfMars:
## Description

Skips validation of Identifiers in BCS bytes for `MoveCall` commands. This is necessary because these fields were not previously validated and is invalid in some failed transactions on the network.

### DIFF
--- a/crates/iota-sdk-types/src/transaction/serialization.rs
+++ b/crates/iota-sdk-types/src/transaction/serialization.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_with::{DeserializeAs, SerializeAs};
 
 use super::Argument;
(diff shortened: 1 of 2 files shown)


crates/iota-sdk-types/src/transaction/serialization.rs
```diff
@@ -1054,6 +1054,22 @@ mod transaction_expiration {
     }
 }
 
+/// Deserialize an `Identifier` without validating that it is a valid Move
+/// identifier. This is used for deserializing type tags in `MoveCall` commands,
```
thibault-martinez: Not really type tags then?

thibault-martinez (approved):

Thoralf-M (approved):

PR #1020, opened 2026-03-10
feat(bindings): add wasm

Thoralf-M:
## Summary

Adds WASM/TypeScript bindings for the IOTA SDK, generated from `iota-sdk-ffi` via UniFFI + `uniffi-bindgen-react-native`, compiled to `wasm32-unknown-unknown` and bundled for the browser. Built with a single `make wasm`. Runs in Node too — same wasm binary, no second build target.

## What's included

- **New `iota-sdk-wasm` crate** + `bindings/wasm/` package: build pipeline (`ubrn` → `wasm-bindgen` → esbuild), browser entrypoint (`initAsync`), and `chain_id` / `staking` browser examples.
- **`async-compat` patch** (`patches/async-compat/`, via `[patch.crates-io]`) so `uniffi_core` works on wasm32 (no `Instant::now` / thread usage).
- **wasm-friendly `sleep` / `timeout`** in `iota-sdk-graphql-client::wait` — tokio's time driver isn't available on `wasm32-unknown-unknown`, so the faucet's `request_and_wait*` and the client's `wait_for_tx` would trap with `RuntimeError: unreachable`. On wasm32 we drive a cancellable `setTimeout`-backed future instead (with `clearTimeout` on drop so Node's event loop doesn't stay alive past an aborted future).
- **wasm-specific feature gating** in `iota-sdk-ffi` (uniffi `wasm-unstable-single-threaded`, per-target `tokio`/`getrandom`, `?Send` async-trait on wasm32).
- **CI**: `wasm.yml` runs the `make wasm32` compile check for the SDK crates on every PR; the full bindings build lives in `bindings.yml`'s `wasm` job.
- **Full Node-runnable example set** mirroring `crates/iota-sdk/examples/*.rs` — `.mjs` scripts under `bindings/wasm/examples/` that `import { … } from "@iota/sdk-wasm"` (resolved to the local package via pnpm workspace) and run against the same bundle that ships to npm. All 43 examples pass against the localnet + gas station started by `run_localnet.sh` (the same setup `bindings.yml`'s `wasm` job uses). Browser HTML examples (`chain_id.html`, `staking.html`) are kept.
- **`bindings.yml` gets a `wasm` job** mirroring `python`/`swift`: build, format-check, start localnet, run the examples, and a final `make is-dirty`.
- **npm release flow**, aligned with the other bindings:
  - `wasm_publish.yml` — split `build` (runs `make wasm`, uploads artifact) and `publish` (`npm publish --ignore-scripts --provenance` on the artifact) jobs. Publishing authenticates via **npm OIDC trusted publishing** (no `NPM_TOKEN`): only the `publish` job is granted `id-token: write`, so a compromised cargo/pnpm script in `build` can't reach any publish credential. The public repo also gets build provenance attached to each tarball.
  - `release.yml` maps `iota-wasm-sdk` → `bindings/wasm`.
  - `pre_publish.yml` supports `iota-wasm-sdk` (bump `package.json` from the latest tag, changelog, open the `release/iota-wasm-sdk/<version>` PR).
- **TypeScript types shipped in the npm package**: declarations emitted to `dist/types/` (`tsc -p tsconfig.types.json`, `noCheck` since the bindings TS is generated), with `package.json` `types` pointing at them. `package.json` made publishable (dropped `private`, added `files`/`main`/`module`/`exports`/`publishConfig`).

## Not done yet / follow-ups

- **Repurposing the existing `@iota/sdk-wasm` package.** The npm name `@iota/sdk-wasm` already exists (the legacy TS-SDK wasm bindings from `iotaledger/iota-sdk`, latest `1.1.3` / `alpha: 2.0.0-beta.1`). The IOTA org owns it and is **intentionally repurposing it**: this PR publishes the new Rust-SDK (uniffi) bindings as `@iota/sdk-wasm@3.0.0-alpha.1`, continuing the version line past `2.x` (the old `1.x`/`2.x` line is no longer needed). This is a hard implementation/API swap under the same name — downstreams on `1.x`/`2.x` upgrading to `3.x` get an entirely different SDK.
- **Remaining DevOps steps to enable publishing** (the workflow side is done — `wasm_publish.yml` uses OIDC trusted publishing + `--provenance`; only the npm-side setup is left):
  1. On npmjs.com, as an owner of `@iota/sdk-wasm`: **Settings → Trusted Publisher → GitHub Actions** with organization/repository `iotaledger/iota-rust-sdk` and workflow filename `wasm_publish.yml` (leave "Environment" blank — the publish job uses none).
  2. Confirm the package's publishing settings permit automation via trusted publishing (trusted publishing satisfies the "require 2FA to publish" setting for CI).
  3. Validate end-to-end with a dry run: **Actions → "Publish WASM SDK to npm" → Run workflow** with `test_publish = true` (exercises the OIDC exchange + packaging without publishing).
  4. First real publish: cut a release tagged `iota-wasm-sdk-v3.0.0-alpha.1` (or run the workflow with `test_publish = false`). Optionally adjust `dist-tags` / deprecate the legacy `1.x`/`2.x` versions afterwards.

## Test plan

- [x] `wasm.yml` (`make wasm32` compile check), `cargo-sort`, `dprint` are green in CI.
- [x] `bindings.yml`'s `wasm` job (build + format-check + localnet + run all `.mjs` examples + `make is-dirty`) is green.
- [x] `make wasm` produces `dist/iota-sdk.js`, `dist/node.js`, `dist/index_bg.wasm`, and `dist/types/index.web.d.ts`.
- [x] `make wasm-example chain_id` runs locally against testnet.
- [x] `make wasm-serve` still serves the HTML examples in a browser.
- [ ] After the npm Trusted Publisher is configured: run `wasm_publish.yml` via `workflow_dispatch` (dry-run) to validate the OIDC publish path.


bindings/wasm/examples/json_query.mjs
thibault-martinez, line 1: Any reason this doesn't have the `RPC_VALIDATOR_FIELDS` fragment like the other bindings?

Thoralf-M: Good catch — no reason, it was just an incomplete port. The whole `validatorSet` selection had been trimmed (it was also missing `activeValidators`/`committeeMembers` and the `*Id` fields). Aligned it with `crates/iota-sdk/examples/json_query.rs` and the other five bindings in 66dc3e3.

Approved by @Thoralf-M :trollface:

---

bindings/wasm/src/ts/index.web.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Load and initialise the WASM module.
+ *
+ * @param input - One of:
+ *   - `string` or `URL`: location of the `.wasm` binary to fetch (browser).
+ *   - `ArrayBuffer` or `Uint8Array`: raw `.wasm` bytes (Node.js, where
+ *     `fetch('file://…')` isn't reliably supported — pass
+ *     `fs.readFileSync(...)` instead).
+ *   - omitted: defaults to `index_bg.wasm` resolved relative to this JS module.
+ */
+export async function uniffiInitAsync(
```
thibault-martinez: Can we find an alternative name to this? I'm not a big fan that it mentions uniffi, it's a bit unrelated to it

Thoralf-M: Good call — renamed `uniffiInitAsync` → `initAsync` in 853ac39. The `uniffi` mention is gone from the public surface; the change covers the entrypoint, the Node entry, all examples, the browser HTML examples, and the README (the implementation comment about UniFFI's `env` stubs stays, since that's accurate internals).

---

bindings/wasm/README.md
```diff
@@ -0,0 +1,129 @@
...
+> **Why a separate `iota-sdk-wasm` crate?** The generated bridge uses `extern "C" { fn uniffi_iota_sdk_ffi_fn_*... }` to reference UniFFI symbols. These must come from an external crate; they cannot be placed inside `iota-sdk-ffi` itself.
+
+## Prerequisites
+
+```bash
+# WASM target
+rustup target add wasm32-unknown-unknown
+
+# wasm-bindgen CLI (version must match the compiled-in library)
+cargo install wasm-bindgen-cli --version 0.2.114
```
Alex6323: use same version as in the CI workflows? There's a slight patch version descrepancy here (0.2.114 vs 0.2.122)

crates/iota-sdk-ffi/Cargo.toml
```diff
@@ -44,8 +44,17 @@ rand.workspace = true
...
 
 # internal dependencies
 iota-sdk = { version = "3.0.0-alpha.1", path = "../iota-sdk", default-features = false, features = ["graphql", "txn-builder", "types", "serde", "rand", "hash"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+tokio = { workspace = true, features = ["time"] }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+# Required to make rand/getrandom work in WASM environments
+getrandom = { version = "0.2", features = ["js"] }
```
DaughterOfMars: Not `get_random_2`?

DaughterOfMars: Also do we even need this now that it's in iota-types?

crates/iota-sdk-graphql-client/src/api/transactions.rs
```diff
@@ -281,12 +281,11 @@ impl Client {
...
         timeout: impl Into<Option<Duration>>,
     ) -> Result<()> {
-        tokio::time::timeout(
+        crate::wait::timeout(
             timeout.into().unwrap_or_else(|| Duration::from_secs(60)),
             async {
-                let mut interval = tokio::time::interval(tokio::time::Duration::from_millis(100));
                 loop {
-                    interval.tick().await;
+                    crate::wait::sleep(Duration::from_millis(100)).await;
```
DaughterOfMars: This is not the same functionality

patches/async-compat/src/lib.rs
DaughterOfMars, line 1: Mind explaining this?

bindings/wasm/iota-sdk-wasm/src/lib.rs
thibault-martinez, line 1: Is there no way we can move this to the wasm crate instead? Or make it a temporary thing while building?

bindings/wasm/README.md
Thoralf-M:
```suggestion
cargo install wasm-bindgen-cli --version 0.2.122
```

Thoralf-M: Updated after README.md creation, updated the README now https://github.com/iotaledger/iota-rust-sdk/pull/1020/changes/6c8053b0b2f852c29c976520a6f674518c9ba80d

bindings/wasm/src/ts/index.web.ts
thibault-martinez: I'm being annoying but do you think `initAsync` is clear enough about what it does?

crates/iota-sdk-graphql-client/src/api/transactions.rs
Thoralf-M: Yes but is it an issue? I can change it to only be like this on wasm target if you think it should stay .tick() otherwise

bindings/wasm/iota-sdk-wasm/src/lib.rs
Thoralf-M: What do you mean, it's in the wasm crate?

patches/async-compat/src/lib.rs
Thoralf-M: It's required so we can compile to wasm
**`async-compat` patch** (`patches/async-compat/`, via `[patch.crates-io]`) so `uniffi_core` works on wasm32 (no `Instant::now` / thread usage).

bindings/wasm/iota-sdk-wasm/src/lib.rs
thibault-martinez: I meant to the wasm binding folder sorry. This feels a bit weird/wrong to have this here idk

bindings/wasm/src/ts/index.web.ts
Thoralf-M: I think for people working with wasm it's clear, it's a common required thing that you need to init the wasm blob by loading it async

bindings/wasm/iota-sdk-wasm/src/lib.rs
Thoralf-M: Ah should be possible, just thought since it's a Rust crate it should be here, uniffi crate could also be moved into bindings folder

thibault-martinez: I think it's fine that the FFI crate lives in the workspace, it's more generic and at least we immediately see if we break it. This new one is more of a build artefact for one specific binding. Plus it leaves some things in git stage, we should probably fix that as well.

Thoralf-M: Moved it https://github.com/iotaledger/iota-rust-sdk/pull/1020/changes/15d017bc2843dab3836505b24a03d9ae910338a7

thibault-martinez:
Please add a wasm example for https://github.com/iotaledger/iota-rust-sdk/pull/1172

crates/iota-sdk-ffi/Cargo.toml
Thoralf-M: You're right, not needed anymore after https://github.com/iotaledger/iota-rust-sdk/pull/1193, removed it

crates/iota-sdk-graphql-client/src/api/transactions.rs
DaughterOfMars: tick doesn't wait on the first pass so it's adding a sleep for no reason

patches/async-compat/src/lib.rs
DaughterOfMars: But like, how does it work?

crates/iota-sdk-graphql-client/src/api/transactions.rs
Thoralf-M: Good point — fixed in f5a2b32: check first, then sleep, so the first poll is immediate (no upfront 100ms).

---

patches/async-compat/src/lib.rs
Thoralf-M: From what I understand it's replacing async-compat interface so it doesn't panic and we don't need any special logic here for this project so it's fine with the dummy interface

crates/iota-sdk-graphql-client/src/api/transactions.rs
DaughterOfMars: But now it always sleeps afterward...
We want the tick functionality, where it doesn't sleep on the first iteration only

Thoralf-M: It will break early without sleep if a condition is true

patches/async-compat/src/lib.rs
DaughterOfMars: it would be nice if we understood this better

crates/iota-sdk-graphql-client/src/api/transactions.rs
DaughterOfMars: I don't think it's the same still. Is there really no way to use interval in wasm? Or some crate that could do this?

patches/async-compat/src/lib.rs
Thoralf-M: What part exactly do you want to understand better? Without it, it panicked at runtime

bindings/wasm/src/ts/index.web.ts
thibault-martinez: Ok

patches/async-compat/src/lib.rs
DaughterOfMars: I would like to know why this is needed and not some other solution. It's not easy to see exactly what this is doing

scripts/cargo_sort/run_consolidate.sh
```diff
@@ -8,4 +8,6 @@ $PYTHON_CMD cargo_sort.py --consolidate-deps \
   --strict \
   --strict-ignore "getrandom" \
   --strict-ignore "tracing:crates/iota-sdk/examples/polling-indexer" \
+  --strict-ignore "tokio:patches/async-compat" \
```
thibault-martinez: where is tokio: coming from?

crates/iota-sdk-graphql-client/src/faucet.rs
```diff
@@ -187,10 +187,9 @@ impl FaucetClient {
         let request_id = self.request(address).await?;
 
         if let Some(request_id) = request_id {
-            let status_response = tokio::time::timeout(FAUCET_REQUEST_TIMEOUT, async {
-                let mut interval = tokio::time::interval(FAUCET_POLL_INTERVAL);
+            let status_response = crate::wait::timeout(FAUCET_REQUEST_TIMEOUT, async {
                 loop {
-                    interval.tick().await;
+                    crate::wait::sleep(FAUCET_POLL_INTERVAL).await;
```
thibault-martinez: This is now waiting first?

crates/iota-sdk-graphql-client/src/api/transactions.rs
Thoralf-M:
```
What's the same: Timing of the first check. tokio::time::interval's first .tick() completes immediately, so the old code (tick at loop top) checked right away — exactly like the new code, which checks before sleeping. No change there.

What differs: fixed-rate vs. fixed-delay scheduling.

interval.tick() aims at a fixed rate: ticks at t=0, 100ms, 200ms, ... regardless of how long each check takes. The 100ms period includes the GraphQL round-trip. If a check takes 60ms, the next one starts 40ms later.
sleep(100ms) at the loop bottom is a fixed delay: each iteration is check-duration + 100ms. With a 60ms round-trip, you poll every ~160ms instead of every 100ms.
The divergence is biggest when the check is slow: if a round-trip exceeds 100ms, interval's default MissedTickBehavior::Burst fires the missed ticks immediately, so you'd poll back-to-back with effectively zero gap. The sleep version always guarantees a 100ms breather between requests — which for hammering a GraphQL endpoint in a wait loop is arguably the more polite behavior. Correctness is unaffected either way; the outer crate::wait::timeout bounds the whole thing identically.
```
tbh to me the sleep sounds more like what we would want here

Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 resolver = "2"
 members = [
+  "bindings/wasm/iota-sdk-wasm",
```
thibault-martinez, line 4: I feel like we don't necessarily have to have this in the workspace (I'd prefer if it was not)

crates/iota-sdk-graphql-client/src/faucet.rs
Thoralf-M: Good catch — it was. Fixed in 4436cc9: poll first, sleep at the end of the loop (same shape as `wait_for_tx`), so it no longer waits a full `FAUCET_POLL_INTERVAL` before the first poll. The `InProgress` arm now falls through to the sleep instead of `continue`-ing — otherwise `continue` would have skipped the bottom sleep and busy-looped.

---

patches/async-compat/src/lib.rs
Thoralf-M: # Why the `async-compat` patch is needed

Experiment: removed the `[patch.crates-io] async-compat = { path = "patches/async-compat" }`
entry from the workspace `Cargo.toml`, rebuilt, and compared behavior. Date: 2026-06-10,
branch `feat/wasm`, uniffi 0.29.5, upstream async-compat 0.2.5.

## What exactly fails without the patch

**Nothing fails at build time.** Both
`cargo check -p iota-sdk-ffi --target wasm32-unknown-unknown` and the full
`make wasm` succeed with upstream `async-compat 0.2.5`.

**Every async FFI call fails at runtime.** Any example that calls an async
function traps immediately:

```
$ node chain_id.mjs
RuntimeError: unreachable
    at GraphQlClient.chainId (dist/iota-sdk.js:49589:37)
```

With a panic hook installed (minimal repro crate that awaits
`async_compat::Compat::new(async { 42 })`, exactly what uniffi generates), the
underlying panic is:

```
panicked at async-compat-0.2.5/src/lib.rs:464:10:
called `Result::unwrap()` on an `Err` value:
Error { kind: Unsupported, message: "operation not supported on this platform" }
```

### Failure chain

1. Every `#[uniffi::export(async_runtime = "tokio")]` makes uniffi wrap the
   exported future in `::uniffi::deps::async_compat::Compat::new(fut)` —
   unconditionally, on all targets
   (`uniffi_macros-0.29.5/src/export/scaffolding.rs:283`). uniffi has no
   per-target switch for this.
2. On first poll, upstream `Compat::poll` calls `get_runtime_handle()`: no
   ambient tokio runtime exists, so it initializes the lazy global `TOKIO1`,
   which builds a current-thread tokio runtime and parks it on a background
   thread via `thread::Builder::spawn(...).unwrap()` (upstream `lib.rs:464`).
3. `std::thread::spawn` returns `Err(Unsupported)` on `wasm32-unknown-unknown`
   (no threads), the `.unwrap()` panics, and with `panic = "abort"` the wasm
   module traps with `RuntimeError: unreachable`.

So the very first poll of *any* async export — `chainId()`, faucet,
transaction builder, all of them — kills the call before it does any work,
with an opaque error for users.

## How the patch fixes it

`patches/async-compat/` is target-split:

- **Native targets** (`src/native.rs`): mirrors upstream's approach — lazily
  create a shared tokio runtime parked on a helper thread and enter its
  context on each poll. The Go/Python/Kotlin/Swift/C# bindings need this:
  uniffi drives futures from its own (non-tokio) foreign executor threads, and
  the SDK internals (reqwest/hyper via the tokio reactor, `tokio::time`)
  require an ambient runtime context.
- **wasm32** (`src/lib.rs`): `Compat::poll` just forwards to the inner future
  (transparent passthrough).

## Why passthrough is safe on wasm32

`Compat`'s *only* job is "enter a tokio runtime context while polling". On
wasm32 there is no tokio runtime to enter and nothing that needs one:

- futures are driven by the JS event loop via `wasm-bindgen-futures`,
- reqwest uses the Fetch API (no tokio reactor),
- the graphql client's `sleep`/`timeout` are `setTimeout`-backed on wasm32
  (also part of this PR).

The context-entering side effect the wrapper exists for has no counterpart on
wasm, so skipping it changes nothing — it just stops the wrapper from trying
(and failing) to spawn a thread for a runtime nobody polls into.

Verification with the patch in place: `make wasm` + `node chain_id.mjs` →
`Chain ID: 2304aa97` (testnet).

## Why a patch and not some other solution

- uniffi hard-codes the `Compat` wrap for `async_runtime = "tokio"`. The only
  source-level alternative is dual `#[cfg_attr(...)]` attributes on every
  exported async fn/impl block (dozens of sites, easy to forget on each new
  export).
- Dropping the `tokio` uniffi feature on wasm doesn't work: the
  macro-generated `async_compat::Compat` path then no longer exists →
  compile error.
- A `[patch.crates-io]` fixes it once, centrally, and its scope is
  self-documenting: it only changes which `async-compat` source is compiled,
  and that source is ~160 lines readable in full.

## Accuracy note

The patched native implementation caches the runtime handle once at first use,
while upstream re-checks `tokio::runtime::Handle::try_current()` on every
poll. For uniffi's usage — polls always come from non-tokio foreign executor
threads — this behaves identically, but it is not a byte-for-byte upstream
copy: it "mirrors upstream's approach" (the unused I/O trait adapters are also
omitted; uniffi only uses `Compat::new` on futures).

bindings/wasm/iota-sdk-wasm/src/lib.rs
thibault-martinez, line 1: Depending if we can dissociate this from the workspace, I'd call this folder just `rust`

scripts/cargo_sort/run_consolidate.sh
Thoralf-M: The consolidation check (`--strict`) fails when a crate
declares a dependency whose version spec diverges from
`[workspace.dependencies]`. The patch crate declares
`tokio = { version = "1", features = ["rt"] }` directly (workspace pins
`tokio = "1.40.0"`) and cannot use `workspace = true` because it is
deliberately not a workspace member — it is a standalone crate consumed only
via `[patch.crates-io]`. Verified: without the ignore rule the script fails
with

```
Package 'tokio' has conflicting versions:
    WORKSPACE: 1.40.0
    patches/async-compat: (1)
❌ STRICT MODE: Version conflicts detected ... Exiting with error.
```

bindings/wasm/scripts/fix-generated-typescript.mjs
thibault-martinez, line 1: copyright header?

patches/async-compat/src/lib.rs
DaughterOfMars: > The only source-level alternative is dual #[cfg_attr(...)] attributes on every
exported async fn/impl block (dozens of sites, easy to forget on each new
export).

Not sure this is that bad actually.

crates/iota-sdk-graphql-client/src/api/transactions.rs
DaughterOfMars: Alright, I suppose that's a fair argument

Cargo.toml
Thoralf-M: Why not? Shares the target folder if it is

bindings/wasm/scripts/fix-generated-typescript.mjs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1020/commits/d3cec773e2c8b5bb6c753bd292153250a5d9536a

bindings/wasm/scripts/fix-generated-typescript.mjs
```diff
@@ -0,0 +1,69 @@
...
+//   ) {
+//     throw new UniffiInternalError.ApiChecksumMismatch(
+//       "uniffi_..._...",
+//     );
+//   }
+// so the match must span newlines ([\s\S]) rather than assume one line.
+fixed = fixed.replace(
+  /\s*if \(\s*nativeModule\(\)\.ubrn_uniffi_[\s\S]*?checksum[\s\S]*?ApiChecksumMismatch\([\s\S]*?\);\s*\}/g,
+  ''
+);
```
thibault-martinez: Is this still necessary? I think I gated `Address::random` out of wasm32

thibault-martinez: Not sure if there was more than just this

thibault-martinez: I guess it's fine to keep anyway

Thoralf-M: Feature gating things out of wasm32 creates the mismatch, so it's still needed

thibault-martinez: Yeah I guess I meant I gated it in then

thibault-martinez: Idk, I just undid what you initially did to Address::random

Makefile
```diff
@@ -39,14 +39,39 @@ package_%.json: crates/integration-tests/%/Move.toml crates/integration-tests/%/
 test-with-localnet: package_test_example_v1.json package_test_example_v2.json ## Run tests with localnet
 	cargo nextest run -p iota-sdk-graphql-client -p integration-tests
 
-.PHONY: wasm
-wasm: ## Build WASM modules
+# Verify that individual SDK crates compile to wasm32-unknown-unknown.
+# This is a quick compatibility check, not the full WASM bindings build.
+.PHONY: wasm-check
```
thibault-martinez: 1. I'd call it wasm32
2. Need to update AGENTS.md because it now builds bindings with `make wasm`

bindings/wasm/scripts/fix-generated-typescript.mjs
Thoralf-M: Correct — since `Address::random` was un-gated in 06a7900, no exported symbol is missing from the wasm binary anymore, so the env-stub checksum mismatch this worked around is gone. Removed the strip in a83c4589e; checksum validation is active again in the generated bindings (verified: `initAsync()` passes all checksum checks, examples run).

patches/async-compat/src/lib.rs
Thoralf-M: This would add quite some noise

Makefile
Thoralf-M: 3. update workflow
https://github.com/iotaledger/iota-rust-sdk/pull/1020/commits/9b88f91b670d3d720c3b186e7e171a0c76eb4cb8

patches/async-compat/src/lib.rs
DaughterOfMars: How many places are there actually in the uniffi crate though?

.github/workflows/wasm.yml
```diff
@@ -30,12 +30,28 @@ jobs:
...
+          cargo install wasm-bindgen-cli --version 0.2.122
+          npm install -g pnpm@10
+
+      - name: Build WASM bindings
+        # `ubrn build` compiles its own bindgen tool from source on first run;
+        # don't impose our `-Dwarnings` on that third-party build. The wasm crate
+        # itself is built with the rustflags from ubrn.config.yaml.
+        run: make wasm
+        env:
+          RUSTFLAGS: ""
```
thibault-martinez: Not sure we should do this here, we're already doing it in the bindings workflow

bindings/wasm/src/ts/index.web.ts
thibault-martinez, line 1: If not generated, add copyright

.github/workflows/wasm.yml
Thoralf-M: Removed in 22c51a16f — wasm.yml now only runs the cheap `make wasm32` compile check on every PR; the full bindings build stays in the path-filtered bindings.yml wasm job, which also took over the `make is-dirty` check that verified the build leaves a clean tree.

patches/async-compat/src/lib.rs
Thoralf-M: 65 currently and each would need
```
#[cfg_attr(not(target_arch = "wasm32"), uniffi::export(async_runtime = "tokio"))]
#[cfg_attr(target_arch = "wasm32", uniffi::export)]
```

bindings/wasm/src/ts/index.web.ts
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1020/changes/44747558afc61b11a567e3ffd64519c46dbde9b9

.github/workflows/bindings.yml
```diff
@@ -144,3 +144,29 @@ jobs:
...
+      - name: Install clang
+        run: sudo apt-get install -y clang
+      - name: Install WASM bindings prerequisites
+        run: |
+          rustup target add wasm32-unknown-unknown
+          cargo install wasm-bindgen-cli --version 0.2.122
+          npm install -g pnpm@10
+      - name: Build the bindings
+        run: make wasm
+      - run: make is-dirty
```
thibault-martinez: Close if you think the probability is very low but this should generally be the last check, in case the other steps are leaving artefacts

thibault-martinez: Claude says that make wasm runs pnpm without --frozen-lockfile and that could leave changes making is-dirty fail

patches/async-compat/src/lib.rs
DaughterOfMars: That's still only 65 added lines

.github/workflows/wasm_publish.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Setup Node
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22
+          registry-url: https://registry.npmjs.org
+
+      - name: Ensure OIDC-capable npm
+        # Trusted publishing needs npm >= 11.5.1, newer than the version bundled
+        # with the runner's Node.
+        run: npm install -g npm@latest
```
thibault-martinez, line 93: npm-ception

patches/async-compat/src/lib.rs
DaughterOfMars: Compared to 160 difficult-to-understand lines and a new crate, I think 65 lines of easy to understand tags is a pretty good option

.github/workflows/wasm_publish.yml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: Set package version from release tag
+        if: github.event_name == 'release'
+        working-directory: package
+        run: |
+          version=$(echo "${{ github.event.release.tag_name }}" | sed -nr 's|iota-wasm-sdk-v(.*)|\1|p')
+          echo "Publishing version $version"
+          npm version "$version" --no-git-tag-version --allow-same-version
```
thibault-martinez, line 104: Where is this setting it exactly?

Makefile
```diff
@@ -39,14 +39,39 @@ package_%.json: crates/integration-tests/%/Move.toml crates/integration-tests/%/
...
 	$(MAKE) -C crates/iota-sdk-types wasm
 
+# Build the full WASM bindings package for browsers.
+# Uses ubrn (uniffi-bindgen-react-native) to generate TS bindings, compile
+# to wasm32, and run wasm-bindgen. Then esbuild bundles into dist/.
+.PHONY: wasm
+wasm: ## Build WASM bindings for browsers
+	cd bindings/wasm && pnpm install && npx ubrn build web --config ubrn.config.yaml --profile wasm-release
+	@# Optionally run wasm-opt on the wasm-bindgen output for size reduction.
+	@if command -v wasm-opt >/dev/null 2>&1; then \
```
thibault-martinez, line 60: Does that mean if we passed --wasm-opt to make wasm?

Makefile
```diff
@@ -149,6 +178,11 @@ endef
 # are already PascalCase, so both `make csharp-example chain_id` and
 # `make csharp-example ChainId` resolve to the same example.
 snake_to_pascal = $(shell printf '%s' "$(1)" | awk -F_ '{ s=""; for (i=1; i<=NF; i++) s = s toupper(substr($$i,1,1)) substr($$i,2); print s }')
+# Build the FFI crate (release) and detect the library extension.
+define build_binding
+cargo build -p iota-sdk-ffi --lib --release; \
+$(detect_lib_ext)
+endef
```
thibault-martinez: socratic interlocutor: why?

.github/workflows/wasm_publish.yml
Thoralf-M: In package.json

Makefile
Thoralf-M: Modified the comment, it will run if wasm-opt is available, otherwise just skip it

Makefile
Thoralf-M: Leftover from intermediate changes, reverted

.github/workflows/bindings.yml
Thoralf-M: Right, changed to use --frozen-lockfile

Thoralf-M: Moved

Thoralf-M: Had to move it back because localnet run creates changes, but it's after formatting now

.github/workflows/wasm_publish.yml
thibault-martinez: Isn't it already changed/set by the pre_publish workflow?

Cargo.toml
thibault-martinez: I don't like it, but I guess it's fine (for now)

crates/iota-sdk-graphql-client/src/wait.rs
```diff
@@ -0,0 +1,128 @@
...
+impl std::fmt::Display for Elapsed {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("deadline elapsed")
+    }
+}
+
+impl std::error::Error for Elapsed {}
+
+#[cfg(target_arch = "wasm32")]
+mod wasm_time {
```
thibault-martinez, line 51: There really is not available code for this?

.github/workflows/wasm_publish.yml
Thoralf-M: In the normal flow yes, claude added it because someone could create a release branch without pre-publish and then it would diverge, but I guess since we don't plan to do that we can also remove
Want me to clean it up?

crates/iota-sdk-graphql-client/src/wait.rs
Thoralf-M: There is, but would add more dependencies

.github/workflows/wasm_publish.yml
thibault-martinez: I guess it doesn't hurt to keep it, was just wondering if this was a mistake, resolving

crates/iota-sdk-graphql-client/src/wait.rs
thibault-martinez: Alright

patches/async-compat/src/lib.rs
Thoralf-M: Changed it https://github.com/iotaledger/iota-rust-sdk/pull/1020/commits/c2ff950aeb85a956683abc1159d49ea135d364db

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1022, opened 2026-03-10
fix: release go with v prefixed tag

thibault-martinez:
As per https://go.dev/ref/mod#versions

> Each version starts with the letter v, followed by a semantic version

We are currently not https://github.com/iotaledger/iota-sdk-go/tags

### DIFF
--- a/.github/workflows/go_publish.yml
+++ b/.github/workflows/go_publish.yml
@@ -134,7 +134,7 @@ jobs:
           git config --global user.email github-actions@github.com
           git config --global user.name github-actions
           git add .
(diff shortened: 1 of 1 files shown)


.github/workflows/go_publish.yml
```diff
@@ -135,5 +136,5 @@ jobs:
           git add .
           git commit -m "Update Go SDK to version ${{ steps.set-version.outputs.version }}"
```
thibault-martinez:
```suggestion
          git commit -m "Update Go SDK to version v${{ steps.set-version.outputs.version }}"
```

DaughterOfMars (approved):

Thoralf-M (approved):

PR #1025, opened 2026-03-10
feat(examples): run faster

Thoralf-M:
[run-ci]


Thoralf-M:
Most time is taken for compilation, I guess it's not really worth it to have the examples outputs mixed up for the little speedup

PR #1030, opened 2026-03-16
feat(workflows): add localnet test and fix localnet on linux

Thoralf-M:
Was broken for me


thibault-martinez:
I'm confused, don't we already have tests that require a localnet?

Thoralf-M:
> I'm confused, don't we already have tests that require a localnet?

Yes, but not using the script (it's used only for coverage)

.github/workflows/localnet.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Start local network
+        run: ./run_localnet.sh start ./iota
+
+      - name: Verify IOTA node is responding
+        run: |
+          curl -sf -X POST http://localhost:9000 \
+            -H 'Content-Type: application/json' \
+            -d '{"jsonrpc":"2.0","id":1,"method":"iota_getLatestCheckpointSequenceNumber"}' \
+            | jq .
```
Alex6323, line 58: Isn't that checked already now in the actual script?

Thoralf-M: No, the script just directly requests funds from the faucet

Alex6323 (approved):

DaughterOfMars (approved):

run_localnet.sh
```diff
@@ -35,45 +36,64 @@ if [ "$1" == "start" ]; then
...
+        echo "Last 20 lines of $IOTA_LOG:"
+        tail -20 "$IOTA_LOG" 2>/dev/null || echo "(no log file)"
+        exit 1
     fi
-    echo "Updating fullnode-url to use host IP: $HOST_IP"
-    sed -i.bak "s|http://localhost:9000|http://$HOST_IP:9000|g" "$CONFIG_PATH" && rm "$CONFIG_PATH.bak"
 
-    echo "Waiting for network to start and requesting faucet coins..."
+    echo "Setting up client environment..."
+    $IOTA_BINARY client -y new-env --alias localnet --rpc http://127.0.0.1:9000 --graphql http://0.0.0.0:9125 --faucet http://127.0.0.1:9123/v1/gas
```
Alex6323: that's a side effect I am not sure every user appreciates tbh

Alex6323: maybe the script should ask for permission to make this change, or just fail if that env doesn't exist. Also if `localnet` env already exists, will those values be overwritten? Or maybe the script should differentiate between CI execution (should add the env), and local execution (should fail or ask)

Thoralf-M: Yes, but what do you suggest to do instead?

Alex6323: once solution suggested by Claude would be to just make a direct curl request to the faucet

Thoralf-M: That's actually the best since the client cmds are split into a new binary and we don't need it for something else https://github.com/iotaledger/iota-rust-sdk/pull/1030/commits/fa41f79bbdcab445bc6839244d313238913aa56d

Alex6323 (approved):

.github/workflows/localnet.yml
thibault-martinez, line 1: If we start using the script in the CI at some point, I'm not sure this workflow is really necessary anymore

thibault-martinez (approved):

PR #1034, opened 2026-03-19
feat: MoveStruct improvements

thibault-martinez:


crates/iota-sdk-types/src/object.rs
```diff
@@ -217,26 +214,427 @@ pub struct MoveStruct {
...
+    pub fn is_type(&self, s: &StructTag) -> bool {
+        self.type_.is(s)
+    }
+
+    pub fn id(&self) -> ObjectId {
+        Self::id_opt(&self.contents).unwrap()
+    }
+
+    // TODO unsure about this
+    pub fn id_opt(contents: &[u8]) -> Result<ObjectId, AddressParseError> {
```
Alex6323: hmm, yeah. seems a bit weird. Like this ... it's just an associated function, not a method, but maybe for a good reason?

crates/iota-sdk-types/src/object.rs
```diff
@@ -217,26 +214,427 @@ pub struct MoveStruct {
...
+        ObjectId::from_bytes(contents)
+    }
+
+    /// Return the `value: u64` field of a `Coin<T>` type.
+    /// Useful for reading the coin without deserializing the object into a Move
+    /// value. It is the caller's responsibility to check that `self` is a coin.
+    /// This function may panic or do something unexpected otherwise.
+    pub fn get_coin_value_unsafe(&self) -> u64 {
```
Alex6323: maybe we should prefer `_unchecked` over `_unsafe`, or if we really want it called like this, we might as well just use the `unsafe` keyword?

crates/iota-sdk-types/src/object.rs
thibault-martinez: Sir, this is a draft :trollface:

crates/iota-sdk-types/src/object.rs
thibault-martinez: I'm fine with unchecked

thibault-martinez: Done

crates/iota-sdk-types/src/object.rs
```diff
@@ -281,6 +282,141 @@ pub struct MoveStruct {
     pub contents: Vec<u8>,
 }
 
+impl MoveStruct {
+    pub fn type_(&self) -> &MoveObjectType {
```
DaughterOfMars: Any reason we shouldn't rename this field/method to object_type?

crates/iota-sdk-types/src/object.rs
DaughterOfMars: Not a draft anymore 👀

crates/iota-sdk-types/src/object.rs
```diff
@@ -281,6 +282,141 @@ pub struct MoveStruct {
...
+            });
+        }
+        ObjectId::from_bytes(&contents[0..ObjectId::LENGTH])
+    }
+
+    /// Return the `value: u64` field of a `Coin<T>` type.
+    /// Useful for reading the coin without deserializing the object into a Move
+    /// value. It is the caller's responsibility to check that `self` is a coin.
+    /// This function may panic or do something unexpected otherwise.
+    pub fn get_coin_value_unchecked(&self) -> u64 {
```
DaughterOfMars: Where is this used in the core? Is it justified to add it here?

crates/iota-sdk-types/src/object.rs
```diff
@@ -281,6 +282,141 @@ pub struct MoveStruct {
...
+    #[cfg(feature = "serde")]
+    pub fn to_rust<'de, T: serde::Deserialize<'de>>(&'de self) -> Option<T> {
+        bcs::from_bytes(self.contents()).ok()
+    }
+
+    /// Approximate size of the object in bytes. This is used for gas metering.
+    /// For the type tag field, we serialize it on the spot to get the accurate
+    /// size. This should not be very expensive since the type tag is
+    /// usually simple, and we only do this once per object being mutated.
+    pub fn object_size_for_gas_metering(&self) -> usize {
```
DaughterOfMars: This feels like it should be a trait in the core.

crates/iota-sdk-types/src/object.rs
thibault-martinez: done

crates/iota-sdk-types/src/object.rs
thibault-martinez: moved

crates/iota-sdk-types/src/object.rs
thibault-martinez: moved

crates/iota-sdk-types/src/object.rs
```diff
@@ -281,6 +282,91 @@ pub struct MoveStruct {
...
+        self.object_type.is_coin()
+    }
+
+    pub fn is_staked_iota(&self) -> bool {
+        self.object_type.is_staked_iota()
+    }
+
+    pub fn is_clock(&self) -> bool {
+        self.object_type.is_clock()
+    }
```
DaughterOfMars: Any reason we're not just adding all of these?

crates/iota-sdk-types/src/object.rs
```diff
@@ -281,6 +282,91 @@ pub struct MoveStruct {
...
+
+    pub fn version(&self) -> Version {
+        self.version
+    }
+
+    /// Sets the version of this object to a new value which is assumed to be
+    /// higher (and checked to be higher in debug).
+    pub fn increment_version_to(&mut self, next: Version) {
```
DaughterOfMars: I don't really see the value of these functions here

crates/iota-sdk-types/src/object.rs
```diff
@@ -281,6 +282,91 @@ pub struct MoveStruct {
...
+
+    pub fn into_object_type(self) -> MoveObjectType {
+        self.object_type
+    }
+
+    pub fn type_tag(&self) -> TypeTag {
+        TypeTag::Struct(Box::new(self.object_type().clone().into()))
+    }
+
+    pub fn into_inner(self) -> (MoveObjectType, Vec<u8>) {
```
DaughterOfMars: This is weird. Should it be `into_parts` and return everything?

crates/iota-sdk-types/src/object.rs
thibault-martinez: Seem generic enough to me

crates/iota-sdk-types/src/object.rs
thibault-martinez: What do you mean?

crates/iota-sdk-types/src/object.rs
DaughterOfMars: I don't think so. It suits a specific purpose which only makes sense to the core repo IMO

crates/iota-sdk-types/src/object.rs
DaughterOfMars: All the `is_x()` fns from type_tag

crates/iota-sdk-types/src/object.rs
thibault-martinez: How is it specific? It's literally just a setter

crates/iota-sdk-types/src/object.rs
thibault-martinez: Will change

crates/iota-sdk-types/src/object.rs
thibault-martinez: You want me to add all type_tag methods to this type?

crates/iota-sdk-types/src/object.rs
DaughterOfMars: No, if it was a setter then it would be `set_version` and not check whether it's incrementing

crates/iota-sdk-types/src/object.rs
DaughterOfMars: I mean, why not?

crates/iota-sdk-types/src/object.rs
thibault-martinez: Oh come one this is bad faith, nothing wrong with a setter that checks the parameter, this is very common

crates/iota-sdk-types/src/object.rs
thibault-martinez: That sounds like a lot for not much, I'd rather just remove these 3 and force people to first get the object type if you prefer that

crates/iota-sdk-types/src/object.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1034/commits/c0b675180a4eb71a0147cbb004ff57a22259abc0

crates/iota-sdk-types/src/object.rs
Alex6323: The draft looked at me with sad, big eyes tearing up, I couldn't resist.

crates/iota-sdk-types/src/object.rs
```diff
@@ -281,6 +282,91 @@ pub struct MoveStruct {
     pub contents: Vec<u8>,
 }
 
+impl MoveStruct {
+    pub fn object_type(&self) -> &MoveObjectType {
+        &self.object_type
+    }
+
+    pub fn is_type(&self, s: &StructTag) -> bool {
```
Thoralf-M: Why don't return StructTag in object_type()? Or accept MoveObjectType in is_type()? Right now seems inconsistent

crates/iota-sdk-types/src/object.rs
thibault-martinez: Keeping it for now, will depend on https://github.com/iotaledger/iota-rust-sdk/issues/1043

crates/iota-sdk-types/src/object.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1034/commits/a836f00e5316794521ef55efc3af101fe68459a4

crates/iota-sdk-types/src/object.rs
Thoralf-M: I think I also prefer getting the inner type unless it's called like this in many places

crates/iota-sdk-types/src/object.rs
```diff
@@ -281,6 +283,88 @@ pub struct MoveStruct {
...
+
+    pub fn is_struct_tag(&self, s: &StructTag) -> bool {
+        &self.object_type == s
+    }
+
+    pub fn id(&self) -> ObjectId {
+        Self::id_opt(&self.contents).unwrap()
+    }
+
+    pub fn id_opt(contents: &[u8]) -> Option<ObjectId> {
```
DaughterOfMars, line 308: Can this at least be a method on `&self`?

crates/iota-sdk-types/src/object.rs
DaughterOfMars: Sure

crates/iota-sdk-types/src/object.rs
thibault-martinez: No, it's used without an instance in the monorepo

crates/iota-sdk-types/src/object.rs
thibault-martinez: Will do

thibault-martinez: Removed them

Thoralf-M (approved):

crates/iota-sdk-types/src/object.rs
DaughterOfMars: I don't think it's common at all. And definitely not normal to panic in that situation.

thibault-martinez: I would like to keep them here

thibault-martinez: It's just a debug assert anyway, in release build this is just setting a value, I'm not going to put a setter in an extension trait

crates/iota-sdk-types/src/object.rs
```diff
@@ -281,6 +283,76 @@ pub struct MoveStruct {
     pub contents: Vec<u8>,
 }
 
+impl MoveStruct {
```
Alex6323, line 286: can you claude some docs on these public methods?

Alex6323 (approved):

crates/iota-sdk-types/src/object.rs
DaughterOfMars: I would like a second opinion

crates/iota-sdk-types/src/object.rs
DaughterOfMars: Then we should have a helper fn in there or something, because it doesn't make sense to have a free standing fn that is totally unrelated to this type but which is defined on it

thibault-martinez: I agree but I opened an issue to deal with this later after discussing the whole topic with VM Lang

crates/iota-sdk-types/src/object.rs
Thoralf-M: I'd say the logic is not specific to the iota repo, because that the version is a lamport timestamp and should therefore only increase. However I'm not sure why it's just a debug assert then since this seems critical

DaughterOfMars: Then why is there a decrement method?

crates/iota-sdk-types/src/object.rs
thibault-martinez: Done

crates/iota-sdk-types/src/object.rs
Thoralf-M: It's only used once in genesis builder, might be good to remove this one and just set the version manually there

Alex6323: hmm, I am fine with having these methods tbh, but this is now in the SDK!!, and therefore should never just panic, even in debug, but instead return an error, so the consumer can decide on their end how to handle the situation. I know I already approved (but I was wrong) :trollface:

Alex6323: An alternative to returning an error would be to make a smaller value a no-op, and clearly mentioning it in the docs of this method: "Values smaller then the current version will be discarded and not change the version!", or something similar. But that is not without pitfalls either and could cause hard to find bugs.

thibault-martinez: Moved back to monorepo

PR #1037, opened 2026-03-20
feat(iota-sdk-types): add MoveObjectType

Thoralf-M:
https://github.com/iotaledger/iota/issues/10639

### DIFF
--- a/bindings/swift/Sources/CIotaSDK/CIotaSDK.h
+++ b/bindings/swift/Sources/CIotaSDK/CIotaSDK.h
@@ -275,6 +275,24 @@ void*_Nonnull uniffi_iota_sdk_ffi_fn_clone_address(void*_Nonnull ptr, RustCallSt
 void uniffi_iota_sdk_ffi_fn_free_address(void*_Nonnull ptr, RustCallStatus *_Nonnull out_status
 );
 #endif
+#ifndef UNIFFI_FFIDEF_UNIFFI_IOTA_SDK_FFI_FN_CONSTRUCTOR_ADDRESS_AUTHENTICATOR_STATE
+#define UNIFFI_FFIDEF_UNIFFI_IOTA_SDK_FFI_FN_CONSTRUCTOR_ADDRESS_AUTHENTICATOR_STATE
+void*_Nonnull uniffi_iota_sdk_ffi_fn_constructor_address_authenticator_state(RustCallStatus *_Nonnull out_status
+    
+);
+#endif
+#ifndef UNIFFI_FFIDEF_UNIFFI_IOTA_SDK_FFI_FN_CONSTRUCTOR_ADDRESS_CLOCK
+#define UNIFFI_FFIDEF_UNIFFI_IOTA_SDK_FFI_FN_CONSTRUCTOR_ADDRESS_CLOCK
+void*_Nonnull uniffi_iota_sdk_ffi_fn_constructor_address_clock(RustCallStatus *_Nonnull out_status
+    
+);
+#endif
+#ifndef UNIFFI_FFIDEF_UNIFFI_IOTA_SDK_FFI_FN_CONSTRUCTOR_ADDRESS_DENY_LIST
+#define UNIFFI_FFIDEF_UNIFFI_IOTA_SDK_FFI_FN_CONSTRUCTOR_ADDRESS_DENY_LIST
+void*_Nonnull uniffi_iota_sdk_ffi_fn_constructor_address_deny_list(RustCallStatus *_Nonnull out_status
+    
+);
+#endif
 #ifndef UNIFFI_FFIDEF_UNIFFI_IOTA_SDK_FFI_FN_CONSTRUCTOR_ADDRESS_FRAMEWORK
 #define UNIFFI_FFIDEF_UNIFFI_IOTA_SDK_FFI_FN_CONSTRUCTOR_ADDRESS_FRAMEWORK
 void*_Nonnull uniffi_iota_sdk_ffi_fn_constructor_address_framework(RustCallStatus *_Nonnull out_status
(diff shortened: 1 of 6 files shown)


crates/iota-sdk-types/src/move_core/struct_tag.rs
```diff
@@ -199,6 +199,15 @@ impl StructTag {
         (self.address, self.module, self.name, self.type_params)
     }
 
+    /// Checks if this is a `TimeLock<Balance<T>>` for some T
```
Alex6323: The comment feels a bit over-specific? Maybe "Checks if this `StructTag` represents a timelocked coin balance" works as well and is less "codish"?

crates/iota-sdk-types/src/object.rs
```diff
@@ -170,6 +170,74 @@ impl ObjectData {
...
+}
+
+impl From<StructTag> for MoveObjectType {
+    fn from(tag: StructTag) -> Self {
+        Self(tag)
+    }
+}
+
+impl From<MoveObjectType> for StructTag {
+    fn from(mot: MoveObjectType) -> Self {
```
Alex6323: insignificant, but I'ld prefer `obj` or `type_`

crates/iota-sdk-types/src/move_core/struct_tag.rs
Thoralf-M: I think having the Timelock<Balance<T>> here is good, as one could have the same logic also implemented with another type. But I changed the words before https://github.com/iotaledger/iota-rust-sdk/pull/1037/commits/0adf3e3325f6cf1cb7e942509ac6e0c07ccf29f7

crates/iota-sdk-types/src/object.rs
Thoralf-M: Renamed https://github.com/iotaledger/iota-rust-sdk/pull/1037/commits/0adf3e3325f6cf1cb7e942509ac6e0c07ccf29f7

Alex6323 (approved):

bindings/swift/Sources/CIotaSDK/CIotaSDK.h
thibault-martinez, line 1: How come swift is changing?

crates/iota-sdk-types/src/move_core/struct_tag.rs
```diff
@@ -390,6 +400,18 @@ impl StructTag {
...
+    /// (`0x2::timelock::TimeLock<0x2::balance::Balance<0x2::iota::IOTA>>`)
+    pub fn new_timelocked_gas_balance() -> Self {
+        Self::new_time_lock(Self::new_balance(Self::new_gas()))
+    }
+
+    /// Checks if this is a timelocked IOTA balance type
+    /// (`0x2::timelock::TimeLock<0x2::balance::Balance<0x2::iota::IOTA>>`)
+    pub fn is_timelocked_gas_balance(&self) -> bool {
+        *self == Self::new_timelocked_gas_balance()
+    }
```
thibault-martinez: Can you put these 2 in the first impl block please? Preferably close to `is_timelocked_balance`

bindings/swift/Sources/CIotaSDK/CIotaSDK.h
Thoralf-M: Not sure, maybe wasn't generated before, but CI failed because there was a diff

crates/iota-sdk-types/src/move_core/struct_tag.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1037/commits/b6ab79d92ee411266d3733a28b0312df469014a8

DaughterOfMars (approved):

crates/iota-sdk-types/src/move_core/struct_tag.rs
```diff
@@ -199,6 +199,27 @@ impl StructTag {
         (self.address, self.module, self.name, self.type_params)
     }
 
+    /// Checks if this is a timelocked coin balance `TimeLock<Balance<T>>`
+    pub fn is_timelocked_balance(&self) -> bool {
+        self.is_time_lock()
```
thibault-martinez, line 204: This feels a bit inconsistent...

crates/iota-sdk-types/src/object.rs
```diff
@@ -528,6 +592,7 @@ mod serialization {
     /// specialized variants, e.g. `Other(GasCoin::type_())` instead of
     /// `GasCoin`
     #[derive(serde::Deserialize)]
+    #[serde(rename = "MoveObjectType")]
     enum MoveStructType {
```
thibault-martinez: Is there a reason it's not called like that to start with? Just feels like a mistake

crates/iota-sdk-types/src/move_core/struct_tag.rs
DaughterOfMars: Do you prefer `is_time_locked_balance` or `is_timelock` in the SDK? I lean toward the latter.

thibault-martinez: What I prefer does not really matter, we are broadly using `TimeLock/time_lock`

crates/iota-sdk-types/src/object.rs
Thoralf-M: I think both are valid and it's only used internally in the serialization so I don't care what we use

crates/iota-sdk-types/src/move_core/struct_tag.rs
Thoralf-M: Timelock/timelock is more common

thibault-martinez: Actually I just saw that on-chain we have `TimeLock` but also `TimelockedStakedIota` :(

crates/iota-sdk-types/src/object.rs
thibault-martinez: I think I'd prefer if we renamed, we also have `MoveStruct` so this is a bit confusing

crates/iota-sdk-types/src/move_core/struct_tag.rs
thibault-martinez: Alright, whatever you prefer then

thibault-martinez (approved):

crates/iota-sdk-types/src/object.rs
Thoralf-M: Can't rename to MoveObjectType as this already is another type, fine with `MoveObjectTypeWrapper`?

thibault-martinez: Sure, or something more explicit about it being a space efficient version

Thoralf-M: That would be `MoveObjectType` already, so adding something about space efficient would be confusing I think, renamed to MoveObjectTypeWrapper, the comment for it should be clear enough if someone wonders about it

crates/iota-sdk-types/src/object.rs
```diff
@@ -592,8 +592,7 @@ mod serialization {
     /// specialized variants, e.g. `Other(GasCoin::type_())` instead of
     /// `GasCoin`
     #[derive(serde::Deserialize)]
-    #[serde(rename = "MoveObjectType")]
```
thibault-martinez: We still need this, no?

Thoralf-M: Yes good catch, removed it when I renamed to MoveObjectType just to notice that it doesn't work and forgot to revert

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1037/commits/f47e6f7a3d77ae5acb1f309301be29ad7f9c382c

PR #1040, opened 2026-03-25
feat: BCS ABNF compile-time generation

### LINKED ISSUE #10513: Rethink BCS compatibility check
It's currently achieved through https://github.com/iotaledger/iota/blob/develop/crates/iota-core/src/generate_format.rs which is a real pain. Very often when doing changes we don't actually break BCS compatibility but have to add custom tracing just because of type/field names and serde attributes details.
Maybe an approach through BNF generation could be investigated.

thibault-martinez:
## Summary

This PR introduces a machine-readable BCS schema for `iota-sdk-types`, expressed as an [ABNF](https://datatracker.ietf.org/doc/html/rfc5234) grammar file, along with the tooling to generate and validate it.

### What's included

- **`crates/iota-bcs-schema`** — a new proc-macro crate providing a `#[derive(BcsSchema)]` macro. When applied to a type, it emits a `bcs_schema_definition()` method that returns the type's ABNF grammar fragment. The macro handles structs, enums, and common generic wrappers (`Option<T>`, `Vec<T>`, `HashMap<K, V>`), and supports override attributes (`#[bcs_schema(name = "...", definition = "...")]` on types and `#[bcs_schema(as_type = "...")]` on fields/variants) for cases where the derived output needs manual tuning.

- **`crates/iota-sdk-types/bcs-schema.abnf`** — the auto-generated ABNF schema covering all SDK types (transactions, effects, checkpoints, crypto primitives, etc.). It is committed to the repo and regenerated on demand via `BCS_SCHEMA=1 cargo check -p iota-sdk-types --features bcs-schema`.

- **`crates/iota-sdk-types/build.rs`** — a build script that, when `BCS_SCHEMA=1` is set alongside the `bcs-schema` feature, deletes the old schema file and touches all `.rs` sources to force every `#[derive(BcsSchema)]` to re-run, producing a fresh schema from scratch.

- **Grammar-driven fuzz tests** (`tests/bcs_schema_fuzzing.rs`) — a test suite that parses `bcs-schema.abnf`, randomly generates conforming byte sequences for each rule, and asserts that the BCS deserializer accepts them. This validates that the grammar is a _sound_ description of the wire format.

- **CI** — a new lint workflow job that regenerates the schema and fails if the committed file is out of date.

### Motivation

Having a formal, human- and machine-readable description of the BCS wire format makes it easier to:
- build cross-language BCS parsers/validators without reimplementing the Rust types,
- catch accidental serialization-breaking changes at the schema level, and
- fuzz the deserializer against grammar-valid inputs.

### Usage

    # Regenerate the schema file
    BCS_SCHEMA=1 cargo check -p iota-sdk-types --features bcs-schema

    # Normal development — no impact
    cargo check --all-features

### Derive attributes

| Attribute | Level | Purpose |
|---|---|---|
| `#[bcs_schema(definition = "32OCTET")]` | Type | Override entire RHS (for `[u8; Self::LENGTH]`, etc.) |
| `#[bcs_schema(name = "custom-name")]` | Type | Override the ABNF rule name |
| `#[bcs_schema(as_type = "u64")]` | Field | Override schema type (for type aliases like `Version = u64`) |
| `#[bcs_schema(as_type = "call-arg")]` | Field | Reference a wire-type that differs from the Rust type |
| `#[bcs_schema(skip)]` | Field | Omit from schema |

### Known limitations

- `as_type` and `definition` values are trusted — no validation against the actual serde impl
- Custom `Serialize` impls are invisible to the macro; divergences must be annotated manually
- Serde attributes (`skip`, `rename`, `flatten`) are not read by the macro

## Test plan

- [x] `cargo check -p iota-sdk-types --all-features` — no regressions
- [x] `BCS_SCHEMA=1 cargo check --features bcs-schema` produces complete, sorted `bcs-schema.abnf`
- [x] Two consecutive `BCS_SCHEMA=1` builds produce byte-identical output
- [x] Removing `BcsSchema` from a field type causes a compile error
- [x] `cargo check` without the feature is unaffected (fully incremental)


crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -53,6 +55,7 @@ pub struct TransactionEffectsV1 {
     pub dependencies: Vec<Digest>,
     /// The version number of all the written Move objects by this transaction.
     #[cfg_attr(feature = "schemars", schemars(with = "crate::_schemars::U64"))]
+    #[cfg_attr(feature = "bcs-schema", bcs_schema(as_type = "u64"))]
```
thibault-martinez, line 58: Can we put that inside version sow e don't repeat it every time?

crates/iota-bcs-schema/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "iota-bcs-schema"
+version = "0.1.0"
+edition = "2021"
```
thibault-martinez: 2024

crates/iota-sdk-types/src/effects/v1.rs
DaughterOfMars: I forgot that `Version` is a typedef right now. But it will be a type eventually then we can.

crates/iota-sdk-types/src/crypto/bls12381.rs
```diff
@@ -22,6 +22,8 @@
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
+#[cfg_attr(feature = "bcs-schema", bcs_schema(definition = "%x60 96OCTET"))]
```
thibault-martinez: can these be merged? If yes, do it for all instances please

crates/iota-bcs-schema/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+// ---------------------------------------------------------------------------
+// File writing
+// ---------------------------------------------------------------------------
+
+fn schema_file_path() -> std::path::PathBuf {
+    if let Ok(p) = std::env::var("BCS_SCHEMA_FILE") {
+        return std::path::PathBuf::from(p);
+    }
+    let manifest = std::env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into());
+    std::path::PathBuf::from(manifest).join("bcs-schema.abnf")
```
Thoralf-M: Create consts?

crates/iota-sdk-bcs-schema/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+    let input = parse_macro_input!(input as DeriveInput);
+    match expand(&input) {
+        Ok(ts) => ts.into(),
+        Err(e) => e.to_compile_error().into(),
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Attribute parsing
+// ---------------------------------------------------------------------------
```
Alex6323, line 34: Do you think those headings are worth keeping?

crates/iota-sdk-bcs-schema/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+        && let Some(seg) = p.path.segments.last()
+    {
+        return seg.ident == name;
+    }
+    false
+}
+
+// ---------------------------------------------------------------------------
+// CamelCase → kebab-case
+// ---------------------------------------------------------------------------
```
Alex6323, line 262: this in particular feels a bit useless as there will always only be one function.

crates/iota-sdk-bcs-schema/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+// Tests
+// ---------------------------------------------------------------------------
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn kebab_case() {
```
Alex6323, line 583: is that the only interesting test?

crates/iota-sdk-bcs-schema/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "iota-bcs-schema"
+version = "0.0.0"
```
Alex6323, line 3: should be set to "0.0.1-alpha.1"?

thibault-martinez: No, it's not meant to be released

crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -281,14 +286,9 @@ impl ObjectIn {
...
-///             =/ object-out-object-write
-///             =/ object-out-package-write
-///
-///
-/// object-out-missing        = %x00
-/// object-out-object-write   = %x01 digest owner
-/// object-out-package-write  = %x02 version digest
+/// object-out = %x00             ; Missing
+///            / %x01 digest owner   ; ObjectWrite
+///            / %x02 u64 digest     ; PackageWrite
```
Thoralf-M: Not properly aligned

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -88,14 +91,15 @@ pub struct SignedTransaction {
...
+#[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
 #[non_exhaustive]
 pub enum TransactionExpiration {
     /// The transaction has no expiration
     #[default]
     None,
     /// Validators won't sign a transaction unless the expiration Epoch
     /// is greater than or equal to the current epoch
-    Epoch(EpochId),
+    Epoch(#[cfg_attr(feature = "bcs-schema", bcs_schema(as_type = "u64"))] EpochId),
```
Thoralf-M, line 102: Why is `as_type = "u64"` needed here if the type is u64?

crates/iota-sdk-types/src/type_tag/mod.rs
```diff
@@ -17,29 +17,17 @@ use super::Address;
...
+///          / %x01           ; U8
+///          / %x02           ; U64
+///          / %x03           ; U128
+///          / %x04           ; Address
+///          / %x05           ; Signer
+///          / %x06 type-tag  ; Vector
+///          / %x07 struct-tag ; Struct
+///          / %x08           ; U16
+///          / %x09           ; U32
+///          / %x0a           ; U256
```
Thoralf-M: Also not aligned

crates/iota-sdk-types/src/object.rs
```diff
@@ -23,12 +23,14 @@ pub type Version = u64;
...
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
 pub struct ObjectReference {
     /// The object id of this object.
     pub object_id: ObjectId,
     /// The version of this object.
     #[cfg_attr(feature = "serde", serde(with = "crate::_serde::ReadableDisplay"))]
     #[cfg_attr(feature = "schemars", schemars(with = "crate::_schemars::U64"))]
+    #[cfg_attr(feature = "bcs-schema", bcs_schema(as_type = "u64"))]
     pub version: Version,
```
Thoralf-M, line 34: Why not derive BcsSchema for Version?

Thoralf-M (commented):
We should add a README.md in iota-bcs-schema/ which explains what this is good for and how to use it

crates/iota-sdk-bcs-schema/src/lib.rs
DaughterOfMars: It's kind of nice for reading 🤷

crates/iota-sdk-bcs-schema/src/lib.rs
DaughterOfMars: I can't think of others

crates/iota-sdk-types/src/object.rs
DaughterOfMars: We can't because it's a typedef

crates/iota-sdk-types/src/transaction/mod.rs
DaughterOfMars: Because the derive doesn't know it's a u64 and this happens at compile time

crates/iota-sdk-types/src/object.rs
Thoralf-M: No, it's `pub struct Version(u64)`?

crates/iota-sdk-types/bcs-schema.abnf
```diff
@@ -0,0 +1,467 @@
...
+          (vector object-id)   ; dependencies
+          object-id            ; package
+          argument             ; ticket
+
+upgrade-info = object-id   ; upgraded-id
+               u64         ; upgraded-version
+
+user-signature = bytes
+
+user-signature = bytes
```
thibault-martinez: duplicate

crates/iota-sdk-types/bcs-schema.abnf
```diff
@@ -0,0 +1,467 @@
...
+                      / %x01 u64   ; MutateDeleted
+                      / %x02 u64   ; ReadDeleted
+                      / %x03 u64   ; Cancelled
+                      / %x04   ; PerEpochConfig
+
+unchanged-shared-object = object-id               ; object-id
+                          unchanged-shared-kind   ; kind
+
+unchanged-shared-object = object-id               ; object-id
+                          unchanged-shared-kind   ; kind
```
thibault-martinez, line 424: duplicate

crates/iota-sdk-types/bcs-schema.abnf
```diff
@@ -0,0 +1,467 @@
...
+
+bls12381-signature = 48OCTET
+
+call-arg = %x00 bytes   ; Pure
+         / %x01 object-arg   ; Object
+
+cancelled-transaction = digest                        ; digest
+                        (vector version-assignment)   ; version-assignments
+
+change-epoch = u64                       ; epoch
```
Thoralf-M: Correct that there is no `-v1` here?

crates/iota-sdk-types/bcs-schema.abnf
```diff
@@ -0,0 +1,467 @@
...
+
+jwk = string   ; kty
+      string   ; e
+      string   ; n
+      string   ; alg
+
+jwk-id = string   ; iss
+         string   ; kid
+
+make-move-vector = (option type-tag)   ; type-
```
Thoralf-M: `type-` is there something missing?

crates/iota-sdk-types/bcs-schema.abnf
DaughterOfMars: weird

crates/iota-sdk-types/bcs-schema.abnf
DaughterOfMars: Yes, annoyingly

crates/iota-sdk-types/bcs-schema.abnf
DaughterOfMars: It's because of the field name `type_` which we currently cannot rename.

crates/iota-sdk-types/src/object.rs
DaughterOfMars: Not on develop

Thoralf-M: Oh right, something to remember to update later then

crates/iota-sdk-types/src/crypto/bls12381.rs
```diff
@@ -22,6 +22,11 @@
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[cfg_attr(
+    feature = "bcs-schema",
+    derive(iota_bcs_schema::BcsSchema),
+    bcs_schema(definition = "%x60 96OCTET")
```
thibault-martinez: Can you check if "OCTET" is the standard?

thibault-martinez: Can we have it in decimal?

crates/iota-sdk-types/src/crypto/signature.rs
```diff
@@ -309,6 +309,11 @@ impl std::fmt::Display for InvalidSignatureScheme {
 /// the completely serialized signature.
 #[derive(Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[cfg_attr(
+    feature = "bcs-schema",
+    derive(iota_bcs_schema::BcsSchema),
+    bcs_schema(definition = "bytes")
```
thibault-martinez, line 306: What is `bytes`? Is it defined in the ABNF? length-prefixed?

crates/iota-sdk-types/src/type_tag/mod.rs
```diff
@@ -222,6 +210,11 @@ impl std::error::Error for TypeParseError {}
 /// ```
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[cfg_attr(
+    feature = "bcs-schema",
+    derive(iota_bcs_schema::BcsSchema),
+    bcs_schema(definition = "string")
```
thibault-martinez, line 216: Properly define string in the ABNF?

crates/iota-sdk-types/src/type_tag/serialization.rs
```diff
@@ -25,15 +25,27 @@ impl<'de> Deserialize<'de> for Identifier {
...
+/// BCS wire-format variant indices for [`TypeTag`].
+///
+/// The variant order here matches the on-chain BCS encoding, which differs from
+/// the Rust [`TypeTag`] enum declaration order. This type is also used to
+/// generate the ABNF schema via [`BcsSchema`](iota_bcs_schema::BcsSchema).
 #[repr(u32)]
+#[cfg_attr(
+    feature = "bcs-schema",
+    derive(iota_bcs_schema::BcsSchema),
+    bcs_schema(name = "type-tag")
```
thibault-martinez, line 37: What would happen if I created a different type with the same BCS schema name? Can we make sure it fails?

crates/iota-sdk-types/src/address.rs
```diff
@@ -58,6 +58,11 @@
 #[derive(Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[cfg_attr(
+    feature = "bcs-schema",
+    derive(iota_bcs_schema::BcsSchema),
+    bcs_schema(definition = "32OCTET")
```
thibault-martinez, line 64: Can we put a space? I always read it as 320 :(

crates/iota-sdk-types/src/object.rs
```diff
@@ -897,6 +913,8 @@ mod serialization {
     }
 
     #[derive(serde::Serialize, serde::Deserialize)]
+    #[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
+    #[cfg_attr(feature = "bcs-schema", bcs_schema(name = "genesis-object"))]
```
thibault-martinez: Combine

crates/iota-sdk-types/src/crypto/signature.rs
DaughterOfMars: There are now definitions added to the file for these basic types

crates/iota-sdk-types/src/type_tag/mod.rs
DaughterOfMars: done

crates/iota-sdk-types/src/type_tag/serialization.rs
DaughterOfMars: There is now duplicate detection

crates/iota-sdk-types/src/address.rs
DaughterOfMars: no :c

DaughterOfMars: I could add something like `byte = OCTET` so it would look like `32byte` instead?

crates/iota-sdk-bcs-schema/Cargo.toml
Alex6323: ah, okay

crates/iota-sdk-bcs-schema/src/lib.rs
```diff
@@ -0,0 +1,582 @@
...
+                        format!("(%x00 / %x01 {rhs})")
+                    }
+                    None => "(%x00 / %x01 unknown)".into(),
+                },
+                "Box" => match extract_single_generic(seg) {
+                    Some(inner) => type_to_schema(&inner),
+                    None => "unknown".into(),
+                },
+                "BTreeMap" | "HashMap" => match extract_two_generics(seg) {
+                    // BCS map: `size` length prefix followed by sorted key-value pairs.
```
Thoralf-M: Only BTreeMap is sorted?

crates/iota-sdk-bcs-schema/src/lib.rs
```diff
@@ -0,0 +1,582 @@
...
+    let mut output =
+        String::from("; Auto-generated BCS schema definitions\n; Do not edit manually\n");
+    for (_, def) in &entries {
+        output.push('\n');
+        output.push_str(def);
+        output.push('\n');
+    }
+
+    // Best-effort write — don't break compilation if it fails
+    let _ = std::fs::write(&path, output);
```
Thoralf-M, line 518: Do we really want that behavior?

DaughterOfMars: I think so

Thoralf-M (approved):

Alex6323 (approved):

crates/iota-sdk-types/bcs-schema-diff.md
Thoralf-M, line 1: Remove?

thibault-martinez: done

DaughterOfMars (approved):

PR #1053, opened 2026-04-01
chore: utils module containing write_sep

thibault-martinez:
New utils module starting with write_sep


crates/iota-sdk-types/src/utils.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+pub fn write_sep<T: core::fmt::Display>(
```
Thoralf-M, line 26: We should have some comment for what this is good for

Thoralf-M (approved):

PR #1061, opened 2026-04-07
feat(scripts): add cargo sort script

muXxer:
This PR ports and adapts the `cargo_sort.py` script from the monorepo.

PR #1062, opened 2026-04-07
feat(gRPC): add grpc client, types and proto-build

muXxer:
# Description of change

This PR moves the `iota-grpc-types`, `iota-grpc-client` and the `iota-proto-build` crates from the monorepository to the iota-rust-sdk.

Should be rebased and merged after #1061 

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [ ] Patch-specific tests (correctness, functionality coverage)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have checked that new and existing unit tests pass locally with my changes


crates/iota-sdk-grpc-client/src/api/common.rs
```diff
@@ -0,0 +1,413 @@
...
+    #[error("server error (code {code}): {msg}", code = .0.code, msg = .0.message)]
+    Server(RpcStatus),
+
+    /// Client-side protocol error (e.g. checkpoint stream reassembly).
+    #[error("protocol error: {0}")]
+    Protocol(String),
+
+    /// Error converting signatures.
+    #[error("signature conversion error: {0}")]
+    Signature(String),
```
Thoralf-M: Any reason not to create own error variants (or own Errors enums that are wrapped here) instead of the String? Strings are not that nice if someone wants to do something based on specific errors

crates/iota-sdk-grpc-client/src/response_ext.rs
```diff
@@ -0,0 +1,196 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2025 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
```
Thoralf-M, line 3:
```suggestion
// SPDX-License-Identifier: Apache-2.0

```

Thoralf-M: Same in other files, please make sure to always have a newline after the license header

crates/iota-sdk-grpc-proto-build/src/codegen/accessors.rs
```diff
@@ -0,0 +1,881 @@
...
+                    #stream
+                }
+            };
+
+            let ast: syn::File = syn::parse2(code).expect("not a valid tokenstream");
+            let code = prettyplease::unparse(&ast);
+
+            // Add IOTA license header
+            buf.push_str("// Copyright (c) Mysten Labs, Inc.\n");
+            buf.push_str("// Modifications Copyright (c) 2025 IOTA Stiftung\n");
```
Thoralf-M: Should this be updated to 2026?

crates/iota-sdk/examples/polling-indexer/Cargo.toml
```diff
@@ -6,12 +6,15 @@ license = "Apache-2.0"
 description = "Polling-based custom indexer example using iota_sdk::graphql_client::Client"
 
 [dependencies]
+# external dependencies
```
Alex6323, line 9: Not sure about this change here. This is an example for other developers and they wouldn't differentiate.

.github/workflows/lints.yml
```diff
@@ -42,6 +42,26 @@ jobs:
 
       - run: make is-dirty
 
+  cargo-sort:
```
Alex6323, line 49: could this be a separate PR? Was wondering whether this is connected to grpc, but I guess you just wanted to move everything?

Thoralf-M: There is https://github.com/iotaledger/iota-rust-sdk/pull/1061

crates/iota-sdk-grpc-client/src/client.rs
```diff
@@ -0,0 +1,195 @@
...
+    channel: tonic::transport::Channel,
+    /// Headers interceptor for adding custom headers to requests
+    headers: HeadersInterceptor,
+    /// Maximum decoding message size for responses
+    max_decoding_message_size: Option<usize>,
+}
+
+impl Client {
+    /// Connect to a gRPC server and create a new Client instance.
+    #[allow(clippy::result_large_err)]
```
DaughterOfMars: Why is this lint here?

crates/iota-sdk-grpc-client/src/client.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub async fn connect<T>(uri: T) -> Result<Self>
```
DaughterOfMars: Why not call this `new`?

crates/iota-sdk-grpc-client/src/api/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+/// types using their conversion methods (e.g.,
+/// `response.summary()?.summary()?`, `response.contents()?.contents()?`).
+#[derive(Debug, Clone)]
+pub struct CheckpointResponse {
+    /// The checkpoint sequence number.
+    pub sequence_number: CheckpointSequenceNumber,
+    /// Proto checkpoint summary. Use `response.summary()?.summary()` to convert
+    /// to SDK type.
+    pub summary: Option<iota_grpc_types::v1::checkpoint::CheckpointSummary>,
+    /// Proto validator signature. Use `response.signature()?.signature()` to
```
DaughterOfMars, line 94: This is a bit awkward. Why not have a `summary` and `summary_grpc` or something?

crates/iota-sdk-grpc-client/src/api/common.rs
```diff
@@ -0,0 +1,413 @@
...
+                    page_token,
+                }
+            }
+
+            /// Auto-paginate through all pages, collecting up to `limit` items.
+            ///
+            /// If `limit` is `None`, collects all items across all pages.
+            pub async fn collect(
+                self,
+                limit: Option<u32>,
```
DaughterOfMars: Could use `impl Into<Option<u32>>` here

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
```diff
@@ -0,0 +1,647 @@
...
+    /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// let client = Client::connect("http://localhost:9000").await?;
+    /// let checkpoint = client.get_checkpoint_latest(None, None, None).await?;
+    /// println!("Received checkpoint {}", checkpoint.body().sequence_number,);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn get_checkpoint_latest(
+        &self,
+        read_mask: Option<&str>,
```
DaughterOfMars: Could/should these read masks be represented with a set of structs or enums? Using strings is pretty weird I feel when we could validate it via the type system

crates/iota-sdk-grpc-client/src/api/ledger/objects.rs
```diff
@@ -0,0 +1,121 @@
...
+    ///     println!("Got object ID: {:?}", sdk_obj.object_id());
+    ///     let obj_ref = obj.object_reference()?;
+    ///     println!("Object version: {:?}", obj_ref.version());
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn get_objects(
+        &self,
+        refs: &[(ObjectId, Option<Version>)],
```
DaughterOfMars, line 75: Since the most common use-case is to use None for version always, shouldn't we have a function that skips that and accepts a list of only object IDs?

crates/iota-sdk-grpc-proto-build/src/main.rs
```diff
@@ -0,0 +1,231 @@
...
+                .to_str()
+                .unwrap()
+                .contains("field_info")
+        {
+            let mut content = std::fs::read_to_string(&path).unwrap();
+
+            // Add license header if missing
+            if !content.starts_with("// Copyright") {
+                content = format!(
+                    "// Copyright (c) Mysten Labs, Inc.\n// Modifications Copyright (c) 2025 IOTA Stiftung\n// SPDX-License-Identifier: Apache-2.0\n\n{content}"
```
DaughterOfMars: 2026?

crates/iota-sdk/examples/polling-indexer/Cargo.toml
muXxer: Is it really such a biggie? That is autogenerated by the cargo sort script. If you don't want that, we either need to change the logic of the cargo sort script to skip certain files to not apply the formatting, or we need to exclude the example from the workspace.

thibault-martinez (commented):
Please add a workflow to check that grpc has been generated

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
muXxer: The read masks are kept as &str constants intentionally. They map directly to protobuf FieldMask paths, which are fundamentally string-based. The field_mask!() and field_masks_merge!() macros provide compile-time and runtime composition.

A typed enum system would face challenges with the hierarchical nature of field paths and context-dependent prefixes (e.g., ExecutedTransaction fields appear at different path depths: direct in get_transactions, transactions. in checkpoint endpoints, executed_transaction. in simulate). The current const &str constants in read_masks.rs are documented and linked to their accessor methods, providing practical compile-time safety without the complexity overhead of a type-level encoding.

crates/iota-sdk-grpc-client/src/api/ledger/objects.rs
muXxer: But then we would increase the API surface by another function. What if you want to make a batch call and for some you want the version, and for others not. I would prefer to leave it like it is if you don't mind.

DaughterOfMars: I really think the most likely use-case is to call this with no versions

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
DaughterOfMars: I don't really agree with this assessment. Even the macros only enforce correctness (I assume) but can not, for instance, suggest potential fields. Users must therefore read docs in order to use this, which must be strictly maintained. Whereas with a rust structure we could strictly define all of the possible fields and how they are composed, and tell users what is available via the code itself.

crates/iota-sdk/examples/polling-indexer/Cargo.toml
Alex6323: I don't think it's a biggie, tbh, at least not until people using this stuff complain. I was just pointing out that from the point of the SDK user starting off with an example this differentiation between external and internal dependencies doesn't make much sense anymore, as for them, our SDK is as external as any other dependency.

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/checkpoint.rs
```diff
@@ -0,0 +1,98 @@
...
+include!("../../../generated/iota.grpc.v1.checkpoint.field_info.rs");
+include!("../../../generated/iota.grpc.v1.checkpoint.accessors.rs");
+
+use crate::{
+    proto::{TryFromProtoError, get_inner_field},
+    v1::{bcs::BcsData, versioned::VersionedCheckpointSummary},
+};
+
+// CheckpointSummary
+//
```
Thoralf-M:
```suggestion
```

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/checkpoint.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// [`CHECKPOINT_SUMMARY_BCS`]).
+    ///
+    /// [`CHECKPOINT_SUMMARY_BCS`]: crate::read_masks::CHECKPOINT_SUMMARY_BCS
+    pub fn summary(&self) -> Result<iota_types::CheckpointSummary, TryFromProtoError> {
+        self.try_into()
+    }
+}
+
+// CheckpointContents
+//
```
Thoralf-M:
```suggestion
```

Thoralf-M: Also in other files

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
muXxer: I added now predefined read masks for all endpoints as enums.

crates/iota-sdk-grpc-client/src/client.rs
muXxer: I fixed it

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/checkpoint.rs
muXxer: fixed all occurences

crates/iota-sdk-grpc-client/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "iota-sdk-grpc-client"
+version = "0.0.1-alpha.1"
+authors = ["IOTA Foundation <info@iota.org>"]
+edition = "2024"
+license = "Apache-2.0"
+publish = false
```
Thoralf-M: We want to publish it in the future?
```suggestion
```
Same for the grpc types

.github/workflows/lints.yml
```diff
@@ -90,3 +90,27 @@ jobs:
         run: cargo +nightly install cargo-udeps
       - name: Run Cargo Udeps
         run: cargo +nightly udeps ${{ matrix.flags }}
+
+  proto-build:
```
Thoralf-M, line 94: Any other checks that would make sense to run in the CI? Are the e2e tests only in the iota repo for now?

crates/iota-sdk-grpc-types/src/field/field_mask_tree.rs
```diff
@@ -0,0 +1,379 @@
...
+impl FieldMaskTree {
+    pub fn new_wildcard() -> Self {
+        Self {
+            wildcard: true,
+            root: Default::default(),
+        }
+    }
+
+    pub fn add_field_path(&mut self, path: &str) -> &mut Self {
+        if self.wildcard || !is_valid_path(path) {
```
Thoralf-M, line 33: Don't want to return an error (Result) if it's not a valid path?

.github/workflows/lints.yml
muXxer: I think thats it. Only the monorepo can test the full chain. I added some small gRPC client tests, but that's it.

crates/iota-sdk-grpc-types/src/field/field_mask_tree.rs
muXxer: I added now constants for every possible field, so invalid paths could be passed, but it's highly unlikely if the user doesn't enforce that himself by ignoring the constants and using strings.

My little AI friend said:
No — it silently ignores invalid paths, which is the right behavior here. This is an internal building block, not a user-facing API. The paths come from our own &'static str constants (which are correct by definition) or from protobuf FieldMask values sent by clients. Invalid paths should just be ignored (the server won't return data for a field it doesn't recognize), not cause a panic or error propagation through the call stack.
                                                                                                               
Returning a Result would force every caller to handle an error that can't meaningfully occur when using the SDK constants, and would break the builder-pattern chaining (tree.add_field_path("a").add_field_path("b")).
                                                                                    
If the concern is about user-provided strings being silently swallowed, the server already handles that — requesting a nonexistent field just means it's absent from the response. That's the standard protobuf FieldMask behavior.

🤷 

I would keep it as it is.

crates/iota-sdk-grpc-client/Cargo.toml
muXxer: removed it! 👍

.github/workflows/lints.yml
Thoralf-M: We already have some localnet tests for graphql, should be possible to also have that for gRPC
Are the ones in the iota repo in a shape so they can easily be moved here or should we just create new tests in the SDK for that?
We can also do that later in another PR, just want to know if we should open an issue for that

muXxer: The question for me is, which tests should be where? Should the monorepo test the full functionality in case we break something there? Should the sdk test the full functionality against the localnet in case we break something here? So we have the exact same tests twice?

Thoralf-M: Hm good point, I guess since the iota repo imports the SDK, if everything is tested together there it should be sufficient

Thoralf-M (approved):

thibault-martinez (commented):
Can you impl `ClientMethods` for your grpc client? Otherwise we won't be able to use it with the TransactionBuilder.

Alex6323 (approved):

crates/iota-sdk-grpc-proto-build/README.md
thibault-martinez, line 1: There is some renaming to do here, crate names, `v0` that does not exist, script that does not exist, ...

crates/iota-sdk-grpc-types/proto/iota/grpc/options.proto
thibault-martinez, line 1: Missing copyright header

crates/iota-sdk-grpc-proto-build/src/message_graph.rs
```diff
@@ -0,0 +1,594 @@
...
+                self.path.push(idx as i32);
+                let comments = self.comments();
+                self.path.pop();
+
+                let name = method.name.take().unwrap();
+                let input_proto_type = method.input_type.take().unwrap();
+                let output_proto_type = method.output_type.take().unwrap();
+                // TODO FIX
+                // let input_type = self.resolve_ident(&input_proto_type);
+                // let output_type = self.resolve_ident(&output_proto_type);
```
thibault-martinez: Just checking if this was not forgotten, if not please open an issue

thibault-martinez (changes requested):

crates/iota-sdk-grpc-proto-build/Cargo.toml
thibault-martinez, line 1: This crate is not meant to be published, right? If yes, make it `publish = false`

crates/iota-sdk-grpc-proto-build/src/comments.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+use prost_types::source_code_info::Location;
+
+/// Comments on a Protobuf item.
+#[derive(Debug, Default, Clone)]
+#[allow(unused)]
```
thibault-martinez: Commenting here but this is more of a general question; why are there so many unused attributes? I counted 10 on types or functions. If unused, let's remove?

crates/iota-sdk-grpc-proto-build/src/main.rs
```diff
@@ -0,0 +1,231 @@
...
+use crate::{
+    codegen::generate_fields::FileDescriptorWithPackageVersion, message_graph::DescriptorGraph,
+};
+
+mod codegen;
+mod comments;
+mod context;
+mod dependency_graph;
+mod ident;
+mod message_graph;
```
thibault-martinez: Sorry, my rust OCD kicking in, mod before use please 🙏🏻

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/epoch.rs
```diff
@@ -0,0 +1,219 @@
...
+include!("../../../generated/iota.grpc.v1.epoch.field_info.rs");
+include!("../../../generated/iota.grpc.v1.epoch.accessors.rs");
+
+use tap::Pipe;
+
+use crate::{proto::TryFromProtoError, v1::bcs::BcsData};
+
+// ValidatorCommitteeMember
+
+impl From<iota_types::ValidatorCommitteeMember> for ValidatorCommitteeMember {
```
thibault-martinez, line 15: Are there roundtrip conversion tests for all these? If not, claude would probably do that easily

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/versioned.rs
thibault-martinez, line 1: Matter of taste but I feel like these types should just be with their inner type, not one module grouping versioned stuff, feel free to resolve, not asking you to change

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
```diff
@@ -0,0 +1,582 @@
...
+            request = request.with_events_filter(ef);
+        }
+        if let Some(max_size) = self
+            .max_decoding_message_size()
+            .map(saturating_usize_to_u32)
+        {
+            request = request.with_max_message_size_bytes(max_size);
+        }
+
+        let mut client = self.ledger_service_client();
```
thibault-martinez, line 207: Just wondering, how heavy is it to recreate such clients per request? If heavy, can we not cache it?

crates/iota-sdk-grpc-proto-build/src/comments.rs
muXxer: removed a lot of code, lol

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/versioned.rs
muXxer: I hope that your team replaces those types at some point with versioned types in iota-types. That's why we collected all of them in one file.

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
muXxer: At the beginning I actually thought the same and cached all the clients, but it's actually super cheap to create them and the overhead and complexity is not worth it.

thibault-martinez (approved):

DaughterOfMars (approved):

crates/iota-sdk-grpc-client/src/api/mod.rs
DaughterOfMars: Created issue https://github.com/iotaledger/iota-rust-sdk/issues/1141

crates/iota-sdk-grpc-client/src/api/ledger/objects.rs
DaughterOfMars: Created issue https://github.com/iotaledger/iota-rust-sdk/issues/1142

PR #1077, opened 2026-04-22
feat(CI): use localnet script in CI

### LINKED ISSUE #1052: Actually use `run_localnet.sh` in the CI
Following https://github.com/iotaledger/iota-rust-sdk/pull/1030, we have a CI to test the script but we're still not using the script for test CIs, leading to some duplication.
Maybe the localnet workflow can be removed if we start using the script in another workflow.

Thoralf-M:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/1052


run_localnet.sh
```diff
@@ -7,7 +7,7 @@ COMPOSE_PATH="./.github/actions/start-local-network/gas_station_compose.yml"
 COMPOSE_LOCAL_PATH="./.github/actions/start-local-network/gas_station_compose.local.yml"
 CONFIG_BACKUP="$CONFIG_PATH.backup"
 IOTA_LOG="iota_network.log"
-IOTA_BINARY="${2:-iota}"
+IOTA_BINARY="${2:-iota-localnet}"
```
thibault-martinez: Maybe `IOTA_LOCALNET_BINARY` to avoid confusion

Thoralf-M: Sure

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1077/changes/71b99d67608b51623af7a28ecf8ad849f57d1eba

DaughterOfMars (approved):

run_localnet.sh
```diff
@@ -22,11 +22,16 @@ if [ "$1" == "start" ]; then
 
     # Start PostgreSQL
     echo "Starting PostgreSQL..."
-    docker start postgres || docker run -d --name postgres -e POSTGRES_PASSWORD=postgrespw -e POSTGRES_INITDB_ARGS="-U postgres" -p 5432:5432 postgres:15 -c max_connections=1000
+    docker start postgres || docker run -d --name postgres \
+        -e POSTGRES_PASSWORD=postgrespw \
+        -e POSTGRES_INITDB_ARGS="-U postgres" \
+        ${POSTGRES_DB:+-e POSTGRES_DB="$POSTGRES_DB"} \
+        ${POSTGRES_HOST_AUTH_METHOD:+-e POSTGRES_HOST_AUTH_METHOD="$POSTGRES_HOST_AUTH_METHOD"} \
```
thibault-martinez: Why are the `-e` only in the default case?

Thoralf-M:
```
Because docker run is the "create new container" path, and Docker env vars are only applied at creation. docker start postgres just restarts an existing container with whatever env vars it was created with — you can't retroactively change them
```

thibault-martinez: That's not really what I asked. Maybe I don't understand this syntax
`${POSTGRES_DB:+-e POSTGRES_DB="$POSTGRES_DB"}`
There is `-e` only in one case?

Thoralf-M: Ah that's only if the POSTGRES_DB exists

Thoralf-M: But seems like we can actually remove some stuff here

thibault-martinez (approved):

PR #1082, opened 2026-04-23
refactor(iota-sdk-types): drop byte-level signature serialization

### LINKED ISSUE #1079: Investigate why signatures are manually serialised
And not use the regular BCS flow.
That would allow us having the ABNF schemas for them.

Thoralf-M:
Closes #1079

## Why

The signature types hand-rolled their BCS serialization with manual byte building (`buf.push(flag); buf.extend(sig); buf.extend(pk)`) and matching manual parsers (`bytes[0]` flag checks, hand-counted length offsets). That's error-prone and means the generated ABNF can only describe a signature as opaque `bytes`.

## What

Each signature type now round-trips through a private **wire-shape "Body" type** deriving `Deserialize` + `BcsSchema` (plus `Serialize` where that doesn't force a clone). The custom `Serialize`/`Deserialize` impls shrink to a thin dispatcher:

- **binary** → `serialize_bytes` of the flat `[scheme-flag][payload]` body; decode parses the body and converts back
- **human-readable** → delegate to a `Readable*` enum (preserves the existing `{"scheme": "ed25519", …}` JSON)

No more hand-written byte fiddling, flag parsing, or length accounting.

`MultisigAggregatedSignature` keeps a thin custom `Serialize`/`Deserialize` so its **standalone** BCS form stays the historical length-prefixed `[flag][body]` — a plain derive would have silently changed the public type's wire format. `UserSignatureBody` embeds a flat multisig pivot (and a flat passkey pivot) for the same reason, and its serialize path delegates per-variant rather than cloning the payload.

### Why the custom `Serialize` can't go away entirely

On the IOTA wire, an embedded signature is **length-prefixed bytes**: a ULEB128 length followed by `[scheme-flag][payload]`. A plain `#[derive(Serialize)]` enum produces `[tag][fields]` with no outer length, so the `serialize_bytes` wrapping has to live somewhere. Putting it *on the type* keeps a single canonical wire format and one ABNF rule per type.

## ABNF result

The schema is now structural instead of opaque. `user-signature = bytes` (the wire wrapping) gains a sibling `user-signature-body` enumerating the scheme variants, and embedding sites reference the typed name:

```abnf
user-signature = bytes

user-signature-body = %d00 ed25519-signature ed25519-public-key       ; Ed25519
                    / %d01 secp256k1-signature secp256k1-public-key   ; Secp256k1
                    / %d02 secp256r1-signature secp256r1-public-key   ; Secp256r1
                    / %d03 multisig-aggregated-signature              ; Multisig
                    / %d06 passkey-authenticator                      ; Passkey
                    / %d07 move-authenticator                         ; Move

signed-transaction = transaction              ; transaction
                     (size *user-signature)   ; signatures   <-- was (size *bytes)
```

Reserved scheme flags `0x04` (BLS) and `0x05` (deprecated zkLogin) are kept as placeholder variants — rejected on decode and omitted from the grammar via `#[bcs_schema(skip)]` (the variant-skip attribute landed in #1262). `simple-signature{,-body}`, `multisig-aggregated-signature`, `multisig-member-signature`, `multisig-committee`, `multisig-member`, `passkey-authenticator`, and `move-authenticator{,-v1}` all gain structural rules too.

## Notes

- **Public API unchanged.** `UserSignature` keeps its 5 variants; `SimpleSignature`, `MultisigMemberSignature`, etc. are untouched. The flat `*Body` types are private.
- **BCS wire format is byte-identical** — locked in by the `simple_fixtures` / `multisig_fixtures` / `passkey_fixtures` round-trip tests, plus a standalone-multisig framing test.
- **JSON shape unchanged** — all signature types preserve their existing internally-tagged `{"scheme": "…", …}` form via `Readable*` dispatcher enums.
- **`PasskeyAuthenticator`** keeps its challenge-derivation validation on the deserialize path via an `Authenticator` pivot.

## Test plan

- `cargo test -p iota-sdk-types --all-features` — all unit + fuzzing + doc tests pass
- `cargo build --workspace --all-features` — clean
- `cargo clippy` — clean with `-D warnings`
- ABNF regenerated via `BCS_SCHEMA=1 cargo check`; grammar-driven `bcs_schema_fuzzing` test passes

https://claude.ai/code/session_01Uavu5pByHv9y9ZjrLBamya

### DIFF
--- a/crates/iota-sdk-types/bcs-schema.abnf
+++ b/crates/iota-sdk-types/bcs-schema.abnf
@@ -145,6 +145,10 @@ consensus-determined-version-assignments = %d00 (size *cancelled-transaction)   
 
 digest = %d32 32OCTET
 
+ed25519-public-key = 32OCTET
+
+ed25519-signature = 64OCTET
+
 effects-aux-data-digest = digest
 
 end-of-epoch-data = (size *validator-committee-member)   ; next-epoch-committee
@@ -267,7 +271,13 @@ merge-coins = argument           ; coin
 
 misbehavior-report-digest = digest
 
+move-authenticator = %d00 move-authenticator-v1   ; V1
+
 move-authenticator-digest = digest
+
+move-authenticator-v1 = (size *input)      ; call-args
+                        (size *type-tag)   ; type-args
+                        input              ; object-to-authenticate
 
 move-call = object-id          ; package
             identifier         ; module
@@ -290,6 +300,21 @@ move-package = object-id                          ; id
 move-struct = compressed-struct-tag   ; object-type
               version                 ; version
               bytes                   ; contents
+
+multisig-aggregated-signature = (size *multisig-member-signature)   ; signatures
+                                u16                                 ; bitmap
+                                multisig-committee                  ; committee
+
+multisig-committee = (size *multisig-member)   ; members
+                     u16                       ; threshold
+
+multisig-member = public-key   ; public-key
+                  u8           ; weight
+
+multisig-member-signature = %d00 ed25519-signature     ; Ed25519
+                          / %d01 secp256k1-signature   ; Secp256k1
+                          / %d02 secp256r1-signature   ; Secp256r1
+                          / %d04 bytes                 ; Passkey
 
 object = object-data          ; data
          owner                ; owner
@@ -330,8 +355,19 @@ package-upgrade-error = %d00 object-id             ; UnableToFetchPackage
                       / %d04 u8                    ; UnknownUpgradePolicy
                       / %d05 object-id object-id   ; PackageIdDoesNotMatch
 
+passkey-authenticator = bytes              ; authenticator-data
+                        string             ; client-data-json
+                        simple-signature   ; signature
+
+passkey-public-key = secp256r1-public-key
+
 programmable-transaction = (size *input)     ; inputs
                            (size *command)   ; commands
+
+public-key = %d00 ed25519-public-key     ; Ed25519
+           / %d01 secp256k1-public-key   ; Secp256k1
+           / %d02 secp256r1-public-key   ; Secp256r1
+           / %d04 passkey-public-key     ; Passkey
 
 publish = (size *bytes)       ; modules
           (size *object-id)   ; dependencies
@@ -343,6 +379,14 @@ randomness-state-update = u64                ; epoch
                           bytes              ; random-bytes
                           version            ; randomness-obj-initial-shared-version
 
+secp256k1-public-key = 33OCTET
+
+secp256k1-signature = 64OCTET
+
+secp256r1-public-key = 33OCTET
+
+secp256r1-signature = 64OCTET
+
 sender-signed-data-digest = digest
 
 signed-checkpoint-summary = checkpoint-summary               ; checkpoint
@@ -350,6 +394,12 @@ signed-checkpoint-summary = checkpoint-summary               ; checkpoint
 
 signed-transaction = transaction              ; transaction
                      (size *user-signature)   ; signatures
+
+simple-signature = bytes
+
+simple-signature-body = %d00 ed25519-signature ed25519-public-key       ; Ed25519
+                      / %d01 secp256k1-signature secp256k1-public-key   ; Secp256k1
+                      / %d02 secp256r1-signature secp256r1-public-key   ; Secp256r1
 
 size    = uleb128   ; BCS sequence/string length (ULEB128-encoded)
 
@@ -458,6 +508,13 @@ upgrade-info = object-id   ; upgraded-id
                version     ; upgraded-version
 
 user-signature = bytes
+
+user-signature-body = %d00 ed25519-signature ed25519-public-key       ; Ed25519
+                    / %d01 secp256k1-signature secp256k1-public-key   ; Secp256k1
+                    / %d02 secp256r1-signature secp256r1-public-key   ; Secp256r1
+                    / %d03 multisig-aggregated-signature              ; Multisig
+                    / %d06 passkey-authenticator                      ; Passkey
+                    / %d07 move-authenticator                         ; Move
 
 validator-aggregated-signature = u64                  ; epoch
                                  bls12381-signature   ; signature
--- a/crates/iota-sdk-types/src/crypto/multisig.rs
+++ b/crates/iota-sdk-types/src/crypto/multisig.rs
@@ -1067,24 +1067,31 @@ mod tests {
     fn multisig_standalone_bcs_framing() {
         use base64ct::{Base64, Encoding};
 
-        // A multisig fixture from mainnet, as the `bytes`-wrapped
-        // `UserSignature` form: `[uleb-len][0x03 flag][body]`.
-        let multisig_b64 = "sgIDAwCTLgVngjC4yeuvpAGKVkgcvIKVFUJnL1r6oFZScQVE5DNIz6kfxAGDRcVUczE9CUb7/sN/EuFJ8ot86Sdb8pAFASoQ91stRHXdW5dLy0BQ6v+7XWptawy2ItMyPk508p+PHdtZcm2aKl3lZGIvXe6MPY73E+1Hakv/xJbTYsw5SPMC5dx3gBwxds2GV12c7VUSqkyXamliSF1W/QBMufqrlmdIOZ1ox9gbsvIPtXYahfvKm8ozA7rsZWwRv8atsnyfYgcAAwANfas1jI2tqk76AEmnWwdDZVWxCjaCGbtoD3BXE0nXdQEBAg4XzVk55GsZZkGWjNdZkQuiV34n+nP944dtub7FvOsrAQIDR/uvI/A4q8TDCKJxEXoqTP+u3bxf+Bx1F7xsdKfttDABAgA=";
-        let user_signature_bcs = Base64::decode_vec(multisig_b64).unwrap();
-        let UserSignature::Multisig(sig) =
-            bcs::from_bytes::<UserSignature>(&user_signature_bcs).unwrap()
+        // A multisig `UserSignature` from a mainnet transaction, in the flat
+        // `[0x03 flag][body]` wire form (the `to_bytes` / explorer encoding):
+        // https://explorer.iota.org/txblock/BUPSmkG8QZgr1NtVNjFaJMYsArkaWMECs7RHbEZRZUEU?network=https%3A%2F%2Findexer.mainnet.iota.cafe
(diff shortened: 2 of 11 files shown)


crates/iota-sdk-types/bcs-schema.abnf
```diff
@@ -75,8 +75,8 @@ checkpoint-commitment = %d00 digest   ; EcmhLiveObjectSet
 
 checkpoint-contents = %d00 checkpoint-contents-v1   ; V1
 
-checkpoint-contents-v1 = (size *execution-digests)        ; digests
-                         (size *(size *user-signature))   ; signatures
+checkpoint-contents-v1 = (size *execution-digests)   ; digests
+                         (size *(size *bytes))       ; signatures
```
alexsporn: shouldn't this point to a type if you now have proper types for all signatures

crates/iota-sdk-types/bcs-schema.abnf
```diff
@@ -99,6 +99,10 @@ checkpoint-transaction = %d01 intent-signed-transaction     ; transaction
                          (size *object)                     ; input-objects
                          (size *object)                     ; output-objects
 
+checkpoint-transaction-info = digest          ; transaction
+                              digest          ; effects
+                              (size *bytes)   ; signatures
```
alexsporn: shouldn't this point to a type if you now have proper types for all signatures

crates/iota-sdk-types/bcs-schema.abnf
```diff
@@ -243,9 +247,9 @@ intent-scope = %d00   ; TransactionData
              / %d08   ; DiscoveryPeers
              / %d09   ; AuthorityCapabilities
 
-intent-signed-transaction = intent                   ; intent
-                            transaction              ; transaction
-                            (size *user-signature)   ; signatures
+intent-signed-transaction = intent          ; intent
+                            transaction     ; transaction
+                            (size *bytes)   ; signatures
```
alexsporn: shouldn't this point to a type if you now have proper types for all signatures

crates/iota-sdk-types/bcs-schema.abnf
```diff
@@ -328,8 +364,12 @@ randomness-state-update = u64     ; epoch
 signed-checkpoint-summary = checkpoint-summary               ; checkpoint
                             validator-aggregated-signature   ; signature
 
-signed-transaction = transaction              ; transaction
-                     (size *user-signature)   ; signatures
+signed-transaction = transaction     ; transaction
+                     (size *bytes)   ; signatures
```
alexsporn: shouldn't this point to a type if you now have proper types for all signatures

crates/iota-sdk-types/bcs-schema.abnf
Thoralf-M: Good call — fixed across all four sites. The PR has been reworked so the signature types now own their structural ABNF rules: `user-signature = bytes` with a sibling `user-signature-body` enumerating the scheme variants, and likewise `simple-signature` / `simple-signature-body`, plus full structural rules for `multisig-aggregated-signature`, `multisig-member-signature`, `passkey-authenticator`, and `move-authenticator`.

As a result the embedding sites you flagged now read `(size *user-signature)` (and `(size *(size *user-signature))` for checkpoint contents) instead of `(size *bytes)` — they reference the typed name directly. The `= bytes` on `user-signature` itself is the historical length-prefix wrapping that IOTA/Sui uses for embedded signatures (a ULEB128 length followed by `[scheme-flag][payload]`), which is unavoidable on the wire; the `-body` rule documents what's inside those bytes.

---

DaughterOfMars (approved):

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -599,20 +605,42 @@ mod serialization {
...
 
+    /// Owned wire shape of an aggregated multisig: the flat
+    /// `signatures || bitmap || committee` body without the scheme flag or
+    /// length prefix. `UserSignatureBody::Multisig` embeds this type directly
+    /// (the enum tag supplies the flag and the outer `UserSignature` the
+    /// length prefix), while `MultisigAggregatedSignature`'s own serde keeps
+    /// the historical `bytes`-wrapped `flag || body` form.
     #[derive(serde::Deserialize)]
-    pub struct Multisig {
+    pub(crate) struct Multisig {
```
thibault-martinez, line 615: Do we actually need this type?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -1027,6 +1059,65 @@ mod tests {
...
+    /// length-prefixed bytes containing `multisig-flag || body`, exactly the
+    /// `to_bytes` form. A round-trip test cannot detect losing (or doubling)
+    /// that framing, so it is pinned here byte-by-byte.
+    #[test]
+    fn multisig_standalone_bcs_framing() {
+        use base64ct::{Base64, Encoding};
+
+        // A multisig fixture from mainnet, as the `bytes`-wrapped
+        // `UserSignature` form: `[uleb-len][0x03 flag][body]`.
+        let multisig_b64 = "sgIDAwCTLgVngjC4yeuvpAGKVkgcvIKVFUJnL1r6oFZScQVE5DNIz6kfxAGDRcVUczE9CUb7/sN/EuFJ8ot86Sdb8pAFASoQ91stRHXdW5dLy0BQ6v+7XWptawy2ItMyPk508p+PHdtZcm2aKl3lZGIvXe6MPY73E+1Hakv/xJbTYsw5SPMC5dx3gBwxds2GV12c7VUSqkyXamliSF1W/QBMufqrlmdIOZ1ox9gbsvIPtXYahfvKm8ozA7rsZWwRv8atsnyfYgcAAwANfas1jI2tqk76AEmnWwdDZVWxCjaCGbtoD3BXE0nXdQEBAg4XzVk55GsZZkGWjNdZkQuiV34n+nP944dtub7FvOsrAQIDR/uvI/A4q8TDCKJxEXoqTP+u3bxf+Bx1F7xsdKfttDABAgA=";
```
thibault-martinez: Did you create these bytes or did you fetch them from a network?

crates/iota-sdk-types/src/crypto/multisig.rs
Thoralf-M: Yes — it's the flat-body deserialize target for the *embedded* multisig form. Nested inside a `UserSignature`, a multisig is a bare `signatures || bitmap || committee` body (the enum tag supplies the `0x03` flag, the outer `UserSignature` the length prefix). `MultisigAggregatedSignature`'s own `Deserialize` can't be reused there — it reads the *standalone* `bytes`-wrapped `flag || body` form. This type is also where the validating `TryFrom` conversion lives, so it can't just be dropped.

---

crates/iota-sdk-types/src/crypto/multisig.rs
Thoralf-M: They were from sui from the other file, replaced by real ones now https://github.com/iotaledger/iota-rust-sdk/pull/1082/commits/9e93b443a4e90b79d0bb4373383026fae0dc9e98

Alex6323 (approved):

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -692,29 +720,33 @@ mod serialization {
...
+    )]
     enum MemberSignature {
         Ed25519(Ed25519Signature),
         Secp256k1(Secp256k1Signature),
         Secp256r1(Secp256r1Signature),
+        #[cfg_attr(feature = "bcs-schema", bcs_schema(skip))]
         ZkLoginDeprecated,
-        Passkey(PasskeyAuthenticator),
+        Passkey(
+            #[cfg_attr(feature = "bcs-schema", bcs_schema(as_type = "bytes"))] PasskeyAuthenticator,
```
thibault-martinez, line 748: this is length prefixed?

thibault-martinez: Ah yeah, you have an example about it

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -1027,6 +1059,72 @@ mod tests {
         );
     }
 
+    /// A standalone `MultisigAggregatedSignature` BCS-serializes as
```
thibault-martinez, line 1062: What does standalone mean here?

crates/iota-sdk-types/bcs-schema.abnf
```diff
@@ -292,6 +302,21 @@ move-struct = compressed-struct-tag   ; object-type
...
+multisig-committee = (size *multisig-member)   ; members
+                     u16                       ; threshold
+
+multisig-member = public-key   ; public-key
+                  u8           ; weight
+
+multisig-member-signature = %d00 ed25519-signature     ; Ed25519
+                          / %d01 secp256k1-signature   ; Secp256k1
+                          / %d02 secp256r1-signature   ; Secp256r1
+                          / %d04 bytes                 ; Passkey
```
thibault-martinez, line 318: I don't really understand why this is bytes and `user-signature-body::move-authenticator` is not

crates/iota-sdk-types/src/crypto/multisig.rs
Thoralf-M: That it's serialized without wrapping in another type (UserSignature)

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -1027,6 +1059,72 @@ mod tests {
...
+    /// A passkey member signature is BCS-framed as `%d04` followed by
+    /// length-prefixed bytes containing `passkey-flag || passkey body` — the
+    /// historical wrapping produced by `PasskeyAuthenticator`'s serde. A
+    /// round-trip test cannot detect losing (or doubling) that framing, so it
+    /// is pinned here byte-by-byte.
+    #[test]
+    fn member_signature_passkey_bcs_framing() {
+        use base64ct::{Base64, Encoding};
+
+        let passkey_b64 = "BiVYDmenOnqS+thmz5m5SrZnWaKXZLVxgh+rri6LHXs25B0AAAAAnQF7InR5cGUiOiJ3ZWJhdXRobi5nZXQiLCAiY2hhbGxlbmdlIjoiQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQSIsIm9yaWdpbiI6Imh0dHA6Ly9sb2NhbGhvc3Q6NTE3MyIsImNyb3NzT3JpZ2luIjpmYWxzZSwgInVua25vd24iOiAidW5rbm93biJ9YgJMwqcOmZI7F/N+K5SMe4DRYCb4/cDWW68SFneSHoD2GxKKhksbpZ5rZpdrjSYABTCsFQQBpLORzTvbj4edWKd/AsEBeovrGvHR9Ku7critg6k7qvfFlPUngujXfEzXd8Eg";
```
thibault-martinez, line 1107: Where is this from?

Thoralf-M: From Sui

crates/iota-sdk-types/bcs-schema.abnf
Thoralf-M: - A multisig member passkey embeds `PasskeyAuthenticator` via its own `Serialize`, which does `serialize_bytes(flag || body)` → ULEB-length-prefixed `bytes`. That's the pre-existing on-wire framing for multisig members, so it stays byte-identical → grammar shows `%d04 bytes`. (Pinned by the `member_signature_passkey_bcs_framing` test.)
- `MoveAuthenticator` is a plain derive — no bytes wrapping — so in `user-signature-body` it serializes structurally → the grammar references the `move-authenticator` rule directly.

---

thibault-martinez (approved):

PR #1088, opened 2026-04-27
feat(types): enhance Digest with additional methods and constants

DaughterOfMars:
## Summary

Fourth PR in the SDK replacement series, stacked on top of `sdk-replacement/3-object-id`. Enhances `Digest` with additional constants, methods, and a structured parse error, and exposes the new surface through the FFI and generated bindings.

### `iota-sdk-types`

- **`Digest` constants:** added `MIN`, `MAX`, `OBJECT_DELETED`, `OBJECT_WRAPPED`, `OBJECT_CANCELLED`, and `GENESIS_MARKER` to model the special digest values used to mark object lifecycle state and the genesis transaction parent.
- **Object-state predicates:** `is_object_alive`, `is_object_deleted`, `is_object_wrapped` on `Digest`.
- **`Digest::random()`** behind the `rand` feature, replacing ad-hoc `generate(&mut rng)` calls in the FFI.
- **`UpperHex` impl** for `Digest` (with `{:#X}` 0x-prefix support), complementing the existing `LowerHex`.
- **`next_lexicographical_opt`:** new shared `const fn` helper in `lib.rs` returning `Option<[u8; N]>` (vs. the wrapping `next_lexicographical`), exposed on `Digest`, `Address`, and `ObjectId`. `next_lexicographical` is now `const fn`.
- **`DigestParseError` rewrite:** replaced the unit struct + manual `Display`/`Error` impls with a `thiserror`-derived enum carrying `Base58(bs58::decode::Error)` and `InvalidByteLength { actual }` variants for actionable parse failures.
- **`Digest::from_bytes`** now reports the actual byte length on mismatch; `from_base58` is reimplemented in terms of `from_bytes` and now correctly surfaces base58 decode errors (previously they were swallowed into a generic error).
- Test suite expanded to cover valid/empty/short/long base58 and byte inputs.


Thoralf-M (approved):

Alex6323 (approved):

crates/iota-sdk-types/src/lib.rs
```diff
@@ -201,6 +201,25 @@ pub const fn next_lexicographical_array<const N: usize>(array: &[u8; N]) -> [u8;
     next
 }
 
+/// Returns the next array in byte-increasing order, or `None` if the result
+/// would overflow.
+pub const fn next_lexicographical_array_opt<const N: usize>(array: &[u8; N]) -> Option<[u8; N]> {
```
thibault-martinez, line 195: Why is `next_lexicographical_array` not calling this one?

DaughterOfMars: fair point, I changed it

thibault-martinez: I was kinda expecting you to unwrap, I'm wondering if `[0; N]` is a correct behaviour

DaughterOfMars: It wraps, that's the original behavior

thibault-martinez (approved):

PR #1089, opened 2026-04-27
feat(iota-sdk-types): derive AddAssign and SubAssign for GasCostSummary

Thoralf-M:
## Summary

Continues the SDK replacement series. Adds in-place arithmetic to `GasCostSummary` and makes all of its arithmetic overflow-checked, so callers can accumulate or subtract gas costs without manually destructuring all five fields and without silently wrapping on overflow.

- **`AddAssign` / `SubAssign` on `GasCostSummary`** in [crates/iota-sdk-types/src/gas.rs](crates/iota-sdk-types/src/gas.rs):
  - **Borrowed RHS** via hand-written `impl AddAssign<&Self>` and `impl SubAssign<&Self>` so accumulation loops can keep ownership of the summand (`total += &item.gas_cost_summary` instead of `total += item.gas_cost_summary.clone()`).
  - **Owned RHS** via hand-written `impl AddAssign<Self>` and `impl SubAssign<Self>` that delegate to the borrowed impls. (Previously these were produced by `derive_more::AddAssign` / `SubAssign`, but those expand to plain `+=` / `-=` and would silently wrap on overflow — see below.)
- **Overflow-checked arithmetic** on every field-wise operation: `AddAssign`/`SubAssign` (both `&Self` and `Self`) as well as the existing `gas_used()` and `net_gas_usage()` helpers now use `checked_add` / `checked_sub` and panic on overflow/underflow instead of wrapping. Gas accounting must never silently wrap, so an explicit panic is the correct failure mode here.


DaughterOfMars (approved):

crates/iota-sdk-types/src/gas.rs
roman1e2f5p8s: check for extra whitespaces here and trailing ones in lines 15-16

Alex6323 (approved):

crates/iota-sdk-types/src/gas.rs
thibault-martinez: This doc comment is overall a bit rekt, please clean up

crates/iota-sdk-types/src/gas.rs
```diff
@@ -103,6 +103,26 @@ impl GasCostSummary {
...
 
+impl std::ops::AddAssign<&Self> for GasCostSummary {
+    fn add_assign(&mut self, other: &Self) {
+        self.computation_cost += other.computation_cost;
+        self.computation_cost_burned += other.computation_cost_burned;
+        self.storage_cost += other.storage_cost;
+        self.storage_rebate += other.storage_rebate;
+        self.non_refundable_storage_fee += other.non_refundable_storage_fee;
+    }
+}
```
roman1e2f5p8s, line 133: in the monorepo, these impl blocks were put in a module annotated with `#[iota_macros::with_checked_arithmetic]`, expanding += to `checked_add()` - here they are not. that means in the monorepo on develop += panics on overflow in both debug and release. but here in the SDK `AddAssign` is plain u64 +=, which panics in debug only and silently wraps in release.

crates/iota-sdk-types/src/gas.rs
```diff
@@ -103,6 +103,26 @@ impl GasCostSummary {
...
+
+impl std::ops::SubAssign<&Self> for GasCostSummary {
+    fn sub_assign(&mut self, other: &Self) {
+        self.computation_cost -= other.computation_cost;
+        self.computation_cost_burned -= other.computation_cost_burned;
+        self.storage_cost -= other.storage_cost;
+        self.storage_rebate -= other.storage_rebate;
+        self.non_refundable_storage_fee -= other.non_refundable_storage_fee;
+    }
+}
```
roman1e2f5p8s, line 170: the addition seems to be used for epoch rolling gas cost summary, but I am not sure what the subtraction could be used for. nevertheless, same silent underflow issue in the release applies here.

crates/iota-sdk-types/src/gas.rs
thibault-martinez: This is something we noticed and fixed later in another PR that will also be replayed https://github.com/iotaledger/iota-rust-sdk/pull/555/changes#diff-3bc3a208bd3bcdf91578709955758f7d688f1f2d4949c4a0439efea7e58fcd27. If you prefer, we could integrate these changes here already.

Thoralf-M: Added it now so the history is cleaner https://github.com/iotaledger/iota-rust-sdk/pull/1089/commits/ad05bec2a42031f81764e889b628ebe4901f432a

crates/iota-sdk-types/src/gas.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1089/commits/ad05bec2a42031f81764e889b628ebe4901f432a

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+            .unwrap()
+    }
+}
+
+impl std::ops::AddAssign<&Self> for GasCostSummary {
+    fn add_assign(&mut self, other: &Self) {
+        self.computation_cost = self
+            .computation_cost
+            .checked_add(other.computation_cost)
+            .unwrap();
```
roman1e2f5p8s:
```suggestion
            .expect("GasCostSummary::add_assign overflow on computation_cost");
```

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+        self.computation_cost_burned = self
+            .computation_cost_burned
+            .checked_add(other.computation_cost_burned)
+            .unwrap();
```
roman1e2f5p8s:
```suggestion
            .expect("GasCostSummary::add_assign overflow on computation_cost_burned");
```

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+        self.storage_cost = self.storage_cost.checked_add(other.storage_cost).unwrap();
```
roman1e2f5p8s:
```suggestion
        self.storage_cost = self.storage_cost
            .checked_add(other.storage_cost)
            .expect("GasCostSummary::add_assign overflow on storage_cost");
```

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+        self.storage_rebate = self
+            .storage_rebate
+            .checked_add(other.storage_rebate)
+            .unwrap();
```
roman1e2f5p8s:
```suggestion
            .expect("GasCostSummary::add_assign overflow on storage_rebate");
```

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+        self.non_refundable_storage_fee = self
+            .non_refundable_storage_fee
+            .checked_add(other.non_refundable_storage_fee)
+            .unwrap();
```
roman1e2f5p8s:
```suggestion
            .expect("GasCostSummary::add_assign overflow on non_refundable_storage_fee");
```

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+    }
+}
+
+impl std::ops::SubAssign<&Self> for GasCostSummary {
+    fn sub_assign(&mut self, other: &Self) {
+        self.computation_cost = self
+            .computation_cost
+            .checked_sub(other.computation_cost)
+            .unwrap();
```
roman1e2f5p8s:
```suggestion
            .expect("GasCostSummary::sub_assign underflow on computation_cost");
```

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+        self.computation_cost_burned = self
+            .computation_cost_burned
+            .checked_sub(other.computation_cost_burned)
+            .unwrap();
```
roman1e2f5p8s:
```suggestion
            .expect("GasCostSummary::sub_assign underflow on computation_cost_burned");
```

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+        self.storage_cost = self.storage_cost.checked_sub(other.storage_cost).unwrap();
```
roman1e2f5p8s:
```suggestion
        self.storage_cost = self.storage_cost
            .checked_sub(other.storage_cost)
            .expect("GasCostSummary::sub_assign underflow on storage_cost");
```

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+        self.storage_rebate = self
+            .storage_rebate
+            .checked_sub(other.storage_rebate)
+            .unwrap();
```
roman1e2f5p8s:
```suggestion
            .expect("GasCostSummary::sub_assign underflow on storage_rebate");
```

crates/iota-sdk-types/src/gas.rs
```diff
@@ -93,13 +93,73 @@ impl GasCostSummary {
...
+        self.non_refundable_storage_fee = self
+            .non_refundable_storage_fee
+            .checked_sub(other.non_refundable_storage_fee)
+            .unwrap();
```
roman1e2f5p8s:
```suggestion
            .expect("GasCostSummary::sub_assign underflow on non_refundable_storage_fee");
```

roman1e2f5p8s (commented):
small suggestions

crates/iota-sdk-types/src/gas.rs
Thoralf-M: GasCostSummary::add_assign one should see in the stacktrace already, what do you think about something like `"computation_cost sum overflows u64"` instead?

roman1e2f5p8s: Right. To better match the original panic message from the `with_checked_arithmetic` macro, you could output the operand values too: `.unwrap_or_else(|| panic!("computation_cost overflows u64: {} + {}", self.computation_cost, other.computation_cost))`

thibault-martinez (approved):

crates/iota-sdk-types/src/gas.rs
Thoralf-M: That looks overkill to me since this is not expected to happen in the first place. For errors I would definitely agree that the values should be part of the error message, so it's better to debug, but in this case here it should never happen unless something is completely wrong and then the specific values are probably not that interesting

roman1e2f5p8s (approved):

PR #1092, opened 2026-04-28
feat(types): enhance Identifier, TypeTag and StructTag

DaughterOfMars:
## Summary

Reorganize Move core types and substantially expand their APIs.

- Introduce a new `iota_sdk::types::move_core` module that owns `Identifier`, `TypeTag`, `StructTag`, the type-tag parser, and serde glue (replacing the old `type_tag` module). Re-exported at the crate root for source compatibility where possible.
- Replace the prior `Identifier` / `IdentifierRef` split with a single `Identifier` newtype backed by `bytestring::ByteString`. Adds `const fn from_static`, `is_valid`, `as_str`, `as_bytes`, and a large set of `pub const` module-name and type-name constants (e.g. `Identifier::COIN_MODULE`, `Identifier::IOTA`, `Identifier::TX_CONTEXT`).
- Expand `StructTag` with many constructor / predicate pairs for common framework, system, and std types (`new_gas` / `is_gas`, `new_gas_coin` / `is_gas_coin`, `new_id` / `is_id`, `new_uid`, `new_dynamic_field` / `is_dynamic_field`, `new_clock`, `new_random`, `new_url`, `new_bag`, `new_object_bag`, `new_tx_context`, `new_authenticator_state`, `new_upgrade_cap` / `new_upgrade_ticket` / `new_upgrade_receipt`, `new_iota_treasury_cap`, `new_iota_system_state`, `new_staked_iota`, `new_timelocked_staked_iota`, `new_coin_metadata`, `new_treasury_cap`, `new_coin_manager`, `new_balance`, `new_option`, `new_string` / `new_ascii_string`, `new_receiving`, `new_setting`, `new_wrapper`, `new_config_key` / `new_address_key` / `new_global_pause_key`, `new_display_created` / `new_display_version_updated`, `new_time_lock`, `new_system_epoch_info_event`, `new_iota_system_admin_cap`, `new_config`, …). Constructors are generated by three internal macros (`add_struct_tag_ctor`, `add_struct_tag_ctor_from_struct_tag`, `add_struct_tag_ctor_from_type_tag`) so adding new framework types is a one-line change.
- Add `StructTag::is_same_type_as`, `coin_type` / `coin_type_opt`, `type_params_mut`, `into_parts`, and `to_canonical_string(with_prefix)`. `Display` now uses short-hex addresses; `to_canonical_string` is the way to get full / `0x`-prefixed output.
- Expand `TypeTag` with primitive constructors (`TypeTag::u8()`, `bool()`, `address()`, …), `is_*` predicates via `def_is!`, ref/owned accessors (`as_vector_type_tag` / `as_struct_tag` and their `_opt` / `into_*` siblings), and `to_canonical_string(with_prefix)`.
- Rewrite the type-tag / struct-tag / identifier parser on top of `winnow`. Enforces `MAX_IDENTIFIER_LENGTH = 128` and `MAX_TYPE_TAG_NESTING = 16`, and tolerates whitespace and newlines inside generics. New `TypeParseError` covers nesting, identifier-length, and address-parse failures.
- Reorganize address / object-id constants:
  - [crates/iota-sdk-types/src/address.rs](crates/iota-sdk-types/src/address.rs): add `SYSTEM_STATE` (`0x5`), `CLOCK`, `AUTHENTICATOR_STATE`, `RANDOMNESS_STATE`, `GENESIS_IOTA_BRIDGE`, `DENY_LIST` consts.
  - [crates/iota-sdk-types/src/object_id.rs](crates/iota-sdk-types/src/object_id.rs): rename `STD_PACKAGE` → `STD`, `FRAMEWORK_PACKAGE` → `FRAMEWORK`, `SYSTEM_PACKAGE` → `SYSTEM`, `GENESIS_BRIDGE_PACKAGE` → `GENESIS_BRIDGE`, `STARDUST_PACKAGE` → `STARDUST`. The previous `ObjectId::SYSTEM` (system-state object at `0x5`) becomes `ObjectId::SYSTEM_STATE` to disambiguate from the system-package object id. All constants now derive from their `Address` counterparts.

## Breaking changes

- `IdentifierRef` is removed; use `Identifier::from_static` (const) or `Identifier::new` / `Identifier::new_unchecked` instead.
- `ObjectId` constants renamed (`*_PACKAGE` suffixes dropped). `ObjectId::SYSTEM` now refers to the system **package** (`0x3`); the system-state object is `ObjectId::SYSTEM_STATE` (`0x5`).
- `StructTag::new_iota_coin_type()` is replaced by `StructTag::new_gas()`.
- The top-level `iota_sdk::types::type_tag` module is removed; import from `iota_sdk::types` (re-exports) or `iota_sdk::types::move_core`.


crates/iota-sdk-types/src/move_core/identifier.rs
```diff
@@ -0,0 +1,247 @@
...
+        self.as_str().is_empty()
+    }
+
+    /// Returns `true` if the provided string is a valid Move identifier. Valid
+    /// identifiers must start with an alphabetic character (a-z, A-Z) or an
+    /// underscore (_), and may be followed by up to 127 alphanumeric
+    /// characters (a-z, A-Z, 0-9) or underscores (_). The maximum length of
+    /// an identifier is 128 characters.
+    ///
+    /// Note: this function allows the special identifier `<SELF>`.
```
Thoralf-M, line 94: This seems a bit inconsistent, where is this needed?
```
2. Identifier::is_valid and Identifier::new disagree on <SELF>. 

Identifier::is_valid("<SELF>") returns true (and the doc says so).
Identifier::from_static("<SELF>") succeeds.
Identifier::new("<SELF>") fails — it routes through FromStr → parse_identifier, which only accepts [A-Za-z]… or _….
"<SELF>".parse::<Identifier>() fails.
Identifier's Deserialize uses DisplayFromStr, so an Identifier("<SELF>") will serialize but fail to round-trip through any serde format, including BCS
````

File: crates/iota-sdk-types/src/move_core/identifier.rs [hunk 2]
```diff
@@ -0,0 +1,247 @@
...
+    pub const SYSTEM_EPOCH_INFO_EVENT: Self = Self::from_static("SystemEpochInfoEvent");
+    pub const TIME_LOCK: Self = Self::from_static("TimeLock");
+    pub const TIMELOCKED_STAKED_IOTA: Self = Self::from_static("TimelockedStakedIota");
+    pub const TREASURY_CAP: Self = Self::from_static("TreasuryCap");
+    pub const TX_CONTEXT: Self = Self::from_static("TxContext");
+    pub const UID: Self = Self::from_static("UID");
+    pub const UPGRADE_CAP: Self = Self::from_static("UpgradeCap");
+    pub const UPGRADE_RECEIPT: Self = Self::from_static("UpgradeReceipt");
+    pub const UPGRADE_TICKET: Self = Self::from_static("UpgradeTicket");
+    pub const URL_TYPE: Self = Self::from_static("Url");
```
Thoralf-M: Should this just be URL so it matches the string?
```suggestion
    pub const URL: Self = Self::from_static("Url");
```

crates/iota-sdk-types/src/move_core/struct_tag.rs
```diff
@@ -0,0 +1,441 @@
...
+            module: Identifier::NAME_MODULE,
+            name: Identifier::NAME,
+            type_params: vec![],
+        }
+    }
+
+    pub fn is_name(&self) -> bool {
+        self.module == Identifier::NAME_MODULE
+            && self.name == Identifier::NAME
+            && self.type_params.is_empty()
```
Thoralf-M, line 152: Correct for this type that the self.address is not checked as in the other fns?

crates/iota-sdk-types/src/move_core/identifier.rs
Thoralf-M: Was intended and <SELF> support should be removed in the future

DaughterOfMars: So should I just remove this then?

crates/iota-sdk-types/src/move_core/identifier.rs
DaughterOfMars: good catch

crates/iota-sdk-types/src/move_core/struct_tag.rs
DaughterOfMars: Yeah because what address would we use?

crates/iota-sdk-types/src/move_core/identifier.rs
Thoralf-M: No, just in the future

crates/iota-sdk-types/src/move_core/struct_tag.rs
Thoralf-M: Ok it's for an iota names name, I thought it would be a type from the iota framework or similar.
Please add doc comments above this fn and the new_name() one then explaining that they are for iota names
Also not sure if the is_name() should also accept the address then, otherwise other packages could also have modules and types that match this one, but are something different

DaughterOfMars: done

crates/iota-sdk-types/src/move_core/parse.rs
```diff
@@ -0,0 +1,336 @@
...
+        separated(
+            1..,
+            delimited(multispace0, parse_type_tag_impl(depth + 1), multispace0),
+            ",",
+        )
+        .parse_next(input)
+    }
+}
+
+// TODO add proptests
```
Thoralf-M: What about this TODO?

DaughterOfMars: not sure where that came from

thibault-martinez: I'll create an issue to investigate, there has to be a reason we added it

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/issues/1098

crates/iota-sdk-types/src/move_core/type_tag.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub fn into_struct_tag_opt(self) -> Option<StructTag> {
+        if let Self::Struct(inner) = self {
+            Some(*inner)
+        } else {
+            None
+        }
+    }
+
+    pub fn into_struct_tag(self) -> StructTag {
+        self.into_struct_tag_opt().expect("not a Struct")
```
Thoralf-M, line 117: Should we add a comment that these can panic?

thibault-martinez: Sure

thibault-martinez (approved):

PR #1100, opened 2026-04-30
feat: `Owner` improvements

DaughterOfMars:
## Summary

Improve the `Owner` enum for the core implementation.

### Changes

- Add `Owner::address_or_object()` returning an `Option<&Address>` for `Address`- and `Object`-owned variants (and `None` for `Shared`/`Immutable`).
- Implement `PartialEq<Address>` and `PartialEq<ObjectId>` for `Owner` so callers can compare an owner to an address/object id directly.
- Replace the derived `Serialize`/`Deserialize` with a custom impl backed by a `ReadableOwner` shape that matches the wire format used by the core implementation:
  - `AddressOwner(Address)` / `ObjectOwner(Address)` (object id is serialized as an address since `IotaAddress` is universal).
  - `Shared { initial_shared_version: Version }` (named field instead of a tuple, drops the `ReadableDisplay` wrapper on the version).
  - `Immutable`.

### DIFF
--- a/crates/iota-sdk-types/src/object.rs
+++ b/crates/iota-sdk-types/src/object.rs
@@ -108,9 +108,9 @@ impl Owner {
 
     crate::def_is_as_into_opt!(Address, Object(ObjectId), Shared(Version));
 
-    /// Returns an address if this object is owned by an address or
+    /// Returns an `Address` if this object is owned by an address or
     /// object, and None if it is shared or immutable.
-    pub fn address(&self) -> Option<&Address> {
+    pub fn address_or_object(&self) -> Option<&Address> {
         Some(match self {
             Self::Address(address) => address,
             Self::Object(object_id) => object_id.as_address(),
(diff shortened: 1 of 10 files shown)


Thoralf-M (approved):

crates/iota-sdk-types/src/object.rs
```diff
@@ -113,6 +107,28 @@ impl Owner {
     crate::def_is!(Immutable);
 
     crate::def_is_as_into_opt!(Address, Object(ObjectId), Shared(Version));
+
+    /// Returns an address if this object is owned by an address or
+    /// object, and None if it is shared or immutable.
+    pub fn address(&self) -> Option<&Address> {
```
kodemartin: with `Owner::{Address,Object}` reserving address for representing address owner, I would expect the method `Owner::address` to only be `Some` for the `Address` variant. 

A more consistent name would thus be `Owner::address_or_object`.

DaughterOfMars: done

kodemartin (approved):

thibault-martinez (approved):

roman1e2f5p8s (approved):
in PR description, `Owner::address()` should be updated to `Owner::address_or_object()`

PR #1101, opened 2026-04-30
feat: improvements to Argument (#967)

thibault-martinez:
## Summary

- Derive `Hash` on [`Argument`](crates/iota-sdk-types/src/transaction/mod.rs) and add a `Display` impl rendering variants as `Gas`, `Input(i)`, `Result(i)`, `NestedResult(i,j)`.
- Simplify human-readable serialization: drop the `untagged` + lowercase scheme (which required a workaround `Gas` enum and a `{ result: (u16, u16) }` struct variant) in favor of an externally-tagged form that mirrors `Argument` directly.

## Serialization change (breaking)

Human-readable JSON output for `Argument` changes:

| Variant | Before | After |
| --- | --- | --- |
| `Gas` | `"gas"` | `"Gas"` |
| `Input(1)` | `{"input": 1}` | `{"Input": 1}` |
| `Result(2)` | `{"result": 2}` | `{"Result": 2}` |
| `NestedResult(3, 4)` | `{"result": [3, 4]}` | `{"NestedResult": [3, 4]}` |

Binary (BCS) serialization is unchanged.

## Test plan

- [ ] `cargo test -p iota-sdk-types transaction::serialization::tests::argument`
- [ ] Verify downstream consumers of the human-readable JSON form are updated for the new variant names

### DIFF
--- a/crates/iota-sdk-types/src/transaction/mod.rs
+++ b/crates/iota-sdk-types/src/transaction/mod.rs
@@ -933,7 +933,7 @@ impl std::fmt::Display for Argument {
             Argument::Gas => write!(f, "Gas"),
             Argument::Input(i) => write!(f, "Input({i})"),
             Argument::Result(i) => write!(f, "Result({i})"),
-            Argument::NestedResult(i, j) => write!(f, "NestedResult({i},{j})"),
+            Argument::NestedResult(i, j) => write!(f, "NestedResult({i}, {j})"),
         }
     }
 }
--- a/crates/iota-sdk-types/src/transaction/mod.rs
+++ b/crates/iota-sdk-types/src/transaction/mod.rs
@@ -907,7 +907,7 @@ pub struct Upgrade {
 /// argument-result         = %d02 u16
 /// argument-nested-result  = %d03 u16 u16
 /// ```
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


DaughterOfMars (approved):

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -927,6 +927,17 @@ pub enum Argument {
     NestedResult(u16, u16),
 }
 
+impl std::fmt::Display for Argument {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Argument::Gas => write!(f, "Gas"),
+            Argument::Input(i) => write!(f, "Input({i})"),
+            Argument::Result(i) => write!(f, "Result({i})"),
+            Argument::NestedResult(i, j) => write!(f, "NestedResult({i},{j})"),
```
roman1e2f5p8s:
```suggestion
            Argument::NestedResult(i, j) => write!(f, "NestedResult({i}, {j})"),
```
to match with how Rust normally prints tuples in Debug, but feel free to ignore

roman1e2f5p8s (approved):

crates/iota-sdk-types/src/transaction/mod.rs
thibault-martinez: We will probably have to update all snapshot again but that's fine

PR #1102, opened 2026-04-30
feat: ExecutionStatus changes for monorepo (#558)

thibault-martinez:
# ExecutionStatus changes for monorepo

## Summary

Brings `ExecutionStatus` and the related error types in line with what the monorepo needs to drop its own copies and consume `iota-sdk-types` instead.

### `ExecutionStatus`
- New constructor `ExecutionStatus::new_failure(error, command)`.
- `unwrap()` / `unwrap_err()` helpers, mirroring `Result`'s API for tests and call sites that already know which variant they expect.

### `ExecutionError`
- Derives `thiserror::Error` and provides a `Display`/`Error` message for every variant. The wording matches what the monorepo emits today, so error strings stay stable for downstream consumers (logs, RPC error bodies, etc.).
- Variants reordered so the BCS discriminants line up with the monorepo wire format: `ExecutionCancelledDueToSharedObjectCongestionV2` now sits at position 37 (after `ExecutionCancelledDueToRandomnessUnavailable`), with `InvalidLinkage` remaining at 38.
- New `ExecutionError::command_argument_error(kind, argument)` constructor — convenience for the common case of wrapping a `CommandArgumentError` with its argument index.
- Doc/typo fixes: "Attempted to **use** feature…" and "PackageId does not **match**…".
- A `WARNING` comment on the enum and on the binary serde shadow enums spelling out that the variant order is protocol-significant and that new variants must be appended.

### `CommandArgumentError`, `PackageUpgradeError`, `TypeArgumentError`
- Now derive `Hash` and `thiserror::Error`, with `#[error(...)]` messages on every variant.

### `MoveLocation`
- Now derives `Hash`.
- New `Display` impl: prints `package::module::function_name (function index N) at offset M` when the function name is known, and falls back to `package::module in function definition N at offset M` when it is not. A small `display_move_location_opt` helper renders `Option<MoveLocation>` as either the location or `UNKNOWN` for use in `#[error(...)]` strings.
- A `display_congested_objects` helper renders `Vec<ObjectId>` as a comma-separated list for the two congestion-related error messages.

### Serde shadow enums
- `BinaryExecutionStatus`, `BinaryExecutionError`, `BinaryCommandArgumentError`, and `BinaryPackageUpgradeError` get `#[serde(rename = "...")]` so the human-readable names in non-BCS formats (e.g. JSON) match the public type names rather than the internal `Binary*` shadow names. Note that `BinaryExecutionError` is renamed to `ExecutionFailureStatus` to match the monorepo.

### `Digest`
- Adds `PartialEq<[u8; 32]>` and `PartialEq<Vec<u8>>` (in both directions) so digests can be compared directly against raw byte buffers without an explicit conversion. Covered by two new unit tests.

### Tests / tooling
- New `execution_error_bcs_discriminants` snapshot test (using `insta`) that records the BCS discriminant of every `ExecutionError` variant. Reordering or inserting variants makes the test fail, which is the intended guard now that the order is part of the protocol contract. Adding a new variant at the end is a one-line `cargo insta review` accept.
- `insta = "1.42"` added as a workspace dev-dependency and pulled into `iota-sdk-types`.

## Test plan

- [ ] `cargo test -p iota-sdk-types`
- [ ] `cargo test -p iota-sdk-types --all-features`
- [ ] Confirm the snapshot at `crates/iota-sdk-types/tests/snapshots/iota_sdk_types__execution_status__tests__execution_error_bcs_discriminants.snap` matches the monorepo's expected discriminant order.


crates/iota-sdk-types/src/execution_status.rs
```diff
@@ -501,15 +699,284 @@ impl PackageUpgradeError {
...
+        let dummy_location = MoveLocation {
+            package: zero_id,
+            module: "m".parse().unwrap(),
+            function: 0,
+            instruction: 0,
+            function_name: None,
+        };
+
+        // New variants MUST be appended at the end.
+        let variants: Vec<(&str, Vec<u8>)> = vec![
```
DaughterOfMars: Couldn't we use strum or something?

crates/iota-sdk-types/src/execution_status.rs
```diff
@@ -147,119 +201,194 @@ impl ExecutionStatus {
...
     /// Certificate is cancelled due to congestion on shared objects
+    #[error("Certificate is cancelled due to congestion on shared objects: {}.", display_congested_objects(.congested_objects))]
     ExecutionCancelledDueToSharedObjectCongestion { congested_objects: Vec<ObjectId> },
-    /// Certificate is cancelled due to congestion on shared objects;
-    /// suggested gas price can be used to give this certificate more priority.
-    ExecutionCancelledDueToSharedObjectCongestionV2 {
-        congested_objects: Vec<ObjectId>,
-        suggested_gas_price: u64,
-    },
+
```
Thoralf-M:
```suggestion
```

crates/iota-sdk-types/src/execution_status.rs
thibault-martinez: No, but actually now the ABNF already ensures the BCS order so we can simply remove this test

DaughterOfMars: nice

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1102/commits/498b6612a0e1b2ceafd6af74cb0a66ea170e6f7d

DaughterOfMars (approved):

Thoralf-M (approved):

PR #1103, opened 2026-05-04
chore: Use `thiserror` for SignatureFromBytesError

DaughterOfMars:
## Description

This is the remaining changes from #977 after rebasing. SInce zklogin was removed, this is simply a small refactor.


crates/iota-sdk-types/src/execution_status.rs
Thoralf-M, line 461: Can/should we rename this field to `function_index`?

Thoralf-M (approved):
Maybe rename title to just mention `errors` as it's also using it for ExecutionError

thibault-martinez:
> Maybe rename title to just mention `errors` as it's also using it for ExecutionError

It was just not up to date

thibault-martinez (approved):

crates/iota-sdk-types/src/execution_status.rs
DaughterOfMars: I would prefer it that way also but the json would be a problem

thibault-martinez: You mean JSON or JSON-RPC? JSON-RPC would have its own struct with the original name?

DaughterOfMars: if it has its own type in the json rpc types crate entirely then that's okay

DaughterOfMars: However, it does not seem to at the moment

PR #1104, opened 2026-05-04
feat: MovePackage changes (#973)

thibault-martinez:
## Summary

Brings `MovePackage` and its companion types in line with what the monorepo needs to drop its own copy and consume `iota-sdk-types` instead. Adds the constructors, accessors, and digest helpers that the monorepo relies on, fixes the `TypeOrigin` field name to match the upstream type system (which now covers enums as well as structs), and relaxes the parser/identifier types just enough to round-trip the special `<SELF>` identifier used in module self-references.

### `MovePackage` / `TypeOrigin` / `UpgradeInfo`

- Moved `MovePackage`, `TypeOrigin`, and `UpgradeInfo` out of [crates/iota-sdk-types/src/object.rs](crates/iota-sdk-types/src/object.rs) and into the dedicated [crates/iota-sdk-types/src/move_package.rs](crates/iota-sdk-types/src/move_package.rs) module. Re-exports from the crate root are unchanged for downstream consumers.
- Renamed `TypeOrigin::struct_name` → `TypeOrigin::datatype_name` to reflect that it now identifies either a struct *or* an enum. A `#[serde(alias = "struct_name")]` is kept on the new field so existing serialized payloads continue to deserialize. Doc comments updated to match.
- New `MovePackage::new(id, version, modules, max_move_package_size, type_origin_table, linkage_table) -> Result<Self, ExecutionError>` constructor that enforces the package size limit and returns `ExecutionError::PackageTooBig` if exceeded.
- New methods on `MovePackage`:
  - `digest()` — computes the canonical package digest from modules + linkage-table dependencies.
  - `compute_digest_for_modules_and_deps(modules, object_ids)` — the underlying helper, exposed as an associated function so the same digest logic can be shared with on-chain computation.
  - `get_module(name)` — fetches a serialized module by [`Identifier`](crates/iota-sdk-types/src/move_core/identifier.rs).
  - `size()` — returns the in-memory size in bytes (matching the monorepo's accounting for the package size limit check).
  - `id()`, `version()`, `increment_version()`, `decrement_version()`.
  - `serialized_module_map()`, `type_origin_table()`, `linkage_table()` — borrowing accessors over the underlying maps/vecs.
  - `type_origin_map()` — returns the `type_origin_table` flattened into a `BTreeMap<(Identifier, Identifier), ObjectId>` keyed by `(module_name, datatype_name)`.

### Serialization change (breaking, human-readable form only)

- Dropped the `ReadableDisplay` serde wrapper on the `version` fields of `MovePackage` and `UpgradeInfo`. BCS encoding is unchanged; in human-readable formats (e.g. JSON) `version` is now emitted as the underlying number rather than a quoted decimal string.

### `Version`

- New `Version::OBJECT_START` constant (= 1) — the version that a freshly-created object starts at, used as a default in places that need to mint a brand-new package.

### `Identifier`

- Added three `PartialEq` impls — `PartialEq<&str> for Identifier`, `PartialEq<String> for Identifier`, and `PartialEq<Identifier> for str` — so identifiers can be compared directly against string literals and `String`s in either direction without an explicit `.as_str()`.

### Move type parser

- The identifier parser now accepts the literal `<SELF>` as a valid identifier in addition to the regular `[a-zA-Z][a-zA-Z0-9_]*` / `_[a-zA-Z0-9_]+` shapes. This is the special name used by the bytecode for self-references, and was previously rejected by `parse_identifier`. Covered by a new `test_parse_self_identifier` unit test that exercises both `0x1::<SELF>::Foo` and `0x1::Foo::<SELF>` as valid `StructTag`s.

### FFI

- [crates/iota-sdk-ffi/src/types/object.rs](crates/iota-sdk-ffi/src/types/object.rs): `TypeOrigin.struct_name` → `TypeOrigin.datatype_name` to match the type-system change, with the conversions in both directions and the doc comments updated. The Go, Kotlin, and Python bindings are regenerated to match.

## Test plan

- [ ] `cargo test -p iota-sdk-types`
- [ ] `cargo test -p iota-sdk-types --all-features`
- [ ] `cargo check --workspace`
- [ ] Regenerate / smoke-test the Go, Kotlin, and Python bindings against the new `datatype_name` field name on `TypeOrigin`.

### DIFF
--- a/crates/iota-sdk-ffi/src/types/object.rs
+++ b/crates/iota-sdk-ffi/src/types/object.rs
@@ -392,7 +392,7 @@ impl ObjectData {
 pub struct TypeOrigin {
     /// The name of the module the data type resides in.
     pub module_name: Arc<Identifier>,
-    // The name of the data type. Either refers to an enum or a struct
+    /// The name of the data type. Either refers to an enum or a struct
     /// identifier.
     pub datatype_name: Arc<Identifier>,
     /// ID of the package, where the given type first appeared.
--- a/crates/iota-sdk-ffi/src/types/object.rs
+++ b/crates/iota-sdk-ffi/src/types/object.rs
@@ -377,7 +377,9 @@ impl ObjectData {
     }
 }
 
-/// Identifies a struct and the module it was defined in
+/// Stores the origin of a data type where it first appeared in the version
+/// chain. A data type is identified by the name of the module and the name of
+/// the struct/enum in combination.
 ///
 /// # BCS
 ///
@@ -388,8 +390,12 @@ impl ObjectData {
 /// ```
 #[derive(uniffi::Record)]
 pub struct TypeOrigin {
+    /// The name of the module the data type resides in.
     pub module_name: Arc<Identifier>,
-    pub struct_name: Arc<Identifier>,
+    // The name of the data type. Either refers to an enum or a struct
+    /// identifier.
+    pub datatype_name: Arc<Identifier>,
+    /// ID of the package, where the given type first appeared.
     pub package: Arc<ObjectId>,
 }
 
@@ -397,7 +403,7 @@ impl From<iota_sdk::types::TypeOrigin> for TypeOrigin {
     fn from(value: iota_sdk::types::TypeOrigin) -> Self {
         Self {
             module_name: Arc::new(value.module_name.into()),
-            struct_name: Arc::new(value.struct_name.into()),
+            datatype_name: Arc::new(value.datatype_name.into()),
             package: Arc::new(value.package.into()),
         }
     }
@@ -407,7 +413,7 @@ impl From<TypeOrigin> for iota_sdk::types::TypeOrigin {
     fn from(value: TypeOrigin) -> Self {
         Self {
             module_name: value.module_name.0.clone(),
-            struct_name: value.struct_name.0.clone(),
+            datatype_name: value.datatype_name.0.clone(),
             package: **value.package,
         }
     }
@@ -424,7 +430,7 @@ impl From<TypeOrigin> for iota_sdk::types::TypeOrigin {
 /// ```
 #[derive(uniffi::Record)]
 pub struct UpgradeInfo {
-    /// Id of the upgraded packages
+    /// ID of the upgraded package
     pub upgraded_id: Arc<ObjectId>,
     /// Version of the upgraded package
     pub upgraded_version: Arc<Version>,
--- a/crates/iota-sdk-types/src/move_package.rs
+++ b/crates/iota-sdk-types/src/move_package.rs
@@ -69,29 +69,12 @@ pub struct MovePackageData {
 impl MovePackageData {
     #[cfg(feature = "hash")]
     pub fn new(modules: Vec<Vec<u8>>, dependencies: Vec<ObjectId>) -> Self {
-        let mut components = dependencies
-            .iter()
-            .map(|o| o.into_bytes())
-            .chain(modules.iter().map(|module| {
-                let mut hasher = Hasher::new();
-                hasher.update(module);
-                hasher.finalize().into_inner()
-            }))
-            .collect::<Vec<_>>();
-
-        // Sort so the order of the modules and the order of the dependencies
-        // does not matter.
-        components.sort();
-
-        let mut hasher = Hasher::new();
-        for c in components {
-            hasher.update(c);
-        }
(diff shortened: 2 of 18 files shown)


crates/iota-sdk-ffi/src/types/object.rs
```diff
@@ -388,16 +390,20 @@ impl ObjectData {
 /// ```
 #[derive(uniffi::Record)]
 pub struct TypeOrigin {
+    /// The name of the module the data type resides in.
     pub module_name: Arc<Identifier>,
-    pub struct_name: Arc<Identifier>,
+    // The name of the data type. Either refers to an enum or a struct
```
Thoralf-M:
```suggestion
    /// The name of the data type. Either refers to an enum or a struct
```

Thoralf-M: and then `make bindings`

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -90,6 +96,258 @@ impl MovePackageData {
...
+            return Err(ExecutionError::PackageTooBig {
+                object_size,
+                max_object_size: max_move_package_size,
+            });
+        }
+        Ok(pkg)
+    }
+
+    /// Calculate the digest of the [MovePackage].
+    pub fn digest(&self) -> Digest {
```
Thoralf-M, line 221: Should also be behind `#[cfg(feature = "hash")]` as it's calling `compute_digest_for_modules_and_deps()` which is feature gated

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -90,6 +96,258 @@ impl MovePackageData {
...
+
+        // NB: sorting so the order of the modules and the order of the dependencies
+        // does not matter.
+        components.sort();
+
+        let mut digest = Hasher::new();
+        for c in components {
+            digest.update(c);
+        }
+        digest.finalize()
```
Thoralf-M, line 256: This is the same as in `MovePackageData::new()` should share this code

crates/iota-sdk-types/src/move_package.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1104/commits/e1d27b7257667a16564d02eaf6d742c35cb8c07f

crates/iota-sdk-ffi/src/types/object.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1104/commits/53547e636ba23404967490b29811aede3f55ce1e

crates/iota-sdk-types/src/move_package.rs
Thoralf-M: No

thibault-martinez: Chill, posted to the wrong thread

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1104/commits/309a9d6cf9943ec912a650d7701c211edfa78353

Thoralf-M: Yes

Thoralf-M: Thanks

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -62,31 +69,266 @@ pub struct MovePackageData {
...
-            dependencies,
-            digest: Digest::from(hasher.finalize().into_inner()),
-        }
+    /// Retrieve the module from this package with the given [Identifier].
+    pub fn get_module(&self, name: &Identifier) -> Option<&Vec<u8>> {
+        self.modules.get(name)
+    }
+
+    /// Return the size of the package in bytes
+    pub fn size(&self) -> usize {
```
Thoralf-M, line 265: Were there no unit tests in the iota repo for this? Also for the digest might be good to have one

thibault-martinez: I think there were none

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/issues/1108

Thoralf-M (approved):

PR #1109, opened 2026-05-04
feat(types): add MovePackage tests

### LINKED ISSUE #1108: Add tests for MovePackage::{digest, size}
Were there no unit tests in the iota repo for this? Also for the digest might be good to have one

_Originally posted by @Thoralf-M in https://github.com/iotaledger/iota-rust-sdk/pull/1104#discussion_r3181567589_

Thoralf-M:
Add MovePackage tests to make sure future changes don't break anything

Fixes https://github.com/iotaledger/iota-rust-sdk/issues/1108


crates/iota-sdk-types/src/move_package.rs
```diff
@@ -413,3 +413,181 @@ mod tests {
...
+    }
+
+    #[cfg(feature = "hash")]
+    mod digest {
+        use super::*;
+
+        #[test]
+        fn digest_matches_pinned_hex() {
+            const EXPECTED_HEX: &str =
+                "087322ae613945320a83dce5d45626eb3a112a0d98ed6885788635194f2d566f";
```
thibault-martinez, line 522: Was this generated by the test itself or did you get it somewhere else?

DaughterOfMars (approved):

crates/iota-sdk-types/src/move_package.rs
Thoralf-M: Generated by the test

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -427,3 +427,181 @@ mod tests {
         assert_eq!(json_package.digest, package.digest);
     }
 }
+
+#[cfg(test)]
+mod move_package_tests {
```
thibault-martinez: I would personally add to the existing module and move the serde feature but feel free to resolve

crates/iota-sdk-types/src/move_package.rs
```diff
@@ -427,3 +427,181 @@ mod tests {
...
+                upgrade_info(ObjectId::from_u16(2), 1),
+            )],
+        );
+        let modules_size =
+            (mod_a_name.len() + mod_a_bytes.len()) + (mod_b_name.len() + mod_b_bytes.len());
+        let type_origin_size = type_module_name.len() + type_datatype_name.len() + ObjectId::LENGTH;
+        let linkage_size = ObjectId::LENGTH + ObjectId::LENGTH + std::mem::size_of::<Version>();
+        let expected =
+            std::mem::size_of::<Version>() + modules_size + type_origin_size + linkage_size;
+        assert_eq!(pkg.size(), expected);
```
thibault-martinez, line 487: Is that the exact on-chain byte size? Things like strings don't incur a length byte?

crates/iota-sdk-types/src/move_package.rs
thibault-martinez: Not ideal but alright

thibault-martinez (approved):

crates/iota-sdk-types/src/move_package.rs
Thoralf-M: I checked but this digest is not pushed on chain somewhere so we can reference it, but during package upgrade it's used and that works fine with the SDK

crates/iota-sdk-types/src/move_package.rs
Thoralf-M: Done — folded into the existing `tests` module and moved the `serde` gate onto just the serde tests.

crates/iota-sdk-types/src/move_package.rs
Thoralf-M: It's not the exact BCS-serialized size — `size()` is the node's package-size heuristic used for the `max_move_package_size` check. It sums raw content bytes only (identifier + module bytes, `ObjectId::LENGTH` per id, `size_of::<Version>()`), so the BCS length prefixes on the strings/vectors and outer collections aren't counted. The test mirrors that formula on purpose so it trips if `size()` ever changes.

thibault-martinez: Makes sense, thanks

crates/iota-sdk-types/src/move_package.rs
thibault-martinez: Yeah, I was suggesting we should have computed one in the monorepo back then and use it, but alles gut

PR #1114, opened 2026-05-05
feat: `ConsensusCommitPrologueV1` improvements

Alex6323:
## Summary

Continues the SDK replacement series, stacked on top of `sdk-replacement/17-GasPayment`. Updates `ConsensusCommitPrologueV1` and its related types to match upstream's serialization layout, derives `Hash` to bring them in line with the other transaction types, and adds a small constructor.

- **`Hash`** derived on `ConsensusCommitPrologueV1`, `ConsensusDeterminedVersionAssignments`, `CancelledTransaction`, and `VersionAssignment` in [crates/iota-sdk-types/src/transaction/mod.rs](crates/iota-sdk-types/src/transaction/mod.rs).
- **`VersionAssignment::new`** constructor added in [crates/iota-sdk-types/src/transaction/mod.rs](crates/iota-sdk-types/src/transaction/mod.rs).
- **Custom `Serialize`/`Deserialize`** impls for `VersionAssignment` and `CancelledTransaction` (replacing the previously derived ones), and `ConsensusDeterminedVersionAssignments` switched from struct-style to tuple-style enum variants for both human-readable and binary representations, in [crates/iota-sdk-types/src/transaction/serialization.rs](crates/iota-sdk-types/src/transaction/serialization.rs).

### DIFF
--- a/crates/iota-sdk-types/src/transaction/serialization.rs
+++ b/crates/iota-sdk-types/src/transaction/serialization.rs
@@ -261,36 +261,34 @@ mod end_of_epoch {
 
 mod version_assignments {
     use super::*;
-    use crate::transaction::{CancelledTransaction, ConsensusDeterminedVersionAssignments};
+    use crate::transaction::{
+        CancelledTransaction, ConsensusDeterminedVersionAssignments, VersionAssignment,
+    };
 
     #[derive(serde::Serialize)]
-    #[serde(tag = "kind", rename_all = "snake_case")]
+    #[serde(rename = "ConsensusDeterminedVersionAssignments")]
     enum ReadableConsensusDeterminedVersionAssignmentsRef<'a> {
-        CancelledTransactions {
-            cancelled_transactions: &'a Vec<CancelledTransaction>,
-        },
+        CancelledTransactions(&'a Vec<CancelledTransaction>),
     }
 
+    /// Uses an enum to allow for future expansion of the
+    /// ConsensusDeterminedVersionAssignments.
     #[derive(serde::Deserialize)]
-    #[serde(tag = "kind", rename_all = "snake_case")]
+    #[serde(rename = "ConsensusDeterminedVersionAssignments")]
     enum ReadableConsensusDeterminedVersionAssignments {
-        CancelledTransactions {
-            cancelled_transactions: Vec<CancelledTransaction>,
-        },
+        CancelledTransactions(Vec<CancelledTransaction>),
     }
 
     #[derive(serde::Serialize)]
+    #[serde(rename = "ConsensusDeterminedVersionAssignments")]
     enum BinaryConsensusDeterminedVersionAssignmentsRef<'a> {
-        CancelledTransactions {
-            cancelled_transactions: &'a Vec<CancelledTransaction>,
-        },
+        CancelledTransactions(&'a Vec<CancelledTransaction>),
     }
 
     #[derive(serde::Deserialize)]
+    #[serde(rename = "ConsensusDeterminedVersionAssignments")]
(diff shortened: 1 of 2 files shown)


crates/iota-sdk-types/src/transaction/serialization.rs
```diff
@@ -261,36 +261,34 @@ mod end_of_epoch {
...
 mod version_assignments {
     use super::*;
-    use crate::transaction::{CancelledTransaction, ConsensusDeterminedVersionAssignments};
+    use crate::transaction::{
+        CancelledTransaction, ConsensusDeterminedVersionAssignments, VersionAssignment,
+    };
 
     #[derive(serde::Serialize)]
-    #[serde(tag = "kind", rename_all = "snake_case")]
+    #[serde(rename = "ConsensusDeterminedVersionAssignments")]
```
Thoralf-M, line 269: Correct without `Ref` at the end?

DaughterOfMars: yes

DaughterOfMars (approved):

piotrm50 (approved):

thibault-martinez (approved):

crates/iota-sdk-types/src/transaction/serialization.rs
```diff
@@ -261,36 +261,34 @@ mod end_of_epoch {
...
     enum ReadableConsensusDeterminedVersionAssignmentsRef<'a> {
-        CancelledTransactions {
-            cancelled_transactions: &'a Vec<CancelledTransaction>,
-        },
+        CancelledTransactions(&'a Vec<CancelledTransaction>),
     }
 
+    /// Uses an enum to allow for future expansion of the
+    /// ConsensusDeterminedVersionAssignments.
```
roman1e2f5p8s, line 275: why is this docstring added only here? it applies to all 4 helpers. either remove (preferrable) it, or add to all.

roman1e2f5p8s (commented):
the PR description says `ConsensusDeterminedVersionAssignments` "switched from struct-style to tuple-style enum variants for both human-readable and binary representations", but the impl still keeps four parallel helper types (`Readable*`, `Readable*Ref`, `Binary*`, `Binary*Ref`) and an `is_human_readable()` dispatch in both `Serialize` and `Deserialize`

as far as I can tell the readable and binary helpers now have identical serde schemas, so both branches produce the same outputs - is there a reason to keep the split, or should the helpers and the dispatch collapse to a single path?

crates/iota-sdk-types/src/transaction/serialization.rs
Alex6323: Only here not for the user but for the openrpc spec generation.

Alex6323: you'll find that exact string in the description of this type in the `openrpc.json` file in the monorepo.

Alex6323:
> the PR description says `ConsensusDeterminedVersionAssignments` "switched from struct-style to tuple-style enum variants for both human-readable and binary representations", but the impl still keeps four parallel helper types (`Readable*`, `Readable*Ref`, `Binary*`, `Binary*Ref`) and an `is_human_readable()` dispatch in both `Serialize` and `Deserialize`
> 
> as far as I can tell the readable and binary helpers now have identical serde schemas, so both branches produce the same outputs - is there a reason to keep the split, or should the helpers and the dispatch collapse to a single path?

You are right in that the split isn't required and we can collapse them. I will create an issue for this.

Alex6323:
@roman1e2f5p8s I created this issue to deal with the redundancy at some point in the future: https://github.com/iotaledger/iota-rust-sdk/issues/1117

PR #1115, opened 2026-05-05
feat(iota-sdk-types)!: add MoveObjectType

Thoralf-M:
## Summary

Continues the SDK replacement series, stacked on top of `sdk-replacement/18-ConsensusCommitPrologueV1`. Introduces a `MoveObjectType` newtype that owns the compact BCS encoding for object types, and adds a few `StructTag` constructors/predicates for timelocked balances and coin-metadata types.

- **New `MoveObjectType` newtype** in [crates/iota-sdk-types/src/object.rs](crates/iota-sdk-types/src/object.rs): wraps `StructTag` and carries the compact `Other / GasCoin / StakedIota / Coin` BCS encoding directly on the type. Provides `new`, `into_inner`, `Deref<Target = StructTag>`, `From`/`Into` between `StructTag` and `MoveObjectType`, `PartialEq<StructTag>`, `Display`, and `FromStr`. Re-exported from the crate root in [crates/iota-sdk-types/src/lib.rs](crates/iota-sdk-types/src/lib.rs).
- **`MoveStruct::type_` is now `MoveObjectType`** (was `StructTag` plus a `serde(with = BinaryMoveStructType)` adapter). The compact BCS form is intrinsic to the type, and serialization now branches on `is_human_readable()`: BCS uses the wrapped enum (unchanged), JSON serializes the underlying `StructTag` directly via its `Display`/`FromStr`. The `BinaryMoveStructType` `serde_with` adapter is removed.
- **JSON form simplified**: previously the `BinaryMoveStructType` adapter applied to both BCS and JSON, so JSON emitted the externally-tagged wrapper — `"GasCoin"`, `"StakedIota"`, `{"Coin": "<type-tag>"}`, or `{"Other": <struct-tag>}`. JSON now emits the canonical `StructTag` string in every case (e.g. `"0x2::coin::Coin<0x2::iota::IOTA>"` instead of `"GasCoin"`). BCS is unchanged. The JSON form is not used on any public API surface, so this is internal-only.
- **Internal serde rename** in [crates/iota-sdk-types/src/object.rs](crates/iota-sdk-types/src/object.rs): the binary helper enums (`MoveStructType` → `MoveObjectTypeWrapper`, `MoveStructTypeRef` → `MoveObjectTypeRef`) carry `#[serde(rename = "MoveObjectType")]` so generated schemas keep the public type name.
- **`StructTag` additions** in [crates/iota-sdk-types/src/move_core/struct_tag.rs](crates/iota-sdk-types/src/move_core/struct_tag.rs): `new_timelocked_gas_balance`, `is_timelocked_gas_balance`, `is_timelocked_balance`, plus `RegulatedCoinMetadata` and `DenyCapV1` constructors.

BCS round-trips identically. Breaking at the Rust API level for any caller that constructs `MoveStruct { type_: <StructTag>, .. }` or pattern-matches the field — wrap with `.into()` / `MoveObjectType::new(..)`.

### DIFF
--- a/crates/iota-sdk-types/src/move_core/struct_tag.rs
+++ b/crates/iota-sdk-types/src/move_core/struct_tag.rs
@@ -216,6 +216,27 @@ impl StructTag {
         (self.address, self.module, self.name, self.type_params)
     }
 
+    /// Checks if this is a timelocked coin balance `TimeLock<Balance<T>>`
+    pub fn is_timelocked_balance(&self) -> bool {
+        self.is_time_lock()
+            && matches!(
+                self.type_params.as_slice(),
+                [TypeTag::Struct(inner)] if inner.is_balance()
+            )
+    }
+
+    /// Creates a new timelocked IOTA balance struct tag
+    /// (`0x2::timelock::TimeLock<0x2::balance::Balance<0x2::iota::IOTA>>`)
+    pub fn new_timelocked_gas_balance() -> Self {
+        Self::new_time_lock(Self::new_balance(Self::new_gas()))
+    }
+
+    /// Checks if this is a timelocked IOTA balance type
+    /// (`0x2::timelock::TimeLock<0x2::balance::Balance<0x2::iota::IOTA>>`)
+    pub fn is_timelocked_gas_balance(&self) -> bool {
+        *self == Self::new_timelocked_gas_balance()
+    }
+
     /// Returns the string representation of this struct tag using the
     /// canonical display, with or without a `0x` prefix.
     pub fn to_canonical_string(&self, with_prefix: bool) -> String {
@@ -396,6 +417,7 @@ impl StructTag {
     );
     add_struct_tag_ctor!(STD, string::String);
     add_struct_tag_ctor!(@with_module STD, ascii::String);
+    add_struct_tag_ctor!(FRAMEWORK, coin::RegulatedCoinMetadata, coin::DenyCapV1);
     add_struct_tag_ctor_from_struct_tag!(
         FRAMEWORK,
         coin::CoinMetadata,
--- a/crates/iota-sdk-types/src/move_core/struct_tag.rs
+++ b/crates/iota-sdk-types/src/move_core/struct_tag.rs
@@ -225,6 +225,16 @@ impl StructTag {
             )
     }
 
+    /// Checks if this is an IOTA balance type
+    /// (`0x2::balance::Balance<0x2::iota::IOTA>`)
+    pub fn is_gas_balance(&self) -> bool {
+        self.is_balance()
+            && matches!(
+                self.type_params.as_slice(),
+                [TypeTag::Struct(inner)] if inner.is_gas()
+            )
+    }
+
     /// Creates a new timelocked IOTA balance struct tag
     /// (`0x2::timelock::TimeLock<0x2::balance::Balance<0x2::iota::IOTA>>`)
     pub fn new_timelocked_gas_balance() -> Self {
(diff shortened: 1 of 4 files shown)


DaughterOfMars (approved):

crates/iota-sdk-types/src/object.rs
```diff
@@ -586,25 +652,29 @@ mod serialization {
...
         where
             S: Serializer,
         {
-            let move_object_type = MoveStructTypeRef::from_struct_tag(source);
-            move_object_type.serialize(serializer)
+            if serializer.is_human_readable() {
+                self.0.serialize(serializer)
+            } else {
+                MoveObjectTypeRef::from_struct_tag(&self.0).serialize(serializer)
+            }
```
roman1e2f5p8s, line 665: JSON serialization form for `MoveStruct::type_` itself is changed here, but the PR description says it is unchanged.

Thoralf-M: Good catch, updated the description, but the JSON representation is not used in the API so it's fine to break it here

crates/iota-sdk-types/src/move_core/struct_tag.rs
```diff
@@ -216,6 +216,27 @@ impl StructTag {
...
+    /// (`0x2::timelock::TimeLock<0x2::balance::Balance<0x2::iota::IOTA>>`)
+    pub fn new_timelocked_gas_balance() -> Self {
+        Self::new_time_lock(Self::new_balance(Self::new_gas()))
+    }
+
+    /// Checks if this is a timelocked IOTA balance type
+    /// (`0x2::timelock::TimeLock<0x2::balance::Balance<0x2::iota::IOTA>>`)
+    pub fn is_timelocked_gas_balance(&self) -> bool {
+        *self == Self::new_timelocked_gas_balance()
+    }
```
roman1e2f5p8s, line 252:
```suggestion
    /// Checks if this is an IOTA balance type
    /// (`0x2::balance::Balance<0x2::iota::IOTA>`)
    pub fn is_gas_balance(&self) -> bool {
        self.is_balance()
            && matches!(
                self.type_params.as_slice(),
                [TypeTag::Struct(inner)] if inner.is_gas()
            )
    }

    /// Checks if this is a timelocked IOTA balance type
    /// (`0x2::timelock::TimeLock<0x2::balance::Balance<0x2::iota::IOTA>>`)
    pub fn is_timelocked_gas_balance(&self) -> bool {
        self.is_time_lock()
            && matches!(
                self.type_params.as_slice(),
                [TypeTag::Struct(inner)] if inner.is_gas_balance()
            )
    }
```
how about this? similar to `is_timelocked_balance()` above and avoiding allocations? but I am not sure whether `is_gas_balance()` should be pub exposed

crates/iota-sdk-types/src/object.rs
```diff
@@ -173,6 +173,74 @@ impl ObjectData {
     crate::def_is_as_into_opt!(Struct(MoveStruct), Package(MovePackage));
 }
 
+/// A [`StructTag`] with optimized BCS serialization for object types.
+///
+/// GasCoin, StakedIota, and Coin variants use compact enum encoding
+/// instead of the full StructTag representation.
```
roman1e2f5p8s:
```suggestion
/// GasCoin, StakedIota, and Coin variants use compact enum encoding
/// instead of the full StructTag representation. The Other variant
// carries the full StructTag inline.
```

roman1e2f5p8s (approved):

crates/iota-sdk-types/src/object.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1115/commits/a64868db73f8ed4ff955e7f23f1a55c02add8ec4

crates/iota-sdk-types/src/move_core/struct_tag.rs
Thoralf-M: It's not needed at the moment, but also no harm in adding it, makes sense if there is also the timelocked variant I guess, added in https://github.com/iotaledger/iota-rust-sdk/pull/1115/commits/a64868db73f8ed4ff955e7f23f1a55c02add8ec4

PR #1118, opened 2026-05-05
feat: Command improvements

thibault-martinez:
## Summary

Continues the SDK replacement series, stacked on top of `sdk-replacement/19-MoveObjectType`. Brings `Command` and its variants in line with the other transaction types: derives `Hash`, adds ergonomic constructors, and provides `Display` impls so commands can be rendered in logs and error messages.

- **`Hash`** derived on `Command`, `MoveCall`, `TransferObjects`, `SplitCoins`, `MergeCoins`, `Publish`, `MakeMoveVector`, and `Upgrade` in [crates/iota-sdk-types/src/transaction/mod.rs](crates/iota-sdk-types/src/transaction/mod.rs).
- **Constructors** added on `Command` — `new_move_call`, `new_transfer_objects`, `new_split_coins`, `new_merge_coins`, `new_publish`, `new_make_move_vector`, `new_upgrade` — to avoid having to name each inner struct at the call site, in [crates/iota-sdk-types/src/transaction/mod.rs](crates/iota-sdk-types/src/transaction/mod.rs).
- **`Display`** impls for `Command` and each variant, plus a small `write_sep` helper for rendering delimited lists, in [crates/iota-sdk-types/src/transaction/mod.rs](crates/iota-sdk-types/src/transaction/mod.rs).
- **`PartialEq<Identifier> for String`** added to mirror the existing `PartialEq<Identifier> for str` and `PartialEq<String> for Identifier` impls, in [crates/iota-sdk-types/src/move_core/identifier.rs](crates/iota-sdk-types/src/move_core/identifier.rs).
- **`#[serde(rename = "Command")]`** added to the internal `BinaryCommandRef` / `BinaryCommand` helpers so the BCS schema name matches the public type, in [crates/iota-sdk-types/src/transaction/serialization.rs](crates/iota-sdk-types/src/transaction/serialization.rs).
- **`#[serde(alias = "struct_name")]`** on `TypeOrigin::datatype_name` gated behind `#[cfg_attr(feature = "serde", ...)]` so the type compiles without the `serde` feature, in [crates/iota-sdk-types/src/move_package.rs](crates/iota-sdk-types/src/move_package.rs).

### DIFF
--- a/crates/iota-sdk-types/src/transaction/mod.rs
+++ b/crates/iota-sdk-types/src/transaction/mod.rs
@@ -913,7 +913,7 @@ impl SharedObjectReference {
 /// command-make-move-vector    = %d05 make-move-vector
 /// command-upgrade             = %d06 upgrade
 /// ```
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
 #[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
 #[non_exhaustive]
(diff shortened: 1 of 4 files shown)


DaughterOfMars (approved):

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -959,6 +959,201 @@ impl Command {
...
+        Command::Upgrade(Upgrade {
+            modules,
+            dependencies,
+            package,
+            ticket,
+        })
+    }
+}
+
+pub fn write_sep<T: core::fmt::Display>(
```
Alex6323, line 1030: Add docs?

Thoralf-M (approved):

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -1024,6 +1024,9 @@ impl Command {
     }
 }
 
+/// Writes `items` to `f`, separated by `sep` and optionally wrapped in
```
Alex6323, line 1027: don't like that Claude comment that much, but better than nothing 😂

Alex6323 (approved):

crates/iota-sdk-types/src/transaction/mod.rs
thibault-martinez: It's not really meant to be consumed anyway

Alex6323: what do you mean? it's consumed by the monorepo, no?

thibault-martinez: Yeah in the monorepo as a helper, but not really for users

PR #1124, opened 2026-05-06
feat: add `iota-sdk-move-types` crate

### LINKED ISSUE #573: Add Rust versions of Move system package types
Have all the types of the move system packages
```Rust
    pub const STD: Self = Self::from_u16(1);
    pub const FRAMEWORK: Self = Self::from_u16(2);
    pub const SYSTEM: Self = Self::from_u16(3);
    pub const STARDUST: Self = Self::from_u16(0x107a);
    pub const SYSTEM_STATE: Self = Self::from_u16(5);
```
defined in Rust with JSON/BCS de/serialization
Monorepo already contains many types:
https://github.com/search?q=repo%3Aiotaledger%2Fiota+%22%2F%2F%2F+Rust+version+of%22+language%3ARust&type=code

Alex6323:
## Summary

New crate `iota-sdk-move-types` providing Rust mirrors of every Move system type used by the IOTA blockchain. Each top-level module corresponds to a system package:

- `std` — `0x1`, the Move standard library
- `framework` — `0x2`, the IOTA framework
- `iota_system` — `0x3`, the IOTA system package
- `stardust` — `0x107a`, the Stardust migration package

Inside each package, every Move source module is mirrored 1:1 as a Rust `pub mod`. Generic Move types stay generic in Rust (with `PhantomData<T>` for phantom parameters).

Feature flags:

- `serde` — derive `Serialize`/`Deserialize`, plus `try_from_object` constructors that validate the full on-chain type tag before BCS-decoding the contents. For generic mirrors this includes the coin marker: the `MoveType` trait (`fn type_tag() -> TypeTag`, implemented for `IOTA`; a one-line impl for third-party coins) ties `T` to the tag, so a `BasicOutput<OTHER>` can no longer silently decode as `BasicOutput<IOTA>` — the marker is phantom, the bytes are identical, only the tag differs. `try_from_object_with_type` is the escape hatch for coin types only known at runtime.
- `bcs-schema` — generate `bcs-schema.abnf`, an ABNF description of the wire format.

All key-ability mirrors with a clear consumer are wired into the FFI bindings (20 shims): `StakedIota`, `TimelockedStakedIota`, `IotaSystemState`, `IotaSystemStateV2`, `IotaCoinMetadata`, `Clock`, `TimelockedIotaBalance` (`TimeLock<Balance<IOTA>>`), `UpgradeCap`, `Publisher`, `Kiosk`, `KioskOwnerCap`, and the Stardust types (`Alias`, `Nft`, `Irc27Metadata`, `BasicOutput`, `NftOutput`, `AliasOutput`, plus the unlock-condition records). `Coin` is intentionally not duplicated — the existing type-erased `Coin` binding already decodes any coin type. Every remaining mirror is accounted for in #1207. A `decode_staked_iota` example ships for Rust and every binding language, demonstrating the contents-decoding workflow.

The crate is WASM-compatible: the library only depends on `iota-types`, `bcs`, and `serde`, and the full 64-test suite runs under `wasm32-unknown-unknown` via `make wasm` (`wasm-pack test --node`).

Closes https://github.com/iotaledger/iota-rust-sdk/issues/573

## Testing strategy

Four layers, all driven from a single `cargo test -p iota-sdk-move-types --features serde` (64 tests):

1. **Structural shape cross-check** (`move_shape_compare::shapes_match`, 180 entries). Each Rust mirror carries a `#[derive(MoveShape)]` (new derive in `iota-sdk-bcs-schema` under the `move-shape` feature) that emits a structural fingerprint at compile time, which the harness compares against the Move definition parsed from the compiled package bytecode — catching renamed, reordered, or retyped fields. A completeness guard (`registry_matches_published_api`) enforces that **every** public type in the upstream manifest has a registered mirror, so the check is exhaustive by construction. A set of negative tests mutates known-good shapes one element at a time to prove the comparator rejects each kind of drift (it can't silently regress to always-`Ok`).
2. **Real-bytes fixture roundtrip** (`tests/fixture_roundtrip.rs`, 15 tests). Each test decodes a pinned `.bcs` capture from a real mainnet object, re-encodes, and asserts bit-equality. Covers `IotaSystemState` (whose 62 KB inner blob transitively exercises most of `iota_system::*`), `Clock`, `Coin<IOTA>`/`CoinMetadata<IOTA>`, `StakedIota`, `TimelockedStakedIota`, `Random`, `DenyList`, `UpgradeCap`, and all 5 stardust output types — roughly 25% of the registry, transitively.
3. **Type-tag validation** (`tests/try_from_object.rs`, 30 tests). Wraps fixtures (or synthetic values, for types without one) in objects carrying correct and forged type tags and asserts every `try_from_object` constructor accepts its own tag and rejects others — including the generic coin-marker matrix (wrong marker rejected, custom `MoveType` impls, the `try_from_object_with_type` runtime path) and the compositional `Balance<T>` tag used by `TimeLock<Balance<IOTA>>`.
4. **Rust-side validation** (`url::tests::url_rejects_non_ascii`), guarding the Rust-only ASCII check on `Url`.

The shape check is *static* (declared layout vs bytecode); the fixtures are *runtime* (real bytes through the actual serde path). They can fail independently — a misapplied `#[serde(...)]` attribute slips past the shape check but fails a fixture, and vice versa a missing fixture is backstopped by the shape check — so together they confirm both the declared shape and the encode/decode path.

## Compiled package artifacts (fetched, not committed)

The Move-side artifacts (the four compiled package blobs + `published_api.txt`) are **not committed**. They are fetched from the [iota monorepo](https://github.com/iotaledger/iota) into the gitignored `crates/iota-sdk-move-types/src/packages_compiled/` directory, at the monorepo rev pinned by the `move-binary-format` dev-dependency in the crate's `Cargo.toml` — the single source of truth, since that parser must match the blobs it parses. The fetch script resolves the rev via `cargo metadata` (not by parsing the TOML), so manifest formatting can't break it. The `make` targets that compile tests (`test`, `clippy`, `wasm`) fetch them automatically when missing, so CI, local runs, and the wasm suite all test the exact same bytes.

```bash
make update-compiled-packages              # at the pinned rev (default)
make update-compiled-packages REF=develop  # preview a branch
make update-compiled-packages REF=<sha>    # a specific commit
```

To refresh the on-chain BCS fixtures (committed under `tests/fixtures/`), run `cargo run -p iota-sdk --example capture_move_type_fixtures` — manual, see the crate README for details on adding new fixtures.

## Supporting a new package

To mirror a new system package (or any Move package whose types we want):

1. **Register the blob.** Add an entry to the `ARTIFACTS` array in `crates/iota-sdk-move-types/update_compiled_packages.sh`, then run `make update-compiled-packages`.
2. **Wire the blob into the comparator** (`move_shape_compare.rs`): a `const` + a `Package` enum variant + arms in `Package::blob()`/`Package::label()`.
3. **Create the Rust package module** under `src/packages/`, mirroring the Move source layout 1:1.
4. **Re-export at the crate root** (`src/packages/mod.rs` + `lib.rs`).
5. **Derive `MoveShape`** (and optionally serde) on every mirror.
6. **Register comparator entries** — one `entry!` per mirror; `cargo test` flags any drift, and the completeness guard flags anything forgotten.
7. **(Optional) add on-chain fixtures** for high-value types.

Types Move marks `#[test_only]` don't appear in the compiled bytecode — skip them with a brief comment.

## Drift detection

`.github/workflows/move_drift_nightly.yml` runs daily and diffs the *type surface* (`public struct`/`public enum` records only — function churn is filtered out) of upstream `published_api.txt` between the pinned rev and `develop` HEAD.

- A red nightly means *upstream types changed*: add/update the mirrors, bump the `move-binary-format` rev (which moves the artifact pin with it), and open a PR — that PR's CI validates the new mirrors against the new rev. `failure_slack_report.yml` posts the failure to Slack; manual runs via `workflow_dispatch`.
- Unrelated PRs are deliberately **not** affected by upstream drift: they keep testing against the pinned rev until the catch-up PR moves it.

It catches type additions, removals, renames, module moves, and `struct`↔`enum` flips. It intentionally doesn't watch field-level changes: Move's upgrade compatibility checker rejects layout changes on existing system-package types (schema evolution happens via new `*V2` types, which show up as additions).

## Test plan

- [x] `cargo test -p iota-sdk-move-types --features serde` passes (50 tests)
- [x] `cargo check -p iota-sdk-move-types` with no features / `serde` / `bcs-schema` compiles
- [x] `cargo doc -p iota-sdk-move-types --all-features` builds clean
- [x] `make wasm` builds for `wasm32` and runs the suite under node
- [x] `make update-compiled-packages` succeeds at the pinned rev and with `REF=develop`
- [x] `make test` / `make clippy` auto-fetch the artifacts on a checkout without them
- [x] "Move Drift Nightly" steps executed locally end to end and exit clean (`workflow_dispatch` is only possible once the file exists on `develop`, i.e. after merge — trigger it once then as the final check)

## Future work

- Remaining FFI exposure (situational keyed types, type-erased shims for user-generic types, event-struct decoding): tracked exhaustively in #1207
- A generic `move_objects<T: MoveType>()` client API that filters and decodes pages by `T::type_tag()` (suggested in review)


crates/iota-sdk-move-system-types/README.md
```diff
@@ -0,0 +1,35 @@
+# iota-sdk-move-system-types
+
+Rust representations of Move system types used by the IOTA blockchain.
```
Thoralf-M: Can you link to the source path for the types?
https://github.com/iotaledger/iota/tree/develop/crates/iota-framework

Maybe also list the different packages with their IDs on chain

Thoralf-M: Similar to what you have in the PR description

crates/iota-sdk/Cargo.toml
```diff
@@ -71,6 +71,7 @@ bcs.workspace = true
 cynic.workspace = true
 eyre.workspace = true
 hex.workspace = true
+iota-move-system-types = { workspace = true, features = ["serde"] }
```
thibault-martinez: Not just dev-dependency, the SDK should re-export this, under a feature like the other crates

Cargo.toml
```diff
@@ -64,6 +65,7 @@ iota-graphql-client-build = { version = "0.0.1-alpha.1", package = "iota-sdk-gra
 iota-grpc-client = { version = "0.0.1-alpha.1", package = "iota-sdk-grpc-client", path = "crates/iota-sdk-grpc-client", default-features = false }
 iota-grpc-proto-build = { version = "0.0.1-alpha.1", package = "iota-sdk-grpc-proto-build", path = "crates/iota-sdk-grpc-proto-build", default-features = false }
 iota-grpc-types = { version = "0.0.1-alpha.1", package = "iota-sdk-grpc-types", path = "crates/iota-sdk-grpc-types", default-features = false }
+iota-move-system-types = { version = "0.0.1-alpha.1", package = "iota-sdk-move-system-types", path = "crates/iota-sdk-move-system-types", default-features = false }
```
thibault-martinez: Just wondering if `system-types` is the best name, given that one of the 4 packages is called `iota-system`

crates/iota-sdk-move-types/tests/fixture_roundtrip.rs
thibault-martinez, line 1: I'm not too sure about this tbh. For roundtrips, it doesn't matter that they come from real on-chain bytes. Plus this is not exhaustive as you are missing lots of types.

crates/iota-sdk-move-system-types/README.md
Alex6323: sure

crates/iota-sdk-ffi/src/types/mod.rs
```diff
@@ -12,6 +12,7 @@ pub mod gas;
 pub mod iota_names;
 pub mod move_core;
 pub mod move_package;
+pub mod move_system_types;
```
thibault-martinez: The FFI crate modules are mirroring the crates, move this out of `types`

crates/iota-sdk-move-types/src/framework.rs
```diff
@@ -0,0 +1,3450 @@
...
+    #[derive(Debug, Clone, Eq, PartialEq)]
+    #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+    #[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
+    pub struct MetadataOwnershipRenounced {
+        pub coin_name: AsciiString,
+    }
+}
+
+#[cfg(all(test, feature = "serde"))]
+mod round4_tests {
```
thibault-martinez: What is this? What does round4 mean? Why are you having tests here when for some other modules you have tests within them?

Cargo.toml
Alex6323: hmm, maybe, but we're running out of names. We can't call it:
`iota-move-core-types` and neither `iota-move-package-types`, and `iota-system-package-types` is overly verbose, and `iota-move-types` probably too general (but could work)?

Suggestions?

crates/iota-sdk-move-types/tests/fixture_roundtrip.rs
Alex6323: yeah, it's not exhaustive, but how can we make it so? Ideas?

Cargo.toml
thibault-martinez: I think I would be fine with `iota-move-types`? After all, why would we limit this crate to system/framework stuff? We could add useful packages like IOTA-Name at some point?

crates/iota-sdk-move-types/tests/fixture_roundtrip.rs
thibault-martinez: Use proptest? Real on-chain data doesn't really add anything

crates/iota-sdk-move-types/src/framework.rs
Alex6323: Seems like an artifact. I instructed Claude to add tests in different rounds.

Cargo.toml
Alex6323: good point, `iota-move-types` it is then :+1:

crates/iota-sdk-move-types/tests/fixture_roundtrip.rs
Alex6323: I don't think so. Only real on-chain data can give us the guarantee that we actually can deserialize on-chain data. The proptest won't give you that, because it doesn't take into account structural type changes. The proptest will essentially just test the `bcs` crate.

thibault-martinez: What changes?

Alex6323: Like with the `proptests` we could from the SDK side change the types structure and have them all pass, but still if we'ld try to deserialize from the chain, it wouldn't work.

Alex6323: The bare BCS round trips are just there to prevent accidental asymmetrie changes to a type de/serialization. We could also remove them if we assume we are super careful, but I kept them because they are cheap.

thibault-martinez: I got confused because you have two types or roundtrip but actually it's the other kind that is useless

thibault-martinez: Still, I feel like there has to be a better and more exhaustive way to do this but atm idk, will think about it over the weekend

thibault-martinez:
nice PR

crates/iota-sdk-move-types/tests/fixture_roundtrip.rs
Alex6323: Did you see what I had before this morning?... you can go a few commits back to see the `moverox` solution. It is a solution that is exhaustive, but also not really, because you have to trust a single guy's crate to be correct and it pulls in the auto-generated files. We'ld have to trust this crate that: OnChainMoveTypes <-> AutoGeneratedRustTypes serialiation always matches. I think it would give us false security.

Also a lot more types are transitively checked by what we have right now. I'll add to the PR description what the coverage is atm.

crates/iota-sdk-move-types/update_compiled_packages.sh
Thoralf-M, line 1: I think I would move this file into the iota-sdk-move-types/ as it's specific to that

crates/iota-sdk-move-types/README.md
Thoralf-M, line 1: Please also add a short section about the packages_compiled files, bit hard to reason about why they exist, how they should be updated. I found the update_compiled_packages.sh script now, but in the crate itself there is basically no info about these files and how/from where they are updated...
Also https://github.com/iotaledger/iota-rust-sdk/pull/1124/#discussion_r3248964045 is not done yet, somewhat related

bindings/python/examples/decode_staked_iota.py
```diff
@@ -0,0 +1,43 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# Decode `StakedIota` objects into typed Python values via the Rust mirror in
+# `iota-sdk-move-types`.
+#
+# The GraphQL client returns each object's contents as raw BCS bytes. Without
+# a Rust mirror of the Move type, an FFI caller would have to walk those bytes
+# by hand. With the move-types crate exposed through the bindings, a
+# single `StakedIota.try_from_object(obj)` call gives typed, named-field
+# access to id / pool_id / activation_epoch / principal.
```
Thoralf-M: This comment, especially the second part, sounds more like PR description that explains the reason why iota-sdk-move-types exist, I think the comment should focus more on the example itself. `Rust` and `iota-sdk-move-types` should probably not be mentioned at all here, as for bindings it's just `iota_sdk`

crates/iota-sdk-move-types/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+mod move_shape_compare;
+
+/// Error returned by the `try_from_object` constructors on type mirrors.
+///
+/// All Tier 1 types share this error shape: the caller passed an `Object`
+/// that either isn't a Move struct, has a type tag that doesn't match the
+/// expected type, or whose BCS contents fail to decode.
+#[cfg(feature = "serde")]
+#[derive(Debug)]
+pub enum FromObjectError {
```
Thoralf-M, line 79: `#[non_exhaustive]`?

Makefile
Thoralf-M: Should we add the move types crate here?

crates/iota-sdk-move-types/bcs-schema.abnf
```diff
@@ -0,0 +1,398 @@
...
+              u64   ; stake-activation-epoch
+              u64   ; principal
+
+staking-request-event = id        ; pool-id
+                        address   ; validator-address
+                        address   ; staker-address
+                        u64       ; epoch
+                        u64       ; amount
+
+stardust_-upgrade_-label = bool   ; dummy-field
```
Thoralf-M: iota-sdk-bcs-schema/src/lib.rs:277 —`to_kebab_case` produces invalid ABNF. It doesn't treat `_` as a word boundary, so `STARDUST_UPGRADE_LABEL` generates the malformed rule `stardust_-upgrade_-label = bool`

crates/iota-sdk/examples/decode_staked_iota.rs
```diff
@@ -0,0 +1,63 @@
...
+    if page.data().is_empty() {
+        println!("No StakedIota objects on testnet right now.");
+        return Ok(());
+    }
+
+    println!("Decoded {} StakedIota object(s):\n", page.data().len());
+
+    let mut total_principal: u128 = 0;
+    for object in page.data() {
+        let staked: StakedIota = bcs::from_bytes(object.as_struct().contents())?;
```
Thoralf-M: Also use `StakedIota::try_from_object()` as in the bindings, it does validation/error handling we don't have here, `as_struct()` could panic

crates/iota-sdk/Cargo.toml
```diff
@@ -22,6 +22,7 @@ default = [
   "graphql",
   "hash",
   "mnemonic",
+  "move-types",
```
Thoralf-M: This is not reachable without
```Rust
#[cfg(feature = "move-types")]
pub use iota_move_types as move_types;
```
in lib.rs

crates/iota-sdk/examples/decode_staked_iota.rs
```diff
@@ -0,0 +1,63 @@
...
+//!
+//! - `staked.id()` — the staked object's [`ObjectId`]
+//! - `staked.pool_id()` — the staking pool the stake belongs to
+//! - `staked.activation_epoch()` — the epoch the stake activates at
+//! - `staked.principal()` — the staked amount in nanos
+//!
+//! [`ObjectId`]: iota_types::ObjectId
+
+use eyre::Result;
+use iota_move_types::iota_system::staking_pool::StakedIota;
```
Thoralf-M: Should import via `iota_sdk`

crates/iota-sdk-move-types/examples/capture_fixtures.rs
```diff
@@ -0,0 +1,304 @@
...
+/// What to capture and where to put it.
+struct Fixture {
+    /// Output file name (without extension). Written to
+    /// `tests/fixtures/<file>.bcs`.
+    file: &'static str,
+    /// How to find an instance.
+    source: Source,
+}
+
+#[allow(dead_code)]
```
Thoralf-M: Why `allow(dead_code)`?

crates/iota-sdk-move-types/src/packages/iota_system.rs
```diff
@@ -0,0 +1,900 @@
...
+
+        pub fn id(&self) -> &ObjectId {
+            self.id.object_id()
+        }
+
+        pub fn pool_id(&self) -> &ObjectId {
+            &self.pool_id.bytes
+        }
+
+        pub fn activation_epoch(&self) -> u64 {
```
Thoralf-M: Intentionally left out the `stake_` prefix here?

crates/iota-sdk/examples/decode_staked_iota.rs
```diff
@@ -0,0 +1,63 @@
...
+        .await?;
+
+    if page.data().is_empty() {
+        println!("No StakedIota objects on testnet right now.");
+        return Ok(());
+    }
+
+    println!("Decoded {} StakedIota object(s):\n", page.data().len());
+
+    let mut total_principal: u128 = 0;
```
Thoralf-M: u64 should be safe here and would be more consistent with bindings examples

crates/iota-sdk/examples/decode_staked_iota.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    let page = client
+        .objects(
+            ObjectFilter {
+                type_: Some("0x3::staking_pool::StakedIota".to_owned()),
+                ..Default::default()
+            },
+            Default::default(),
+        )
+        .await?;
```
DaughterOfMars, line 43: Question about a possible enhancement: Could we add another generic method like `fn move_objects<T: MoveObject>(...) -> Page<T>` where it will automatically filter and decode the type?

crates/iota-sdk-move-types/tests/fixture_roundtrip.rs
Alex6323: Removed the BCS round-trips and replaced them with the shape comparator which ensures that the type layout between the Move type and its Rust Mirror matches by vendoring the compiled Move packages. This is exhaustive. All types can be checked with this. However, it doesn't make the fixtures redundant, because those also include the full serde path. I would keep them as there can also be a drift in how shape-identical types de/serialize and having 25% of all types checked with that is already a good chunk. Also more fixtures can be added to increase this percentage over time.

crates/iota-sdk-move-types/src/packages/iota_system.rs
Alex6323: I added the prefix again so it's faithful 1:1 mirror again. Although checked for any other instances, but couldn't find any. So, even if it's slightly redundant, I think it's better to exactly mirror everything.

crates/iota-sdk/examples/decode_staked_iota.rs
Alex6323: Should be possible, but probably better in a follow-up, because we need to discuss how we want to handle failures on individual decode errors: fail the whole call, or return Results for each page item? What about grpc client then? It should have it as well, no? Handling the generic Move types might also be a bit more involved. Not sure what others think @thibault-martinez @Thoralf-M , but IMHO I think this is a separate issue.

Thoralf-M: The move type struct doesn't contain the package and module name? Unless we add that info to the type I don't see how this could work
But yes, either case it shouldn't be done in this PR

bindings/go/examples/decode_staked_iota/main.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+// Decode StakedIota objects into typed Go values via the Rust mirror in
+// iota-sdk-move-types.
+//
+// The GraphQL client returns each object's contents as raw BCS bytes. With the
+// move-types crate exposed through the bindings, a single
+// iota_sdk.StakedIotaTryFromObject(obj) call gives typed, named-field access
+// to Id / PoolId / ActivationEpoch / Principal.
```
Thoralf-M: Please align the examples comment across the languages (can mention the individual language, but overall structure/text should be the same)

crates/iota-sdk-move-types/src/packages/std.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+}
+
+/// Types from `0x1::uq32_32`.
+pub mod uq32_32 {
+    /// Rust version of the Move `std::uq32_32::UQ32_32` type.
+    ///
+    /// An unsigned fixed-point numeric type with 32 integer bits and 32
+    /// fractional bits, represented by an underlying `u64`.
+    #[allow(non_camel_case_types)]
```
Thoralf-M: Switch to `#[expect(`, also in uq64_64

crates/iota-sdk/examples/decode_staked_iota.rs
Alex6323: You would add this information via the `MoveObject` trait.

crates/iota-sdk-move-types/src/packages/mod.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Rust mirrors of the IOTA system packages. Each child module corresponds to
+//! one package; `lib.rs` re-exports them at the crate root so the public API
+//! stays `iota_sdk_move_types::{framework,std,iota_system,stardust}::…`.
```
Thoralf-M: `lib.rs` re-exports them at the crate root so the public API
//! stays `iota_sdk_move_types::{framework,std,iota_system,stardust}::…`.

Would remove this part here, mention it in lib.rs above the re-exports there or not at all

crates/iota-sdk-move-types/bcs-schema.abnf
```diff
@@ -0,0 +1,398 @@
...
+
+nft = bool   ; dummy-field
+
+object-bag = uid   ; id
+             u64   ; size
+
+package-metadata-key = bool   ; dummy-field
+
+per-type-config-created = config-key   ; key
+                          id           ; config-id
```
Thoralf-M, line 168: That's correct with `config-` switched?

crates/iota-sdk/examples/capture_move_type_fixtures.rs
```diff
@@ -0,0 +1,331 @@
...
+                eprintln!(
+                    "  {:32} {:>6} bytes  →  {}",
+                    fixture.file,
+                    bytes.len(),
+                    path.display()
+                );
+            }
+            Err(e) => {
+                failures += 1;
+                eprintln!("  {:32} FAILED: {}", fixture.file, e);
```
Thoralf-M:
```suggestion
                eprintln!("  {:32} FAILED: {e}", fixture.file);
```

crates/iota-sdk-move-types/src/packages_compiled/iota-framework
thibault-martinez, line 1: Do we really need to push them to the repository?

crates/iota-sdk-move-types/bcs-schema.abnf
Alex6323: hmm, I can have a look

Alex6323:
```
pub struct PerTypeConfigCreated {
    pub key: ConfigKey,
    pub config_id: ID,
}
```
so it seems correct to me! (the generated ABNF mirrors the fields and their types which happens to be like this)

crates/iota-sdk-move-types/src/packages_compiled/iota-framework
Alex6323: I think it's worth keeping them. Those blobs are not too big (164KB) to be a concern and if we remove them from the repo the CI needs to pull them in each time to run the tests, which introduces flakiness in case that fails. Also ... if we don't pin them like this, the tests stop being deterministic (2 test runs might test against different package compilations). We'ld have to pin a particular commit to prevent that, but that still doesn't fix potential network errors (Gh having a bad day etc)

thibault-martinez: What if we make this CI check nightly? Pinning them will force us to manually update once in a while and during this time we could break the types without knowing it. As nightly it wouldn't cost CI and we would know daily if we broke them.

Alex6323: That can't really happen, because we have `move_drift_nightly` CI now which makes sure that our compiled packages type surface matches the one in the monorepo with a max delay of 1 day. Let's maybe play this through in our heads:
(1) Monorepo updates their compiled packages (adds a function to some sytem type)
(2) `move_drift_nightly` runs, detects no type actually changed -> passes, we don't need to do anything even though we drifted. 
(3) Monorepo adds a new type to `iota_system` on develop
(4) `move_drift_nightly` runs and warns us on Slack about it
(5) we run the script to pull the new compiled packages and add the new Rust mirror. (The CI messages also states what has do be done exactly.) and open a PR (can for sure be completely be clauded as it can just look how it was done in the past)

Note that because these packages are pretty stable, we don't expect this workflow to happen very often at all. So the effort of manually updating is not a big deal anyway. As I said, this is a task that can be assigned to Claude with "run the download script, create a new mirror type that match the existing ones in design, style, whatever, add it to the testing pipeline, and make a PR". No effort on our side.

At least with the current approach will never merge a change to develop, that doesn't mirror the corresponding Move type at most 1 day before on `develop`. If we just run the tests nightly, then we lose that guarantee and have things on our develop that NEVER were correct mirrors at all at any point in time.

Thoralf-M: 4) what, where is the part reporting to slack? 😂

Alex6323: `failure_slack_report.yml`?

Thoralf-M: Ah ok that makes sense, sounded like `move_drift_nightly` itself does it

Alex6323: yeah, sounded like it, didn't think too long about what I wrote there, haha

bindings/python/examples/decode_staked_iota.py
```diff
@@ -0,0 +1,40 @@
...
+        type_tag="0x3::staking_pool::StakedIota"))
+
+    if len(page.data) == 0:
+        print("No StakedIota objects on testnet right now.")
+        return
+
+    print(f"Decoded {len(page.data)} StakedIota object(s):\n")
+    total_principal = 0
+    for obj in page.data:
+        staked = StakedIota.try_from_object(obj)
```
Thoralf-M, line 28: How would the import look for types from different packages with the same name? For example `ascii::String` and `string::String`

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -0,0 +1,583 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Bindings for selected Rust mirrors of Move types.
```
Thoralf-M, line 4: Please open an issue to add the remaining types

bindings/python/examples/decode_staked_iota.py
Alex6323: we need to disambiguate type conflicts at the FFI layer. We already have a precedence for that with `IotaCoinMetadata`, so in case of a conflict we'ld have manually attach a certain prefix.

You example is actually not a problem because such Strings only appear as fields in wrapping types. Those strings are turned to the uniffi String native type at the serialization layer totally inside of Rust.

crates/iota-sdk-move-types/tests/fixture_roundtrip.rs
```diff
@@ -0,0 +1,124 @@
...
+//! the assertion.
+//!
+//! To refresh the fixtures against current chain state, run:
+//!
+//! ```bash
+//! cargo run -p iota-sdk --example capture_move_type_fixtures
+//! ```
+
+#![cfg(feature = "serde")]
+
```
Thoralf-M:
```suggestion
```

deny.toml
```diff
@@ -45,6 +45,10 @@ ignore = [
   "RUSTSEC-2024-0388",
```
thibault-martinez: While you are at it, this doesn't seem necessary anymore

crates/iota-sdk/Cargo.toml
thibault-martinez: Wondering if this should be default or not

Alex6323: > [!NOTE]
> This analysis is from Claude — I asked it to form an independent opinion on this thread and post it for discussion.

I'd keep it default, based on measuring what it actually costs:

- **Zero new third-party deps.** With `move-types` off, the default graph already contains `bcs` (via graphql-client, tx-builder, types) and `iota-sdk-types` with `serde` enabled (`cargo tree -i bcs` / `-e features -i iota-sdk-types`). Enabling it adds exactly one crate of plain structs, ~1.5s of compile time, wasm-clean.
- **Consistency with the existing default/opt-in line.** The only opt-in feature today is `grpc` — opt-in because tonic/hyper is heavy and doesn't build on wasm. `move-types` is the opposite on both axes.
- **Default-workflow coherence.** `graphql` is default and returns objects whose `contents` are raw BCS — without `move-types`, the default SDK can fetch a `StakedIota` but not read it, and `use iota_sdk::move_types::…` from a docs example fails until the user finds the flag.
- **Reversibility.** The real argument for opt-in is that promoting later is non-breaking while demoting isn't. But that option only pays off if the crate ever gets heavy — and if it does, the gate belongs *inside* `iota-move-types` (a feature there), so the SDK-side default never needs to change.

deny.toml
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/1124/commits/9afbf85c454c7935bb0973966004337ac34bfa96

crates/iota-sdk-move-types/update_compiled_packages.sh
```diff
@@ -1,28 +1,50 @@
...
+# instead of silently fetching the wrong artifacts.
+pinned_rev() {
+    local rev
+    rev="$(sed -n 's/^move-binary-format = .*rev = "\([0-9a-f]\{40\}\)".*/\1/p' "$SCRIPT_DIR/Cargo.toml")"
+    if [[ ! "$rev" =~ ^[0-9a-f]{40}$ ]]; then
+        echo "error: could not extract a 40-hex 'rev' from the move-binary-format line in $SCRIPT_DIR/Cargo.toml" >&2
+        exit 1
+    fi
+    echo "$rev"
+}
```
DaughterOfMars, line 54: eww

Alex6323: Either something like that or dedicated file that holds the SHA pin. Since I wanted a single source of truth, I instructed Claude to extract it from the Cargo.toml.

Alex6323: We could use `cargo` and `jq` instead of `sed` though.

Alex6323:
```
pinned_rev() {
    command -v jq >/dev/null || {
        echo "error: jq is required to resolve the pinned rev from cargo metadata" >&2
        exit 1
    }
    local source rev
    source="$(cargo metadata --no-deps --format-version 1 --manifest-path "$SCRIPT_DIR/Cargo.toml" \
        | jq -r '.packages[]
                 | select(.name == "iota-sdk-move-types")
                 | .dependencies[]
                 | select(.name == "move-binary-format")
                 | .source')"
    rev="${source##*rev=}"
    if [[ ! "$rev" =~ ^[0-9a-f]{40}$ ]]; then
        echo "error: could not resolve a 40-hex 'rev' for the move-binary-format dependency (source: ${source:-<none>})" >&2
        exit 1
    fi
    echo "$rev"
}
```
prefer that?

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/issues/1207

crates/iota-sdk-move-types/src/packages_compiled/iota-framework
Alex6323: decision has been made to remove the compiled packages. The CI runners now fetch those before running the tests: https://github.com/iotaledger/iota-rust-sdk/pull/1124/commits/9afbf85c454c7935bb0973966004337ac34bfa96

crates/iota-sdk-move-types/update_compiled_packages.sh
Alex6323: > [!NOTE]
> From Claude, on Alex's behalf.

Fair. Replaced the sed in 9e6525bdf7ba7abb606dfea9da224dc161efc03b: the rev is now resolved through `cargo metadata --no-deps` + `jq` — cargo's own manifest parser — so manifest reformatting can't break it, and the script can never disagree with the rev the build actually resolves. The only parsing left is taking `rev=` off the machine-generated source URL (`git+https://…?rev=<sha>`), still validated as 40-hex with a loud failure.

Makefile
Alex6323: > [!NOTE]
> From Claude, on Alex's behalf.

Done in 3ccb9db9ba6ed54cee77809ceda1bb913a75d11f — `make bcs-schema` now regenerates the move-types schema too, so the existing `is-dirty` check in CI covers `crates/iota-sdk-move-types/bcs-schema.abnf` from here on.

crates/iota-sdk-move-types/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+/// against `T`. The marker is phantom, so the BCS bytes of e.g. a
+/// `BasicOutput<IOTA>` and a `BasicOutput<OTHER>` are identical — the type
+/// tag is the only place the coin type is recorded, and without this check
+/// one would silently decode as the other.
+///
+/// To decode objects holding your own coin type, define an empty marker
+/// struct and implement this trait for it:
+///
+/// ```
+/// struct FOO;
```
Thoralf-M, line 52: Also needs to implement `T: DeserializeOwned` for `try_from_object()` to work?

crates/iota-sdk-move-types/src/packages/stardust.rs
```diff
@@ -0,0 +1,585 @@
...
+    use iota_types::Address;
+
+    use super::irc27::Irc27Metadata;
+    use crate::framework::object::UID;
+
+    /// Rust version of the Move `stardust::nft::NFT` type.
+    ///
+    /// One-time witness marker. The Move struct is empty; the Rust mirror
+    /// carries a `dummy_field` to preserve the BCS wire format.
+    #[allow(non_camel_case_types)]
```
Thoralf-M:
```suggestion
```
Not needed?

crates/iota-sdk-move-types/src/packages/stardust.rs
```diff
@@ -0,0 +1,585 @@
...
+
+/// Types from `0x107a::stardust_upgrade_label`.
+pub mod stardust_upgrade_label {
+    /// Rust version of the Move
+    /// `stardust::stardust_upgrade_label::STARDUST_UPGRADE_LABEL` type.
+    ///
+    /// Name of the label applied to vested rewards migrated from Stardust.
+    /// The Move struct is empty; the Rust mirror carries a `dummy_field`
+    /// to preserve the BCS wire format.
+    #[allow(non_camel_case_types)]
```
Thoralf-M:
```suggestion
    #[expect(non_camel_case_types)]
```

Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/1124/commits/c0f096a9c48efb089defd0706b0b333702aa0b1a

crates/iota-sdk-move-types/src/packages/stardust.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/1124/commits/c0f096a9c48efb089defd0706b0b333702aa0b1a

crates/iota-sdk-move-types/src/lib.rs
Alex6323: https://github.com/iotaledger/iota-rust-sdk/pull/1124/commits/c0f096a9c48efb089defd0706b0b333702aa0b1a

crates/iota-sdk-move-types/update_compiled_packages.sh
DaughterOfMars: yes, that is better

crates/iota-sdk/Cargo.toml
Thoralf-M: Even if it doesn't add much, I would also be in favour of not having it as default because I think most users won't need it

crates/iota-sdk-move-types/src/move_shape_compare.rs
```diff
@@ -0,0 +1,1654 @@
...
+};
+
+// ---------------------------------------------------------------------------
+// Compiled package blobs (fetched, not committed — see module docs)
+// ---------------------------------------------------------------------------
+
+const IOTA_FRAMEWORK: &[u8] = include_bytes!("packages_compiled/iota-framework");
+const MOVE_STDLIB: &[u8] = include_bytes!("packages_compiled/move-stdlib");
+const IOTA_SYSTEM: &[u8] = include_bytes!("packages_compiled/iota-system");
+const STARDUST: &[u8] = include_bytes!("packages_compiled/stardust");
```
thibault-martinez: I'd prefer reading the files at run time rather than including the bytes at compile time. Preventing compilation if we don't have the files and no network is a big no IMO. Plus at run time we can give better error reporting and exactly what command to run, which is not obvious atm.

Alex6323: yeah, but this is so that the tests can run under WASM

thibault-martinez: What for?

bindings/csharp/examples/DecodeStakedIota/Program.cs
thibault-martinez, line 1: Please add this example to WASM as well

crates/iota-sdk-move-types/src/packages/iota_system.rs
```diff
@@ -0,0 +1,941 @@
...
+        ) -> Result<Self, crate::FromObjectError> {
+            let move_struct = object
+                .as_struct_opt()
+                .ok_or(crate::FromObjectError::NotAMoveStruct)?;
+            if !move_struct.object_type().is_staked_iota() {
+                return Err(crate::FromObjectError::WrongType);
+            }
+            bcs::from_bytes(move_struct.contents()).map_err(crate::FromObjectError::Bcs)
+        }
+    }
```
thibault-martinez, line 120: What's the logic of adding `try_from_bcs`/`try_from_object` or not?

crates/iota-sdk-move-types/src/move_shape_compare.rs
Alex6323: you convinced me, not needed

crates/iota-sdk-move-types/src/packages/iota_system.rs
Alex6323: those make only sense to have for types with the `Key` ability. I only added these for the types that are exposed via FFI. This issue here tracks what types we may want to add in the future: https://github.com/iotaledger/iota-rust-sdk/issues/1207

crates/iota-sdk/Cargo.toml
```diff
@@ -78,3 +77,10 @@ tokio = { workspace = true, features = ["full"] }
 
 [build-dependencies]
 iota-graphql-client-build.workspace = true
+
+# `move-types` is not a default feature, so the example is built only when it
+# is enabled. Without this, default-feature example builds would fail to
+# resolve `iota_sdk::move_types`.
```
Thoralf-M: Would delete that, should be clear enough without
```suggestion
```

crates/iota-sdk/examples/capture_move_type_fixtures.rs
```diff
@@ -0,0 +1,331 @@
...
+                eprintln!("  {:32} FAILED: {e}", fixture.file);
+            }
+        }
+    }
+
+    eprintln!();
+    if failures > 0 {
+        eprintln!(
+            "{} of {} fixtures failed to capture; a pinned object may have been spent/deleted",
+            failures,
```
Thoralf-M:
```suggestion
            "{failures} of {} fixtures failed to capture; a pinned object may have been spent/deleted",
```

Thoralf-M (approved):

crates/iota-sdk-move-types/src/packages/mod.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Rust mirrors of IOTA packages.
+//! Each child module corresponds to one package.
+
+pub mod framework;
+pub mod iota_system;
```
thibault-martinez: Why did you remove `iota_` from framework but not system?

thibault-martinez: I would bring back `iota_framework`

Makefile
```diff
@@ -15,13 +16,17 @@ check-fmt: ## Check code formatting
 fmt: ## Format code
 	cargo +nightly fmt
 
+.PHONY: ensure-compiled-packages
+ensure-compiled-packages: ## Fetch compiled system packages at the pinned rev if missing/stale
```
thibault-martinez: ensure is a bit of a weird name for this, let's just call it update/fetch or whatever

crates/iota-sdk-move-types/Makefile
thibault-martinez, line 1: let's just have rules we actually call from the main Makefile

crates/iota-sdk-move-types/src/packages/mod.rs
Alex6323: I can't remember, but now that I compared again with the monorepo shall I rename it to `iota_framework` and `std` to `move_stdlib`?

crates/iota-sdk-types/src/move_core/struct_tag.rs
thibault-martinez, line 1: Does this need an equivalent update to the FFI?

crates/iota-sdk-move-types/src/packages/mod.rs
thibault-martinez: Yeah, I would have the exact same names as in the monorepo

Alex6323: Should we still re-export `move-stdlib` as `std` and `iota-framwork` as `iota` to match the Move import paths? Also is a bit nicer to read on imports IMO with one consistent ugliness of having `iota::iota::IOTA` instead of `iota_framework::iota::IOTA`.

crates/iota-sdk-move-types/tests/fixture_roundtrip.rs
```diff
@@ -0,0 +1,140 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! BCS roundtrip tests against real on-chain bytes.
+//!
+//! Each test loads a fixture captured from mainnet by the
+//! `capture_move_type_fixtures` example in the `iota-sdk` crate, decodes
+//! it into the corresponding hand-curated Move-mirror type, re-encodes,
+//! and asserts the bytes are bit-identical. A wire-shape mismatch
+//! (renamed field, wrong type, reordered fields, missing variant) fails
+//! the assertion.
```
thibault-martinez: how does a renamed field fails the assertion?

crates/iota-sdk-move-types/src/packages/mod.rs
thibault-martinez: I would personally not re-export

crates/iota-sdk-bcs-schema/src/move_shape.rs
```diff
@@ -0,0 +1,261 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Implementation of `#[derive(MoveShape)]`.
+//!
+//! Emits an `impl crate::move_shape::MoveShape for T` whose `move_shape()`
+//! returns a [`crate::move_shape::Shape`] mirroring the BCS wire layout of `T`
+//! — comparable against a Move struct's `normalized::Type` so a drifted Rust
+//! mirror fails loudly at test time.
```
thibault-martinez, line 9: loudly 🔊

crates/iota-sdk-move-types/src/packages/mod.rs
Alex6323: fine for me

crates/iota-sdk-move-types/tests/fixture_roundtrip.rs
Alex6323: I guess Claude confused it with the shape compare. This is the serde path, so it can't detect a renamed field, which is why we need these 2 separate layers in the first place. They detect different things.

crates/iota-sdk-move-types/src/move_shape_compare.rs
```diff
@@ -0,0 +1,1742 @@
...
+            "transfer_policy",
+            "TransferPolicyDestroyed",
+            iota_framework::transfer_policy::TransferPolicyDestroyed<()>
+        ),
+        // -- 0x1 move-stdlib --------------------------------------------------
+        entry!(
+            MoveStdlib,
+            "fixed_point32",
+            "FixedPoint32",
+            crate::move_stdlib::fixed_point32::FixedPoint32
```
DaughterOfMars: Why the `crate::` prefix for these?

crates/iota-sdk-move-types/tests/fixtures/kiosk_listing.bcs
DaughterOfMars, line 1: Can we not just generate valid fixtures like with the abnf stuff?

crates/iota-sdk-move-types/src/move_shape_compare.rs
Alex6323: to disambiguate from `std`, but we renamed so not necessary anymore. fixing...

crates/iota-sdk-move-types/tests/try_from_object.rs
```diff
@@ -0,0 +1,295 @@
...
+        timelock::TimeLock,
+    },
+    iota_system::iota_system::IotaSystemState,
+    move_stdlib::ascii,
+    stardust::{alias_output::AliasOutput, basic_output::BasicOutput, nft_output::NftOutput},
+};
+use iota_types::{
+    Address, Digest, MoveObjectType, MoveStruct, Object, ObjectData, ObjectId, Owner, StructTag,
+    TypeTag, Version,
+};
```
DaughterOfMars, line 40:
```suggestion
};

```

crates/iota-sdk-move-types/src/packages/move_stdlib.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Rust version of the Move `std::ascii::String` type.
+    ///
+    /// The Move type guarantees that all bytes are valid ASCII. This Rust
+    /// mirror does **not** enforce that invariant — it is up to the caller.
+    /// Wire format: a length-prefixed byte vector.
+    #[derive(Clone, Debug, Default, Eq, PartialEq)]
+    #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+    #[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
+    #[cfg_attr(
+        all(test, not(target_arch = "wasm32")),
```
thibault-martinez, line 41: Just wondering, is there no way to somehow have this condition without the proc-macro? It's pretty verbose

thibault-martinez: I guess not

crates/iota-sdk-move-types/tests/fixtures/kiosk_listing.bcs
Alex6323: I am not sure how that would help, because you then would need to string-compare(!!!) abnf to prevent drift. Also I don't know of any generator from abnf.

crates/iota-sdk-move-types/src/packages/iota_system.rs
```diff
@@ -0,0 +1,1031 @@
...
+        pub fn principal(&self) -> u64 {
+            self.principal.value()
+        }
+    }
+
+    #[cfg(feature = "serde")]
+    impl StakedIota {
+        /// Decode a [`StakedIota`] from BCS bytes (e.g. the `contents` of an
+        /// on-chain Move struct) without verifying the on-chain type tag.
+        pub fn try_from_bcs(bytes: &[u8]) -> Result<Self, bcs::Error> {
```
thibault-martinez: I'd drop the `try_`, we don't usually have it

Alex6323: okay

crates/iota-sdk-move-types/src/packages/iota_system.rs
```diff
@@ -0,0 +1,1031 @@
...
+            bcs::from_bytes(bytes)
+        }
+
+        /// Decode a [`StakedIota`] from an on-chain object, validating that
+        /// the object's type tag matches `0x3::staking_pool::StakedIota`.
+        pub fn try_from_object(
```
thibault-martinez: I'd rather actually impl TryFrom<Object>, more versatile

thibault-martinez: We would need to keep `try_from_object` in FFI ofc

crates/iota-sdk-move-types/src/packages/move_stdlib.rs
Alex6323: I am not sure exactly what you want tbh. By "this condition" you mean all attributes you selected or just the wasm part? I agree though that it's very verbose, but also very explicit. If we'ld somehow merge them or hide them I am not sure that it hits readability/understanding even more :shrug:

crates/iota-sdk-move-types/tests/fixtures/kiosk_listing.bcs
DaughterOfMars: We do it for the other abnf types

crates/iota-sdk-move-types/src/packages/iota_system.rs
Alex6323: I tend to agree. We then lose the naming consistency with `pub fn try_from_object_with_type` though, which is a bit annoying. In the new design it would appear as:
`try_from` and `try_from_object_with_type`? It's a bit less clear that they belong together, but I also don't mind that much. Just making sure you guys want that. @DaughterOfMars  @Thoralf-M

Thoralf-M: I'm fine with either way

crates/iota-sdk-move-types/src/packages/move_stdlib.rs
thibault-martinez: Yeah but it's not like the condition will ever change, it will always be test + not wasm, if there is a way to map it into the proc-macro, it would definitely be more readable

thibault-martinez: I'm actually just more curious about if it's actually doable with rust than asking you to do it tbh

.github/workflows/move_drift_nightly.yml
```diff
@@ -0,0 +1,65 @@
+name: Move Drift Nightly
```
thibault-martinez: Can you rename everything to be Move Types please, not just move, too broad

crates/iota-sdk-move-types/src/packages/iota_system.rs
thibault-martinez: As you want, IMO consistency matters very little when there is a useful trait that can unlock more usecases

crates/iota-sdk-move-types/Makefile
thibault-martinez, line 1: clippy and test unused by the main Makefile

crates/iota-sdk-move-types/tests/fixtures/kiosk_listing.bcs
Alex6323: yeah, I know, but this is not possible here, because how do you know it can decode on-chain instances?

.github/workflows/move_drift_nightly.yml
Alex6323: okay

crates/iota-sdk-move-types/tests/fixtures/kiosk_listing.bcs
Alex6323: so the problem is that for the move types the source of truth is external (monorepo Move system packages) while for `iota-sdk-types` the types themselves are the source of truth.

crates/iota-sdk-move-types/src/packages/iota_system.rs
Alex6323: I am also okay either way

crates/iota-sdk-move-types/tests/fixtures/kiosk_listing.bcs
DaughterOfMars: Well the instances can drift right?

Alex6323: ~~yes, but not without us noticing long before it lands on mainnet through the nightly CI (as soon as a drift appears on `develop` we will get notified).~~

instances can drift (like `Clock`), but not the way how they de/serialize. That's frozen forever. The fixtures only prevent regression, for example, if we accidentally add a #[serde(skip)] on some MIrror type field or something, because that would slip through the Shape comparison.

DaughterOfMars: So then why do these fixtures matter?

Alex6323: (Updated my comment above, but you probably won't get a notification on that)

The fixtures only prevent regression, for example, if we or Claude accidentally add a #[serde(skip)] on some MIrror type field or something, because that would slip through the Shape comparison.

DaughterOfMars: Yeah but we could do that by generating bytes in the right form

Alex6323: From the ABNF you mean?

crates/iota-sdk-bcs-schema/src/lib.rs
```diff
@@ -29,6 +32,16 @@ pub fn derive_bcs_schema(input: TokenStream) -> TokenStream {
     }
 }
 
+#[cfg(feature = "move-shape")]
+#[proc_macro_derive(MoveShape, attributes(bcs_schema))]
```
thibault-martinez: Probably need to remove the attribute

crates/iota-sdk-move-types/tests/try_from_object.rs
thibault-martinez, line 1: Resolve if you are not planning on doing this
---

try_from_object.rs doesn't cover everything:

Missing TryFrom<&Object> tests: Nft, Alias, StakedIota, TimelockedStakedIota.

try_from_object_with_type: tested directly for the 5 coin-marker generics, but only indirectly for TimeLock (its test calls try_from, never try_from_object_with_type).

crates/iota-sdk-move-types/tests/fixtures/kiosk_listing.bcs
DaughterOfMars: yeah the same way we do it in the other crate

crates/iota-sdk-bcs-schema/src/move_shape.rs
```diff
@@ -0,0 +1,261 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+//! Implementation of `#[derive(MoveShape)]`.
+//!
+//! Emits an `impl crate::move_shape::MoveShape for T` whose `move_shape()`
+//! returns a [`crate::move_shape::Shape`] mirroring the BCS wire layout of `T`
+//! — comparable against a Move struct's `normalized::Type` so a drifted Rust
```
thibault-martinez, line 8: Just curious, why not emit the real thing?

crates/iota-sdk-move-types/src/packages/move_stdlib.rs
thibault-martinez: @DaughterOfMars in case you have an input

crates/iota-sdk-move-types/src/lib.rs
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) Mysten Labs, Inc.
```
thibault-martinez: It's not?

crates/iota-sdk-move-types/src/packages/move_stdlib.rs
DaughterOfMars: I'm not entirely sure what you meant either tbh

crates/iota-sdk-move-types/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+
+#[cfg(feature = "serde")]
+impl core::error::Error for FromObjectError {
+    fn source(&self) -> Option<&(dyn core::error::Error + 'static)> {
+        match self {
+            Self::Bcs(e) => Some(e),
+            _ => None,
+        }
+    }
+}
```
thibault-martinez: I'd just use thiserror as usual

crates/iota-sdk-move-types/src/packages/move_stdlib.rs
thibault-martinez: Idk, can we bit emit nothing in the proc macro if the attributes are not there? The same attributes will always be repeated all the tie, can we not somehow factor them in?

crates/iota-sdk-bcs-schema/src/lib.rs
Alex6323: done

crates/iota-sdk-move-types/tests/try_from_object.rs
Alex6323: done

crates/iota-sdk-move-types/src/lib.rs
Alex6323: it's not :trollface: Fixed

crates/iota-sdk-move-types/tests/fixtures/kiosk_listing.bcs
Alex6323: Here's an analysis from Claude with max effort:

```
 The three mirror bug classes

  1. Layout drift — Rust struct's fields/order/types/generics diverge from the on-chain Move type.
  2. Serde-path divergence — the struct shape is right, but the serde behavior emits/accepts different bytes than the chain. Splits into:
    - 2a. asymmetric (e.g. #[serde(skip)] on a real field — the value is lost)
    - 2b. symmetric (e.g. a #[serde(with=…)] / serde_with wrapper / manual impl Deserialize that is its own inverse but encodes differently than chain BCS)
  3. Internal asymmetry — decode(encode(x)) != x for plain derive (rare; basically a serde bug).

  What each layer's oracle can actually see

  ┌────────────────────────────────────────────────┬─────────────────────────────────────────┬────────────────────────────────┬─────┬─────┬─────┬─────┐
  │                     Layer                      │                  Input                  │             Oracle             │  1  │ 2a  │ 2b  │  3  │
  ├────────────────────────────────────────────────┼─────────────────────────────────────────┼────────────────────────────────┼─────┼─────┼─────┼─────┤
  │ Shape comparator                               │ the Rust struct AST                     │ Move bytecode normalized::Type │ ✅  │ ❌  │ ❌  │ ❌  │
  ├────────────────────────────────────────────────┼─────────────────────────────────────────┼────────────────────────────────┼─────┼─────┼─────┼─────┤
  │ proptest round-trip (the iota-sdk-types style) │ Arbitrary instance from the type itself │ itself (decode∘encode = id)    │ ❌  │ ✅  │ ❌  │ ✅  │
  ├────────────────────────────────────────────────┼─────────────────────────────────────────┼────────────────────────────────┼─────┼─────┼─────┼─────┤
  │ real fixtures                                  │ chain bytes                             │ the chain bytes themselves     │ ✅  │ ✅  │ ✅  │ ✅  │
  └────────────────────────────────────────────────┴─────────────────────────────────────────┴────────────────────────────────┴─────┴─────┴─────┴─────┘

  Two cells are the whole story:

  - The comparator never runs serde — it reads the struct definition. A #[serde(skip)] field is still in the AST, so the shape still matches Move; the comparator passes while the wire is wrong. It is blind to all of class 2.
  - proptest's oracle is the type itself. It generates x from the mirror's Arbitrary, encodes with the mirror's serde, decodes with the mirror's serde. It can only prove the mirror is self-consistent. It has no chain byte to compare against, so 2b is invisible to it —
  and 2b is invisible to the comparator too.

  The separating witness (your proof)

  A bug that passes the comparator AND passes proptest but fails a real fixture:

  ▎ Put a self-inverse-but-non-chain serde transform on a field — e.g. wrap a u64 so it BCS-encodes as a length-prefixed string instead of 8 little-endian bytes, via #[serde(with = …)].

  - Comparator: field type is still u64, shape matches Move → PASS
  - proptest: the transform is its own inverse, so decode(encode(x)) == x → PASS
  - real fixture: encode(decode(chain_bytes)) != chain_bytes (chain wrote 8 raw bytes) → FAIL

  That witness is non-empty and realistic, so the three layers are provably not equivalent. The fixtures detect a strictly larger class on the serde path. QED for "not redundant."

  But be honest about how much unique value

  For a mirror that uses only plain #[derive(Serialize, Deserialize)] with no custom serde attributes, the BCS bytes are fully determined by the structural shape — so comparator ⟹ correct wire, and the fixtures' layout guarantee is already covered (the comparator covers
  180 types; fixtures cover ~53). In that regime the fixtures' marginal value is concentrated in exactly three things, none of which any other layer provides:

  1. Regression guard for class 2b — the moment anyone adds a serde_with annotation or a manual impl, only fixtures catch a symmetric divergence.
  2. Independence from the comparator's own fidelity — the comparator trusts that MoveShape's model of BCS (Option-as-0/1, enum index as uleb, the as_type="u64" Balance collapse, etc.) actually matches the chain. Fixtures are the one place that assumption is checked
  against real bytes instead of asserted. They validate the validator.
  3. The serde code path is exercised at all — the comparator runs zero serde; a fixture is the only test that proves Deserialize actually runs on real input.

```

crates/iota-sdk-bcs-schema/src/move_shape.rs
Alex6323: Claude:

Good question — `Shape` is deliberately a *mirror* of `normalized::Type` rather than the real thing, and emitting the real type from the derive isn't free. Three reasons:

1. **The derive only knows the Rust identifier, not the Move origin.** It sees `UID` / `Balance`, not `0x2::object::UID`. A real `normalized::Type::Datatype` references the datatype by module address + name, which the Rust AST doesn't carry — I'd have to hand-annotate every mirror with its Move address/module. `Shape::Datatype { name, args }` intentionally drops the origin and matches by name, which is all the AST affords.

2. **`PhantomData<T>` fields have no representation in `normalized::Type`.** Move models phantom params at the struct level, not as fields, so there's no real-type value meaning "this Rust field is a phantom marker, ignore it." `Shape::Phantom` is an explicit slot the comparator filters out; with the real type I'd have to strip those fields out-of-band before comparing.

3. **`normalized::Type` is heavy/pooled/version-pinned, and the derive emits *tokens*, not a value.** It's generic over `RcIdentifier`/`RcPool` and comes from a pinned `move-binary-format` rev. Emitting tokens that build `Shape::Struct { fields: vec![…] }` is trivial; emitting tokens that construct a valid `normalized::Struct` (pooled identifiers, abilities, etc.) is a lot more machinery and couples the macro output to that crate's exact constructors/version.

So `Shape` is the minimal projection that captures the BCS-relevant structure the comparison needs, constructible from nothing but the Rust AST. (Dependency-wise it'd be fine — `move-binary-format` shares the same `all(test, not(wasm32))` gate as the derive — it's the three points above that make the bespoke type worth it.)

Alex6323: So, it's possible to use `normalized::Type`, but less clean.

crates/iota-sdk-move-types/src/lib.rs
Alex6323: Done

crates/iota-sdk-move-types/tests/try_from_object.rs
thibault-martinez: Don't seem to have done all comment but alright

crates/iota-sdk-bcs-schema/src/move_shape.rs
thibault-martinez: Alright 👍🏻

thibault-martinez (approved):

crates/iota-sdk-move-types/tests/try_from_object.rs
Alex6323: I'll open an issue

Alex6323: https://github.com/iotaledger/iota-rust-sdk/issues/1226

crates/iota-sdk-move-types/tests/fixtures/kiosk_listing.bcs
DaughterOfMars: NGL I don't understand a word of that

Alex6323: okay, lemme try with a human brain then :joy:

The fixtures are necessary because only they are able to ensure that all types can be decoded into our Rust mirrors. However, there is one drift due to BCS being purely positional, that they can't detect apart from that it's impossible to get 100% coverage with them (because not every mirror is an object on-chain), and that is what the shape comparator compensates for. It ensures that field name drift, or accidental re-order of same type fields are detected as well. Proptests like we do in `iota-sdk-types` only ensure that the serialization is inverse, but without real on-chain bytes this doesn't guarantee we can successfully decode from the chain. The fixtures use on-chain data as the source-of-truth (which basically means the Move compiler and the Byte code Verifier itself), `iota-sdk-types` asserts its types are the source-of-truth, so they don't need that.

Alex6323: If the comparator and the prop-tests together could ensure we can always successfully decode, we wouldn't need the fixtures, but they cannot!

DaughterOfMars: alright, well I won't fight it

DaughterOfMars (approved):

thibault-martinez:
don't merge yet

PR #1125, opened 2026-05-07
feat: `ObjectData` improvements

thibault-martinez:
## Summary

Continues the SDK replacement series, stacked on top of `sdk-replacement/23-ProgrammableTransaction`. Rounds out `ObjectData` with the accessors the monorepo expects (object id, type, struct tag) and extends the `def_is_as_into_opt!` macro with mutable-borrow accessors so enums generated via the macro support in-place mutation of their inner values.

### `ObjectData`

New methods on `ObjectData` in [crates/iota-sdk-types/src/object.rs](crates/iota-sdk-types/src/object.rs):

- `id()` — returns the inner `ObjectId` regardless of variant (`Struct` delegates to `MoveStruct::id`, `Package` delegates to `MovePackage::id`).
- `object_type()` — returns `Some(&MoveObjectType)` for `Struct`, `None` for `Package`.
- `struct_tag()` — returns `Some(StructTag)` (cloned) for `Struct`, `None` for `Package`.

### `def_is_as_into_opt!` macro

Adds mutable-borrow accessors to the plain-`$inner` arm of the macro in [crates/iota-sdk-types/src/lib.rs](crates/iota-sdk-types/src/lib.rs):

- `as_<rename>_mut(&mut self) -> &mut $inner` — panicking variant.
- `as_<rename>_mut_opt(&mut self) -> Option<&mut $inner>` — fallible variant.

These mirror the existing `as_<rename>` / `as_<rename>_opt` pair and let callers mutate the inner value of an enum variant without re-matching. The `Box<$inner>` arm is intentionally left unchanged.

### DIFF
--- a/crates/iota-sdk-types/src/object.rs
+++ b/crates/iota-sdk-types/src/object.rs
@@ -172,6 +172,27 @@ pub enum ObjectData {
 
 impl ObjectData {
     crate::def_is_as_into_opt!(Struct(MoveStruct), Package(MovePackage));
+
+    pub fn object_type(&self) -> Option<&MoveObjectType> {
+        match self {
+            Self::Struct(m) => Some(m.object_type()),
+            Self::Package(_) => None,
+        }
+    }
+
+    pub fn struct_tag(&self) -> Option<StructTag> {
+        match self {
+            Self::Struct(m) => Some(m.struct_tag().clone()),
+            Self::Package(_) => None,
+        }
+    }
+
+    pub fn id(&self) -> ObjectId {
+        match self {
+            Self::Struct(v) => v.id(),
+            Self::Package(m) => m.id(),
+        }
+    }
 }
 
 /// A [`StructTag`] with optimized BCS serialization for object types.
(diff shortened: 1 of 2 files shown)


Alex6323 (approved):

crates/iota-sdk-types/src/object.rs
```diff
@@ -172,6 +172,27 @@ pub enum ObjectData {
...
     crate::def_is_as_into_opt!(Struct(MoveStruct), Package(MovePackage));
+
+    pub fn object_type(&self) -> Option<&MoveObjectType> {
+        match self {
+            Self::Struct(m) => Some(m.object_type()),
+            Self::Package(_) => None,
+        }
+    }
+
+    pub fn struct_tag(&self) -> Option<StructTag> {
```
Thoralf-M, line 183: Don't we want `_opt` in the names?

thibault-martinez: If that's fine with you, I've added a note in https://github.com/iotaledger/iota/issues/11459 because maybe these function should be `as_opt_` as well so we'll harmonise it all together

Thoralf-M (approved):

PR #1129, opened 2026-05-07
feat(transaction-builder): impl TryFrom<Transaction> for TransactionBuilder

### LINKED ISSUE #1083: `TransactionBuilder` from `Transaction`
Implement a way to turn a `Transaction` into a `TransactionBuilder` (that returns the same `Transaction` once its `finish` method is called).

Thoralf-M:
Fixes https://github.com/iotaledger/iota-rust-sdk/issues/1083


thibault-martinez (commented):
Probably nice to add to FFI/bindings

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -216,6 +216,83 @@ impl TransactionBuilder {
...
+            unimplemented!("a new Transaction enum variant was added and needs to be handled")
+        };
+        let TransactionKind::ProgrammableTransaction(ProgrammableTransaction {
+            inputs: tx_inputs,
+            commands: tx_commands,
+        }) = kind
+        else {
+            return Err(Error::Input(
+                "only programmable transactions can be converted into a TransactionBuilder".into(),
+            ));
```
DaughterOfMars: Maybe a new variant would be better

DaughterOfMars (approved):
Really nice!

crates/iota-sdk-transaction-builder/src/builder/mod.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1129/commits/6d0d3c032fc362d0b55399de6a91a1774b076abb

crates/iota-sdk-transaction-builder/src/unresolved.rs
```diff
@@ -98,6 +98,47 @@ impl Command {
     }
 }
 
+impl From<iota_types::Command> for Command {
```
Alex6323, line 101: How is this related?

Thoralf-M: In the TransactionBuilder we have the unresolved `Command` here, in the actual tx it's `iota_types::Command` so we need a way to go back

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -216,6 +216,81 @@ impl TransactionBuilder {
...
+        let Transaction::V1(TransactionV1 {
+            kind,
+            sender,
+            gas_payment,
+            expiration,
+        }) = tx
+        else {
+            unimplemented!("a new Transaction enum variant was added and needs to be handled")
+        };
+        let TransactionKind::Programmable(ProgrammableTransaction {
```
thibault-martinez: Maybe we can impl `From<ProgrammableTransaction>` and call it in this `TryFrom<Transaction>`

thibault-martinez: Handier if people already have a `ProgrammableTransaction`

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -216,6 +216,81 @@ impl TransactionBuilder {
...
+        let mut inputs = BTreeMap::new();
+        for (i, inp) in tx_inputs.into_iter().enumerate() {
+            inputs.insert(
+                i,
+                Input {
+                    kind: InputKind::Input(inp),
+                    is_gas: false,
+                },
+            );
+        }
```
Alex6323: not sure I get this part: just pushing them to a `Vec` would also preserve their index?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -216,6 +216,81 @@ impl TransactionBuilder {
...
+        else {
+            return Err(Error::UnsupportedTransactionKind);
+        };
+
+        // Non-gas inputs are inserted with keys 0..n preserving their original
+        // index, so that `Argument::Input(i)` remains valid without remapping.
+        // Gas inputs follow with subsequent keys; BTreeMap ordering ensures
+        // `finish` re-emits the inputs and gas vectors in the original order.
+        let mut inputs = BTreeMap::new();
+        for (i, inp) in tx_inputs.into_iter().enumerate() {
```
thibault-martinez: call it input please

crates/iota-sdk-transaction-builder/src/builder/mod.rs
Alex6323: I mean ... in the end we need a BTreeMap, I see, but maybe we should then just collect into a BTreeMap?

Alex6323: So can we not just create a new vec with map to create the Input, then down below we just extend that vec, and in the end we iter and collect it into a `BTreeMap`?

thibault-martinez: We could, but why? Sounds like more allocations than needed?

Alex6323: less LOC and I would argue still a bit more efficient, but you're right in that we don't need to do it.

crates/iota-sdk-transaction-builder/src/builder/mod.rs
Thoralf-M: Sure, https://github.com/iotaledger/iota-rust-sdk/pull/1129/commits/9f892d2fabf9682f12a994bb5ad134639cf6a251

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: Not sure why allocating an extra temporary vec would be more efficient

Thoralf-M: No extra BTreeMap or Vec created anymore

Alex6323: append to a vec is O(1) while for BTreeMap is O(log n), and the compiler can better optimize for combinators (`map`, `collect`). So I think for many inputs it could be worth it?

Alex6323: however, I am fine with both solutions, just pointing out an alternative.

Alex6323 (approved):

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: The average transaction is too small for that to actually make a positive difference, plus the transaction size limit is probably lower for it to ever be able to make a positive difference

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -216,6 +216,107 @@ impl TransactionBuilder {
...
+            commands: tx_commands,
+        } = ptb;
+
+        // Inputs are inserted with keys 0..n preserving their original index,
+        // so that `Argument::Input(i)` referenced from the commands remains
+        // valid without remapping.
+        let inputs = tx_inputs
+            .into_iter()
+            .enumerate()
+            .map(|(i, inp)| {
```
thibault-martinez: call it input please

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1129/commits/67b0ba8702a55ec524ae1d005920367724fffa34

thibault-martinez (approved):

PR #1133, opened 2026-05-08
feat(iota-sdk-types)!: make MoveStruct contents checked

Thoralf-M:
## Summary

Continues the SDK replacement series, stacked on top of `sdk-replacement/28-TransactionData`. Closes the long-standing TODO on `MoveStruct` to enforce that the BCS contents always carry a leading `ObjectId`, by making the fields private and gating construction through a fallible constructor.

- **`MoveStruct` fields are now private** in [crates/iota-sdk-types/src/object.rs](crates/iota-sdk-types/src/object.rs): `object_type`, `version`, `contents`. Read access goes through the existing accessors (`object_type()`, `version()`, `contents()`); write access goes through new setters (`set_object_type`, `set_version`, `set_contents`). Direct field access and `MoveStruct { .. }` literal construction no longer compile.
- **New fallible constructor** `MoveStruct::new(object_type, version, contents) -> Result<Self, MoveStructContentsError>`: rejects contents shorter than `ObjectId::LENGTH` (32 bytes), since the leading bytes of every Move object's BCS contents are its `ObjectId`. `set_contents` enforces the same invariant; `set_version` and `set_object_type` are infallible.
- **`MoveStruct::id()` is now infallible** — the constructor's invariant guarantees the leading 32 bytes parse as an `ObjectId`, so the `unwrap()` is always sound. The previous `MoveStruct::id_opt(&[u8])` associated function and the equivalent free `id_opt` helper in `object.rs` are removed; callers that used to do `id_opt(&struct_.contents)` now call `struct_.id()` directly.
- **`MoveStruct` gets a hand-rolled `Serialize` / `Deserialize`** in the `serialization` module of [crates/iota-sdk-types/src/object.rs](crates/iota-sdk-types/src/object.rs), routed through new private `ReadableMoveStructRef` / `ReadableMoveStructOwned` proxies. Deserialization goes through `MoveStruct::new`, so the contents-length invariant is also enforced on the wire. BCS form is unchanged. JSON form changes only in field naming: the proxy uses `#[serde(rename_all = "camelCase")]`, so `object_type` becomes `objectType` (the other two fields are already camelCase / single-word). The `Object` / `GenesisObject` deserialization paths now construct via `MoveStruct::new` and then verify the contents-encoded id matches the outer `object_id`.
- **`MoveStructContentsError`** is added (using `thiserror`) and re-exported from the crate root in [crates/iota-sdk-types/src/lib.rs](crates/iota-sdk-types/src/lib.rs).
- **Internal call sites and examples updated** to use the accessors / constructor: `Coin::try_from(&Object)` in [crates/iota-sdk-types/src/framework.rs](crates/iota-sdk-types/src/framework.rs), the FFI bindings in [crates/iota-sdk-ffi/src/types/object.rs](crates/iota-sdk-ffi/src/types/object.rs), and the `iota-sdk` examples (`abstract_account`, `get_object`, `package_inspect`, `publish_upgrade`).

BCS round-trips identically. Breaking at the Rust API level for any caller that constructs `MoveStruct { .. }` directly, reads/writes the fields by name, or uses `MoveStruct::id_opt` / the free `id_opt` — replace with `MoveStruct::new(..)?`, the accessor/setter methods, and `.id()` respectively.

### DIFF
--- a/crates/iota-sdk-types/tests/bcs_schema_fuzzing.rs
+++ b/crates/iota-sdk-types/tests/bcs_schema_fuzzing.rs
@@ -410,6 +410,7 @@ impl TestHarness {
             "checkpoint-transaction",
             Self::generate_checkpoint_transaction,
         );
+        overrides.insert("move-struct", Self::gen_move_struct);
 
         Self {
             grammar,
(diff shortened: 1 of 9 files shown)


crates/iota-sdk-types/tests/bcs_schema_fuzzing.rs
thibault-martinez, line 1: I like this file, very well made

thibault-martinez (approved):

Alex6323 (approved):

PR #1136, opened 2026-05-08
fix(bindings,ci): Remove unsafeFlags from swift publish

DaughterOfMars:
## Summary

Fixes `make swift-release-example`, which currently errors with:

```
error: 'release': the target 'IotaSDK' in product 'IotaSDK' contains unsafe build flags
```

The published `Package.swift` on `iotaledger/iota-sdk-swift` carries a `linkerSettings: [.unsafeFlags(["-L", "Sources/CIotaSDK", "-liota_sdk_ffi"])]` clause on the `IotaSDK` target. SwiftPM forbids consuming a package whose targets declare `.unsafeFlags(...)`, so any downstream consumer (including the release example) is broken. The flags were a side-effect of the publish workflow archiving the local-dev `bindings/swift/Package.swift` verbatim and overwriting the consumer repo's `Package.swift` with it on every release — the comment in that file claiming the published copy uses an XCFramework was aspirational.

This PR makes that comment true: the publish flow now builds an XCFramework from the per-target dylibs and ships a clean `Package.swift` that references it via `.binaryTarget`. The local-dev `Package.swift` (still using `.unsafeFlags` for fast in-repo iteration) is unchanged.

## Changes

**`bindings/swift/dist/`** (new)
- `Package.swift` — the manifest that gets written into `iota-sdk-swift` on publish. Uses `.binaryTarget(name: "CIotaSDK", path: "IotaSDK.xcframework")`, drops the example targets and `.unsafeFlags`.
- `module.modulemap` — bundled into each XCFramework slice's `Headers/`.

**`.github/workflows/swift_publish.yml`**
- New `xcframework` job: lipos the macOS x86_64 + arm64 dylibs into a universal slice, then `xcodebuild -create-xcframework` over the macOS fat lib + iOS device dylib + iOS-sim dylib. Produces `IotaSDK.xcframework` as an artifact.
- `release-archives` now consumes `xcframework` (instead of `prepare-libraries-archives` directly), copies `dist/Package.swift` over the local-dev one, removes `Sources/CIotaSDK`, `lib`, and `dist` before commit.
- `prepare-source-archive` now ships `bindings/swift/dist` instead of `bindings/swift/Package.swift`.
- `prepare-libraries-archives` matrix simplified: `runs-on: macos-latest` hoisted out of every entry, and the constant `lib: libiota_sdk_ffi.dylib` field is gone (inlined where used).

**`.github/workflows/{swift,csharp,go}_publish.yml`**
- `if:` guards on `prepare-libraries-archives` standardized from `contains(tag_name, ...)` to `startsWith(tag_name, ...)`, matching the source-archive jobs and being precise about the tag prefix format.


bindings/swift/dist/Package.swift
Thoralf-M, line 1: Should dist/ be committed?

DaughterOfMars: Yes, it is not generated

Thoralf-M (approved):

.github/workflows/swift_publish.yml
```diff
@@ -73,26 +65,66 @@ jobs:
...
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        with:
+          name: xcframework
+          path: IotaSDK.xcframework
+
   release-archives:
     name: Release archives
     runs-on: ubuntu-latest
     needs:
       - prepare-source-archive
```
thibault-martinez: This could be removed as well

.github/workflows/swift_publish.yml
```diff
@@ -101,10 +133,21 @@ jobs:
           ssh-key: ${{ secrets.SSH_PRIVATE_KEY_FOR_SWIFT_SDK }}
           fetch-depth: 0
 
-      - name: Download source and libraries archives
+      - name: Download source archive
         uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
         with:
-          merge-multiple: true
```
thibault-martinez, line 107: Sure about removing this?

thibault-martinez: Ah, I got tricked by the git diff, all good

.github/workflows/swift_publish.yml
DaughterOfMars: Explain?

thibault-martinez: `xcframework` already needs it

DaughterOfMars: ah, I see

bindings/swift/dist/Package.swift
```diff
@@ -0,0 +1,28 @@
+// swift-tools-version: 5.9
+
+import PackageDescription
+
+let package = Package(
+    name: "IotaSDK",
+    platforms: [
+        .macOS(.v13),
+        .iOS(.v16),
```
thibault-martinez, line 9: Not sure if that's a problem but it contradicts the `IPHONEOS_DEPLOYMENT_TARGET=18.5` of the publish workflow

thibault-martinez (approved):

bindings/swift/dist/Package.swift
DaughterOfMars: I reduced the one in the publish workflow to match

thibault-martinez: Pretty sure it was fixed to 18.5 for a reason

thibault-martinez: Why couldn't we bump this one instead?

DaughterOfMars: I think it's a minimum so it's better to be lower

thibault-martinez: Alright

PR #1137, opened 2026-05-10
feat!: `MultisigAggregatedSignature` changes

thibault-martinez:
## Summary

Replaces `MultisigMemberPublicKey` with a unified `PublicKey` enum reused across the SDK, and tightens `MultisigCommittee` construction with input validation.

- New `PublicKey` enum in `iota-sdk-types` (Ed25519 / Secp256k1 / Secp256r1 / ZkLoginDeprecated / Passkey) with `to_bytes`/`from_bytes`/`to_base64`/`from_base64`/`derive_address`. `MultisigMember` now stores a `PublicKey`.
- `MultisigCommittee::new` now validates the committee (member count 1..=10, non-zero weights, non-zero threshold ≤ sum of weights, no duplicate public keys). The previous non-validating constructor is renamed to `insecure_new` for deserialization/test paths. Errors surface via a new `MultisigError`.
- `MultisigAggregatedSignature` validates the bitmap and signature count on construction.
- `UserSignature` gains `to_public_key()`; `from_serialized_bytes` is renamed to `from_bytes` for consistency with the other byte helpers.
- FFI: `PublicKey` is split out as its own type and reused by multisig; added `MultisigAggregatedSignature::new`. Bindings examples updated accordingly.
- Custom proptest strategy for multisigs so generated committees satisfy the new invariants.

[run-ci]

## Test plan

- [x] `make test`
- [x] `make clippy`
- [x] `make wasm`
- [x] `make bindings` + binding examples


crates/iota-sdk-types/src/hash.rs
```diff
@@ -394,6 +404,16 @@ mod signing_message {
             hex::encode(self.signing_message())
         }
     }
+
+    impl<T> IntentMessage<T>
+    where
+        T: serde::Serialize,
+    {
+        // TODO right name?
+        pub fn signing_message(&self) -> Digest {
```
Thoralf-M: Is this even needed anywhere?

crates/iota-sdk-types/src/crypto/signature.rs
```diff
@@ -329,11 +329,41 @@ impl UserSignature {
...
+            UserSignature::Simple(simple) => match simple {
+                SimpleSignature::Ed25519 { public_key, .. } => Ok(PublicKey::Ed25519(*public_key)),
+                SimpleSignature::Secp256k1 { public_key, .. } => {
+                    Ok(PublicKey::Secp256k1(*public_key))
+                }
+                SimpleSignature::Secp256r1 { public_key, .. } => {
+                    Ok(PublicKey::Secp256r1(*public_key))
+                }
+            },
+            // TODO this should maybe work
```
Thoralf-M: How if there are multiple keys?

crates/iota-sdk-types/src/hash.rs
thibault-martinez: In the monorepo, yes

crates/iota-sdk-types/src/crypto/signature.rs
thibault-martinez: Not sure what you mean

Thoralf-M: Why should `to_public_key()` work for `MultiSig` if there are usually multiple public keys in a multisig? Which one would it return?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -265,18 +262,69 @@ impl MultisigAggregatedSignature {
     /// the caller to ensure that the provided signature list is in the same
     /// order as it's corresponding member in the provided committee
     /// and that it's position in the provided bitmap is set.
-    pub fn new(
-        committee: MultisigCommittee,
+    pub fn insecure_new(
```
Thoralf-M: Also update the comment to mention that this doesn't validate the signatures against the committee?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -265,18 +262,69 @@ impl MultisigAggregatedSignature {
...
+    /// pk4, pk5], [sig1, sig2, sig5] is valid, but [sig2, sig1, sig5] is
+    /// invalid.
+    pub fn new(
+        signatures: Vec<UserSignature>,
+        committee: MultisigCommittee,
+    ) -> Result<Self, MultisigError> {
+        committee.is_valid()?;
+
+        if signatures.len() > committee.members.len() || signatures.is_empty() {
+            return Err(MultisigError::InvalidSignatureNumber);
```
Thoralf-M, line 351: No need to check that there signatures.len() >= threshold ?

crates/iota-sdk-types/src/crypto/signature.rs
thibault-martinez: Yeah you're right, I thought there was a intermediate step when deriving the address but there's not

crates/iota-sdk-types/src/crypto/multisig.rs
Thoralf-M: Also that's the same as `is_valid()`? There is one more check, but can we just call `is_valid()` at the end?

crates/iota-sdk-types/src/crypto/signature.rs
thibault-martinez: Removed the TODO

crates/iota-sdk-types/src/crypto/public_key.rs
```diff
@@ -0,0 +1,157 @@
...
+///                                            ; and the decoded bytes are defined
+///                                            ; by legacy-public-key
+/// legacy-public-key = (ed25519-flag ed25519-public-key) /
+///                     (secp256k1-flag secp256k1-public-key) /
+///                     (secp256r1-flag secp256r1-public-key)
+/// ```
+#[derive(Clone, Debug, PartialEq, Eq, derive_more::From)]
+#[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[non_exhaustive]
+// TODO feature gate?
```
Thoralf-M: Feature gate what? In iota-sdk-crypto it makes sense to have feature gates because you might only need a subset to sign/verify your own things, but I don't see why we would feature gate anything here

crates/iota-sdk-types/src/hash.rs
```diff
@@ -176,6 +176,18 @@ impl crate::PasskeyPublicKey {
     }
 }
 
+impl crate::PublicKey {
+    pub fn derive_address(&self) -> Address {
+        match self {
+            Self::Ed25519(pk) => pk.derive_address(),
+            Self::Secp256k1(pk) => pk.derive_address(),
+            Self::Secp256r1(pk) => pk.derive_address(),
+            Self::ZkLoginDeprecated => panic!(),
```
Thoralf-M: At least have a panic message

crates/iota-sdk-types/src/crypto/public_key.rs
thibault-martinez: Yeah I guess I agree, I just thought it would be a bit asymetrical if the enums in crypto are gated and not the ones in types

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: Improved the docs with Claude https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/bf466474787a54380e20224025b223aeb4e1d194

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: > No need to check that there signatures.len() >= threshold ?

No, number of signatures can be lower than threshold (it's the sum of weight that matters)

thibault-martinez: > Also that's the same as is_valid()? There is one more check, but can we just call is_valid() at the end?

I guess we can yeah

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/9ab1f184174f0444f0ff806b512451d0ad5f46ba

crates/iota-sdk-types/src/hash.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/changes/49466a8c330b67dd7825d0b963289cf3ec7d967b

crates/iota-sdk-types/src/crypto/multisig.rs
Thoralf-M: `committee.is_valid()?;` should not stay?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -288,16 +373,39 @@ impl MultisigAggregatedSignature {
...
+}
+
+/// Interpret a bitmap of 01s as a list of indices that is set to 1s.
+/// e.g. 22 = 0b10110, then the result is [1, 2, 4].
+fn as_indices(bitmap: u16) -> Result<Vec<u8>, MultisigError> {
+    if bitmap > MAX_BITMAP_VALUE {
+        return Err(MultisigError::InvalidBitmap(bitmap));
+    }
+    let mut res = Vec::new();
+    for i in 0..10 {
```
Thoralf-M:
```suggestion
    for i in 0..MAX_COMMITTEE_SIZE {
```

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -241,40 +253,113 @@ impl MultisigCommittee {
...
+    /// `[pk1, pk2, pk3, pk4, pk5]`, `[sig1, sig2, sig5]` is valid but
+    /// `[sig2, sig1, sig5]` is not.
+    pub fn new(
+        signatures: Vec<UserSignature>,
+        committee: MultisigCommittee,
+    ) -> Result<Self, MultisigError> {
+        let mut bitmap = 0;
+        let mut member_signatures = Vec::with_capacity(signatures.len());
+        for signature in signatures {
+            let pk = signature.to_public_key().unwrap();
```
Thoralf-M: Handle errors and don't panic?

Thoralf-M:
Please add a PR description and it would be also good to have some more test coverage, for example things like https://github.com/iotaledger/iota/blob/9cda1731c2325a3e54e3045319e9817c14bf021c/crates/iota/src/unit_tests/keytool_tests.rs#L687 or a test that makes sure that a new Multisig fails if an inner variant is also Multisig/not enough or too many members added

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/changes/ec3cab256bfa1ff2162a4dc05fb95ebef790a3b0

crates/iota-sdk-types/src/crypto/multisig.rs
Thoralf-M: Ah it's checked at the end of signature.is_valid()

thibault-martinez: (reopening because I think I saw an issue)

crates/iota-sdk-crypto/src/multisig.rs
```diff
@@ -110,10 +109,12 @@ impl Verifier<MultisigAggregatedSignature> for MultisigVerifier {
...
     ) -> Result<(), SignatureError> {
-        if !signature.committee().is_valid() {
-            return Err(SignatureError::from_source("invalid MultisigCommittee"));
-        }
+        signature
+            .committee()
+            .is_valid()
+            .map_err(|e| SignatureError::from_source(format!("invalid MultisigCommittee: {e}")))?;
 
+        // TODO should we add this to is_valid?
```
Thoralf-M: Makes sense to me, don't see why not

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -149,26 +118,57 @@ impl MultisigMember {
 /// ```
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
```
Thoralf-M, line 152: Why remove proptest?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -185,18 +185,25 @@ impl MultisigCommittee {
...
     ///  - Has at least one member
     ///  - Has at most ten members
     ///  - No member has weight 0
-    ///  - the sum of the weights of all members must be larger than the
-    ///    threshold
+    ///  - the sum of the weights of all members must be at least the threshold
     ///  - contains no duplicate members
-    pub fn is_valid(&self) -> bool {
-        self.threshold != 0
+    pub fn is_valid(&self) -> Result<(), MultisigError> {
```
Thoralf-M: Why change this to a Result?

crates/iota-sdk-crypto/src/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/d56bce22c894c2cbb2ac372b422d3b3938a6f0e9

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: It has to be manually implemented because of the validation rules

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: I wanted the opportunity to have dedicated error messages instead of wrapping everything to a false. I'm not saying I will do it but at least it's future proof. Think I should revert?

Thoralf-M: But did you want to add dedicated error messages only in the future? I was wondering because now it's just one variant
Generally I also prefer the Result, but I think it should then also be called `verify()` instead, for `is_valid()` I rather expect a bool result

Thoralf-M: Or maybe `validate()` would fit better here

crates/iota-sdk-types/src/crypto/multisig.rs
Thoralf-M: Care to elaborate?

thibault-martinez: After cinema 👍

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: I agree, https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/1f96ee26e1ea5a8935d1452f7191b278b9e90b9d

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -2,85 +2,50 @@
...
-                secp256r1_public_key.scheme()
-            }
-            MultisigMemberPublicKey::ZkLoginDeprecated => {
-                SignatureScheme::ZkLoginAuthenticatorDeprecated
-            }
-            MultisigMemberPublicKey::Passkey(passkey_public_key) => passkey_public_key.scheme(),
-        }
-    }
+#[derive(Debug, thiserror::Error)]
+pub enum MultisigError {
```
Thoralf-M, line 25: #[non_exhaustive] ?

crates/iota-sdk-types/src/crypto/public_key.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use base64ct::{Base64, Encoding};
+
+use super::{
+    Ed25519PublicKey, PublicKeyExt, Secp256k1PublicKey, Secp256r1PublicKey, SignatureScheme,
+    passkey::{PasskeyAuthenticator, PasskeyPublicKey},
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum PublicKeyError {
```
Thoralf-M, line 14: #[non_exhaustive] ?

crates/iota-sdk-ffi/src/types/crypto/public_key.rs
```diff
@@ -0,0 +1,117 @@
...
+///                      (secp256r1-flag secp256r1-public-key) /
+///                      (zklogin-flag-deprecated) /
+///                      (passkey-flag passkey-public-key)
+/// ```
+#[derive(Debug, derive_more::From, Eq, PartialEq, uniffi::Object)]
+#[uniffi::export(Debug, Eq)]
+pub struct PublicKey(pub iota_sdk::types::PublicKey);
+
+#[uniffi::export]
+impl PublicKey {
```
Thoralf-M, line 47: Also expose to_bytes/from_bytes/to_base64/from_base64/derive_address ?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -241,42 +252,127 @@ impl MultisigCommittee {
...
+                .get_public_key_index(&pk)
+                .ok_or(MultisigError::NoPublicKeyForSignature(sig_index))?;
+            if let Some(prev) = prev_index {
+                match index.cmp(&prev) {
+                    std::cmp::Ordering::Less => return Err(MultisigError::SignaturesOutOfOrder),
+                    std::cmp::Ordering::Equal => return Err(MultisigError::DuplicatePublicKey),
+                    std::cmp::Ordering::Greater => {}
+                }
+            }
+            bitmap |= 1 << index;
```
Thoralf-M, line 374: crates/iota-sdk-types/src/crypto/multisig.rs:322-359 + :646-664 vs :680-706 — MultisigAggregatedSignature::new runs the bitmap-shift loop before committee.validate() (panic vector when index ≥ 16 with insecure_new-built committees), and the BCS deserialization path validates while the JSON path does not (round-trip divergence)

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/3f9398ca35aac4bc0d7a0564d0dcbcebb2c3481c

crates/iota-sdk-types/src/crypto/public_key.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/3f9398ca35aac4bc0d7a0564d0dcbcebb2c3481c

crates/iota-sdk-ffi/src/types/crypto/public_key.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/3f9398ca35aac4bc0d7a0564d0dcbcebb2c3481c

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: Yes this is why I reopened your comment about reusing validate, I noticed maybe we couldn't

crates/iota-sdk-crypto/src/multisig.rs
```diff
@@ -3,8 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use iota_types::{
-    MultisigAggregatedSignature, MultisigCommittee, MultisigMemberPublicKey,
-    MultisigMemberSignature, UserSignature,
+    MultisigAggregatedSignature, MultisigCommittee, MultisigMemberSignature, PublicKey,
```
semenov-vladyslav: These types are defined in iota_sdk_types, no need to import them from iota_types.

Here and in other places.

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -340,20 +460,141 @@ impl MultisigMemberSignature {
...
+                    Just(committee),
+                    Just(bitmap),
+                    vec(any::<MultisigMemberSignature>(), count..=count),
+                )
+            })
+            .prop_map(|(committee, bitmap, signatures)| Self {
+                signatures,
+                bitmap,
+                committee,
+                bytes: OnceCell::new(),
```
semenov-vladyslav: `bytes` field must be guarded with `#[cfg(feature = "serde")]`

crates/iota-sdk-types/Cargo.toml
```diff
@@ -51,6 +51,7 @@ bs58 = "0.5.1"
 bytestring = "1.5.0"
 derive_more = { workspace = true, features = ["full"] }
 hex.workspace = true
+once_cell = "1.21"
```
semenov-vladyslav: Why not use `std::cell::OnceCell`? Otherwise `once_cell` dependency should be optional as it is only used with `#[cfg(feature = "serde")]`.

crates/iota-sdk-ffi/src/types/crypto/public_key.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub fn is_ed25519(&self) -> bool {
+        self.0.is_ed25519()
+    }
+
+    pub fn as_ed25519_opt(&self) -> Option<Arc<Ed25519PublicKey>> {
```
Alex6323, line 52: should this be `as_opt_ed25519` now? (also the other instances)

thibault-martinez: I'll deal with this in the dedicated PR

crates/iota-sdk-types/src/hash.rs
```diff
@@ -394,6 +406,15 @@ mod signing_message {
             hex::encode(self.signing_message())
         }
     }
+
+    impl<T> IntentMessage<T>
+    where
+        T: serde::Serialize,
+    {
+        pub fn signing_digest(&self) -> Digest {
```
semenov-vladyslav, line 408: Why is it needed? Is it equivalent to `signing_digest(self.intent, self.value)`?

crates/iota-sdk-types/src/crypto/public_key.rs
```diff
@@ -0,0 +1,174 @@
...
+///              secp256k1-public-key-variant /
+///              secp256r1-public-key-variant /
+///              zklogin-public-key-variant-deprecated /
+///              passkey-public-key-variant
+///
+/// ed25519-public-key-variant              = %d00 ed25519-public-key
+/// secp256k1-public-key-variant            = %d01 secp256k1-public-key
+/// secp256r1-public-key-variant            = %d02 secp256r1-public-key
+/// zklogin-public-key-variant-deprecated   = %d03
+/// passkey-public-key-variant              = %d04 passkey-public-key
```
semenov-vladyslav: `from_bytes` uses `SignatureScheme` values as the discriminant byte. But `SignatureScheme` has more variants that will mess up with the encoding: `Multisig = 0x03` and `Bls12381 = 0x04`.

I would consider adding `Multisig(!)` and `Bls12381(!)` variants to `PublicKey` enum; the `!` will indicate that these variants are forbidden and impossible to construct.

crates/iota-sdk-types/src/crypto/public_key.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// Decode a public key from its scheme-flagged byte representation
+    pub fn from_bytes(bytes: impl AsRef<[u8]>) -> Result<Self, PublicKeyError> {
+        let bytes = bytes.as_ref();
+
+        match bytes.first() {
+            Some(x) => {
+                if x == &(SignatureScheme::Ed25519 as u8) {
+                    let pk = Ed25519PublicKey::from_bytes(&bytes[1..])?;
+                    Ok(Self::Ed25519(pk))
+                } else if x == &(SignatureScheme::Secp256k1 as u8) {
```
semenov-vladyslav:
```suggestion
        match bytes.split_first() {
            Some(flag, tail) => {
                match SignatureScheme::from_byte(*flag) {
                    SignatureScheme::Ed25519 => {
                        let pk = Ed25519PublicKey::from_bytes(tail)?;
                        Ok(Self::Ed25519(pk))
                    },
                }
```
There is `split_first()` allowing to avoid taking slice `bytes[1..]` to get the tail.
There is `SignatureScheme::from_byte` that does proper conversion from a byte, no need for manual matching.

crates/iota-sdk-crypto/src/multisig.rs
semenov-vladyslav:
```suggestion
                    .map(|()| member.weight() as ThresholdUnit)
```

semenov-vladyslav:
```suggestion
        self.signed_weight += self.committee.members()[member_idx].weight() as ThresholdUnit;
```
Or even use `checked_add()` just in case.

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -212,6 +219,11 @@ impl MultisigCommittee {
                     .skip(i + 1)
                     .any(|m| member.public_key == m.public_key)
```
semenov-vladyslav: Should it be `member.public_key >= m.public_key` to ensure that the same committee (set of signers) has a unique binary representation? Otherwise a multisig will be malleable -- reordering public keys and corresponding signatures in the `MultisigAggregatedSignature` will result in a different but equally valid signature.

crates/iota-sdk-types/src/crypto/multisig.rs
semenov-vladyslav: What is the reason to add a public "insecure" method?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -241,42 +253,127 @@ impl MultisigCommittee {
...
+            bytes: OnceCell::new(),
+        };
+
+        signature.validate()?;
+
+        Ok(signature)
+    }
+
+    /// Validates the structural integrity of this aggregated signature.
+    pub fn validate(&self) -> Result<(), MultisigError> {
```
semenov-vladyslav, line 391: The `Self::new()` method already does some validation in addition to the check in here. But a user can skip those with seemingly safe code:
```
let s = MultisigAggregatedSignature::insecure_new(signatures, bitmap, committee);
s.validate()?;
```
The validation checks are not that expensive to justify giving user possibility to mess up.

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -288,16 +385,39 @@ impl MultisigAggregatedSignature {
         self.bitmap
     }
 
+    pub fn get_indices(&self) -> Result<Vec<u8>, MultisigError> {
```
semenov-vladyslav: `self.bitmap` should already contain a valid value, why the return type should be `Result`? The API should not allow possibility to construct invalid values of `Self`.

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -340,20 +460,141 @@ impl MultisigMemberSignature {
...
+        match signature {
+            UserSignature::Simple(SimpleSignature::Ed25519 { signature, .. }) => {
+                Ok(Self::Ed25519(signature))
+            }
+            UserSignature::Simple(SimpleSignature::Secp256k1 { signature, .. }) => {
+                Ok(Self::Secp256k1(signature))
+            }
+            UserSignature::Simple(SimpleSignature::Secp256r1 { signature, .. }) => {
+                Ok(Self::Secp256r1(signature))
+            }
```
semenov-vladyslav:
```suggestion
            UserSignature::Simple(simple) => Ok(simple.into()),
```
There is `impl From<SimpleSignature> for MultisigMemberSignature`

crates/iota-sdk-types/src/crypto/multisig.rs
semenov-vladyslav, line 683: There is a safe `MultisigAggregatedSignature::new()` constructor that does all the verification. The input is provided by user and untrusted, we can't just initialize struct fields without validation.

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -438,26 +680,29 @@ mod serialization {
             buf
         }
 
-        pub fn from_serialized_bytes(
-            bytes: impl AsRef<[u8]>,
-        ) -> Result<Self, SignatureFromBytesError> {
+        pub fn from_bytes(bytes: impl AsRef<[u8]>) -> Result<Self, SignatureFromBytesError> {
```
semenov-vladyslav, line 710:
```suggestion
        pub(crate) fn from_bytes(bytes: impl AsRef<[u8]>) -> Result<Self, SignatureFromBytesError> {
```

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -438,26 +680,29 @@ mod serialization {
...
             let bytes = bytes.as_ref();
             let flag =
                 SignatureScheme::from_byte(*bytes.first().ok_or_else(|| {
                     SignatureFromBytesError::new("missing signature scheme flag")
                 })?)
                 .map_err(SignatureFromBytesError::new)?;
```
semenov-vladyslav:
```suggestion
let (flag, tail) = bytes.split_first().ok_or(SignatureFromBytesError::new("missing signature scheme flag"))?;
let scheme = SignatureScheme::from_byte(flag).map_err(SignatureFromBytesError::new)?;
```

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -438,26 +680,29 @@ mod serialization {
...
+
             if flag != SignatureScheme::Multisig {
                 return Err(SignatureFromBytesError::new("invalid multisig flag"));
             }
-            let bcs_bytes = &bytes[1..];
 
-            if let Ok(multisig) = bcs::from_bytes::<Multisig>(bcs_bytes) {
-                Ok(Self {
+            if let Ok(multisig) = bcs::from_bytes::<Multisig>(&bytes[1..]) {
```
semenov-vladyslav:
```suggestion
            if let Ok(multisig) = bcs::from_bytes::<Multisig>(tail) {
```

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -438,26 +680,29 @@ mod serialization {
...
+                let multisig = Self {
                     signatures: multisig.signatures,
                     bitmap: multisig.bitmap,
                     committee: multisig.committee,
-                })
+                    bytes: OnceCell::new(),
+                };
+                multisig
+                    .validate()
```
semenov-vladyslav, line 729: This `validate()` is not enough, it does not validate the `bitmap` field.

crates/iota-sdk-types/src/crypto/multisig.rs
semenov-vladyslav, line 697: There are two ways to "convert multisig to bytes": 1. `multisig.to_bytes()`, and 2. `bcs::to_bytes(&multisig)`. These encodings are not equivalent. Which is the "canonical" one (the one that will generate byte that will end up in a transaction)? Why need for two? This is extremely confusing. The API should not allow to do the "same" operation in two different ways.

semenov-vladyslav (changes requested):

crates/iota-sdk-crypto/src/multisig.rs
thibault-martinez: iota_sdk_types == iota_types, it's just an alias

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: No, we're already under the `proptest` feature that requires `serde`

crates/iota-sdk-types/src/hash.rs
thibault-martinez: It's a convenience method because we repetitively do bcs serialisation + hash of intent messages in the monorepo

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: Why?

crates/iota-sdk-types/Cargo.toml
thibault-martinez: No strong reason, was ported from the monorepo. I'm using the std OnceLock now as OnceCell is `!Sync` which is an issue for the FFI bindings.

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: Wdym two ways? `bcs::to_bytes()` calls `to_bytes()`

crates/iota-sdk-types/src/crypto/multisig.rs
Thoralf-M: Please add some tests to cover the happy and non happy cases here

thibault-martinez: Please enjoy your day off :trollface:

crates/iota-sdk-types/src/crypto/public_key.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/b08ea3764f0d628d6b59cc7c4016b498b72229b8

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/3164bb54e9b45c147475b5020072474e1a102415

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/3164bb54e9b45c147475b5020072474e1a102415

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/f2c18985897b91cd7c1908f38d061f8a84c92761

crates/iota-sdk-crypto/src/multisig.rs
semenov-vladyslav: Ok, didn't notice it was a package alias in Cargo.toml.

crates/iota-sdk-types/src/hash.rs
semenov-vladyslav: I haven't found where it is used and I don't think this method is useful at all. Compare with other implementations of `signing_digest`:
```rs
    fn signing_digest<T: serde::Serialize + ?Sized>(intent: Intent, ty: &T) -> Digest {
        let mut hasher = Hasher::new();
        hasher.update(intent.to_bytes());
        bcs::serialize_into(&mut hasher, ty).unwrap();
        hasher.finalize()
    }
```
Besides the intent bytes it also hashes the content value `ty`. But in `IntentMessage::signing_digest` there is no value at all (or rather it is equivalent to an empty value, eg. `()`). So you will never want to use this method. Maybe, if you added `ty: &T` argument it would make sense.

crates/iota-sdk-types/src/crypto/public_key.rs
semenov-vladyslav: You missed other arms as well: `SignatureScheme::Secp256k1`, `SignatureScheme::Secp256r1`, `SignatureScheme::PasskeyAuthenticator`. They all use `&bytes[1..]`.

crates/iota-sdk-types/src/hash.rs
thibault-martinez: > Maybe, if you added ty: &T argument it would make sense.

But there is literally a T argument, it's wrapped in the `IntentMessage`.
Here is an example in the monorepo of how this method makes things cleaner https://github.com/iotaledger/iota/pull/11505/changes#diff-1a41393988072094a52a6bdba933ba61266f75f08d762b25cf67d2bc6005ba1dL1886-R1887

crates/iota-sdk-types/src/crypto/multisig.rs
semenov-vladyslav: To be symmetric with `pub(crate) fn to_bytes(&self)`. Also, `from_bytes` is already used in `deserialize` and `from_str`, which I assume is enough for users. I might be wrong here, of course

crates/iota-sdk-types/src/crypto/public_key.rs
thibault-martinez: Whoops https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/1c97cf18a38c70e2307f80340a17ed171a5255e7 thanks

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: If I did this, it's probably being used in the monorepo

crates/iota-sdk-types/src/crypto/multisig.rs
semenov-vladyslav: True

crates/iota-sdk-types/src/hash.rs
semenov-vladyslav: You are right. I got confused by the amount of implementations of `signing_digest`.

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: We can't change the validation rules, if such a Multisig already exist on chain, we wouldn't be able to validate it even though it used to be valid.

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: Does not apply anymore, no validation for deserialisation

crates/iota-sdk-types/src/crypto/public_key.rs
thibault-martinez: Fixed the comments

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: It's needed for the monorepo. I'm assuming the name and documentation are enough to express that this should be used with caution.

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: The cost is not the reason, we need the insecure versions as explained earlier

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: We need insecure APIs

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: We need insecure APIs

semenov-vladyslav (approved):

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/b50708a7f0a4ed2dab7a779cd2d763846b084223

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/b50708a7f0a4ed2dab7a779cd2d763846b084223

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -2,85 +2,51 @@
...
 };
-use crate::PublicKeyExt;
+use crate::UserSignature;
 
 pub type WeightUnit = u8;
 pub type ThresholdUnit = u16;
 pub type BitmapUnit = u16;
 
 const MAX_COMMITTEE_SIZE: usize = 10;
 // TODO validate sigs
```
Thoralf-M: Does this still make sense here?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -319,10 +475,10 @@ impl Eq for MultisigAggregatedSignature {}
 /// ed25519-multisig-member-signature               = %d00 ed25519-signature
 /// secp256k1-multisig-member-signature             = %d01 secp256k1-signature
 /// secp256r1-multisig-member-signature             = %d02 secp256r1-signature
-/// zklogin-multisig-member-signature-deprecated    = %d03
-/// passkey-multisig-member-signature               = %d04 passkey-authenticator
+/// zklogin-multisig-member-signature-deprecated    = %d05
+/// passkey-multisig-member-signature               = %d06 passkey-authenticator
```
Thoralf-M: What happened to %d03 %d04 ?

thibault-martinez: PublicKey is a shorter set than SignatureScheme but it uses SignatureScheme as a flag

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -212,11 +235,32 @@ impl MultisigCommittee {
                     .skip(i + 1)
                     .any(|m| member.public_key == m.public_key)
             })
+        {
+            return Ok(());
+        }
+
+        Err(MultisigError::InvalidCommittee)
```
Thoralf-M: Instead of the big && block and just one error it would be better to check each thing individually and return an own error variant for that

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: no, removed

crates/iota-sdk-ffi/src/types/crypto/mod.rs
```diff
@@ -5,6 +5,7 @@ pub mod intent;
 pub mod move_authenticator;
 pub mod multisig;
 pub mod passkey;
+pub mod public_key;
```
Alex6323, line 8: `passkey` and `public_key` (underscore) almost hurts physically, but I know it has to be like this, but calling it `pubkey` is an option as well :trollface:

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -104,18 +69,21 @@ impl MultisigMemberPublicKey {
...
+    public_key: PublicKey,
     weight: WeightUnit,
 }
 
 impl MultisigMember {
-    /// Construct a new member from a `MultisigMemberPublicKey` and a `weight`.
-    pub fn new(public_key: MultisigMemberPublicKey, weight: WeightUnit) -> Self {
-        Self { public_key, weight }
+    /// Construct a new member from a `PublicKey` and a `weight`.
+    pub fn new(public_key: impl Into<PublicKey>, weight: WeightUnit) -> Self {
```
Alex6323, line 88: make `PublicKey` an actual link?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -149,26 +117,57 @@ impl MultisigMember {
...
+    ///  - contain zero or more than ten members;
+    ///  - contain a member with weight 0;
+    ///  - have a `threshold` greater than the sum of all member weights;
+    ///  - contain duplicate public keys.
+    ///
+    /// Note that the order of the members is significant towards deriving the
+    /// [`Address`] governed by this committee.
+    ///
+    /// [`Address`]: crate::Address
+    pub fn new(
```
Alex6323, line 170: we already have `unsafe_` prefix and `_unchecked` suffix to make users aware that they need to be extra careful. I don't really like introducing `insecure_` on top of that tbh.

Alex6323: I would go with `new_unchecked` here because it doesn't check the validity of the created committee.

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -241,40 +252,129 @@ impl MultisigCommittee {
...
+    /// constructor is intended for deserialization paths and tests where the
+    /// inputs are already known to be well-formed.
+    pub fn insecure_new(
         signatures: Vec<MultisigMemberSignature>,
         bitmap: BitmapUnit,
+        committee: MultisigCommittee,
     ) -> Self {
         Self {
             signatures,
             bitmap,
```
Alex6323, line 324: is [struct MultiSig] intended? :thinking:

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -712,22 +911,65 @@ mod tests {
...
+            vec![0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+        );
+    }
+
+    /// `MultisigAggregatedSignature::new` must reject `UserSignature`s that
+    /// are not provided in committee order, since the resulting bitmap and
+    /// signatures vector would otherwise misalign at verification time.
+    #[test]
+    fn new_rejects_out_of_order_signatures() {
+        use crate::{Ed25519PublicKey, Ed25519Signature, SimpleSignature};
```
Alex6323, line 989:
```suggestion
    /// This initializes the underlying bytes representation of MultiSig.
```

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -712,22 +911,65 @@ mod tests {
...
+
```
Alex6323, line 990:
```suggestion
    /// It encodes [struct MultiSig] as the MultiSig flag (0x03) concatenated with the
```

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -241,40 +252,129 @@ impl MultisigCommittee {
...
+    ///
+    /// `signatures` must appear in the same order as their corresponding
+    /// members in `committee`: for committee `[pk1, pk2, pk3, pk4, pk5]`,
+    /// `[sig1, sig2, sig5]` is accepted but `[sig2, sig1, sig5]` is rejected
+    /// with [`MultisigError::SignaturesOutOfOrder`].
+    pub fn new(
+        signatures: Vec<UserSignature>,
+        committee: MultisigCommittee,
+    ) -> Result<Self, MultisigError> {
+        if signatures.len() > committee.members.len() || signatures.is_empty() {
```
Alex6323, line 350: should be `new_unchecked` tbh. If we trust the inputs because they're our own or from a trusted 3rd party there should be nothing insecure about it.

Alex6323: And why does this have to be part of the public API? Couldn't it be `pub(crate)`?

crates/iota-sdk-crypto/src/multisig.rs
```diff
@@ -341,46 +328,37 @@ impl MultisigAggregator {
             },
         );
 
-        Ok(MultisigAggregatedSignature::new(
-            self.committee.clone(),
+        Ok(MultisigAggregatedSignature::insecure_new(
             signatures,
             bitmap,
+            self.committee.clone(),
```
Alex6323, line 387: Is this the only consumer of `insecure_new`? Should this have comment that in this case it's no insecure because the input was already checked?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -2,85 +2,60 @@
...
+use crate::UserSignature;
 
 pub type WeightUnit = u8;
 pub type ThresholdUnit = u16;
 pub type BitmapUnit = u16;
 
 const MAX_COMMITTEE_SIZE: usize = 10;
-// TODO validate sigs
-// const MAX_BITMAP_VALUE: BitmapUnit = 0b1111111111;
+const MAX_BITMAP_VALUE: BitmapUnit = 0b1111111111;
```
Alex6323: I know you didn't change this, but if this is still worth an investigation can we create an issue and link it here?

thibault-martinez: No, was investigated already

crates/iota-sdk-crypto/src/multisig.rs
thibault-martinez: It's not, monorepo uses it, + consumers of the SDK, it's public

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: Needs to be public, used in the monorepo

crates/iota-sdk-types/src/crypto/multisig.rs
Alex6323: I agree. Maybe we can just provide `InvalidCommittee` with an additional message so for the user it becomes:
"invalid committee: threshold == 0",
"invalid committee: no members", etc

crates/iota-sdk-types/src/crypto/multisig.rs
Alex6323: I see that it's used by another crate, so forget about the pub(crate) question.

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -241,40 +276,134 @@ impl MultisigCommittee {
...
+        if signatures.len() > committee.members.len() || signatures.is_empty() {
+            return Err(MultisigError::InvalidSignatureNumber);
+        }
+
+        committee.validate()?;
+
+        let mut bitmap = 0;
+        let mut member_signatures = Vec::with_capacity(signatures.len());
+        let mut prev_index: Option<u8> = None;
+        for (sig_index, signature) in signatures.into_iter().enumerate() {
```
Alex6323, line 359:
```suggestion
        if signatures.is_empty() || signatures.len() > committee.members.len() {
```

sorry for being so picky, but the simpler checks should come first IMHO

thibault-martinez: Not sure what you mean, it's already the case?

thibault-martinez: Ah, the comment doesn't appear on the correct line

thibault-martinez: I'll change if you want but it really does not matter

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -288,16 +417,39 @@ impl MultisigAggregatedSignature {
...
+
     pub fn committee(&self) -> &MultisigCommittee {
         &self.committee
     }
+
+    pub fn has_scheme_signatures(&self, scheme: SignatureScheme) -> bool {
+        self.signatures.iter().any(|s| s.scheme() == scheme)
+    }
+}
+
```
Alex6323, line 441: hm ... bit wondering about this name ... maybe `has_signature_scheme` because it only tests for ONE?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -241,40 +276,134 @@ impl MultisigCommittee {
...
-    /// order as it's corresponding member in the provided committee
-    /// and that it's position in the provided bitmap is set.
-    pub fn new(
-        committee: MultisigCommittee,
+    /// Unlike [`Self::new`], this performs no checks: the `committee` is not
+    /// validated, the `bitmap` is trusted as-is, and the signatures are not
+    /// cross-referenced against the committee. The resulting value may be
+    /// rejected by [`Self::validate`] or by on-chain verification.
+    ///
+    /// The caller must ensure that:
```
Alex6323, line 306: There are some docs missing here for public API getters.

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -340,21 +489,131 @@ impl MultisigMemberSignature {
...
     );
+
+    pub fn scheme(&self) -> SignatureScheme {
+        match self {
+            Self::Ed25519(_) => SignatureScheme::Ed25519,
+            Self::Secp256k1(_) => SignatureScheme::Secp256k1,
+            Self::Secp256r1(_) => SignatureScheme::Secp256r1,
+            Self::Passkey(_) => SignatureScheme::PasskeyAuthenticator,
+        }
+    }
```
Alex6323, line 509: add docs?

crates/iota-sdk-types/src/crypto/multisig.rs
Alex6323: fine with leaving it as it is

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -712,22 +960,163 @@ mod tests {
...
+    fn member_signature_base64_roundtrip() {
+        use crate::Ed25519Signature;
+
+        let sig = MultisigMemberSignature::Ed25519(Ed25519Signature::new([0xAB; 64]));
+        let encoded = sig.to_base64();
+        let decoded = MultisigMemberSignature::from_base64(&encoded)
+            .expect("from_base64 should accept what to_base64 produced");
+        assert_eq!(
+            sig, decoded,
+            "to_base64/from_base64 must be inverses of each other"
```
Alex6323, line 1046: since you also like consistency: you have a different message here than for passkey if the assertion fails

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -712,22 +960,163 @@ mod tests {
...
+        let pk0 = Ed25519PublicKey::new([1; 32]);
+        let pk1 = Ed25519PublicKey::new([2; 32]);
+        let committee = MultisigCommittee::new(
+            vec![MultisigMember::new(pk0, 1), MultisigMember::new(pk1, 1)],
+            2,
+        )
+        .unwrap();
+
+        // 2 sigs, but bitmap bits are at positions 8 and 9 — outside the
+        // committee. count_ones() still matches, so validate() succeeds today.
```
Alex6323: what does "so validate() succeeds today" mean? We validate here that it fails. Ah this wan't the case before?

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -712,22 +960,163 @@ mod tests {
...
+
+        let result = agg.validate();
+        assert!(
+            result.is_err(),
+            "validate() must reject bitmap bits outside committee.members.len(), got Ok"
+        );
+    }
+
+    #[test]
+    fn new_labels_non_adjacent_duplicates_as_duplicates() {
```
Alex6323: I don't understand the adjancency argument ... why does it matter? It should fail no matter where the duplicate is located, no?

Alex6323: ah, now I understand the adjacency limitation: Claude suggested we use the `bitmap` to check for duplicates:

```
let mut bitmap: BitmapUnit = 0;
  let mut prev_index: Option<u8> = None;
  for (sig_index, signature) in signatures.into_iter().enumerate() {
      let pk = signature.to_public_key()...;
      let index = committee.get_public_key_index(&pk)
          .ok_or(MultisigError::NoPublicKeyForSignature(sig_index))?;
      let bit = 1 << index;

      // Duplicate check: have we seen THIS index at any prior position?
      if bitmap & bit != 0 {
          return Err(MultisigError::DuplicatePublicKey);
      }
      // Ordering check: is this index strictly greater than the previous?
      if let Some(prev) = prev_index {
          if index < prev {
              return Err(MultisigError::SignaturesOutOfOrder);
          }
      }

      bitmap |= bit;
      prev_index = Some(index);
      member_signatures.push(signature.try_into()?);
  }
  ```

Comment (thibault-martinez) [reply crates/iota-sdk-types/src/crypto/multisig.rs]: This test is about giving the proper error between Duplicate and Unordered

Comment (thibault-martinez) [reply crates/iota-sdk-types/src/crypto/multisig.rs]: not sure what you snippet is supposed to do, we already check for duplicates

File: crates/iota-sdk-types/src/crypto/passkey.rs [hunk 50]
```diff
@@ -34,39 +34,36 @@ use super::{Secp256r1PublicKey, Secp256r1Signature, SimpleSignature};
...
 }
 
 impl PasskeyAuthenticator {
-    /// Opaque authenticator data for this passkey signature.
-    ///
-    /// See [Authenticator Data](https://www.w3.org/TR/webauthn-2/#sctn-authenticator-data) for
-    /// more information on this field.
-    pub fn authenticator_data(&self) -> &[u8] {
-        &self.authenticator_data
+    /// The passkey public key
```
Alex6323:
```suggestion
    /// The passkey public key.
```

crates/iota-sdk-types/src/crypto/public_key.rs
```diff
@@ -0,0 +1,278 @@
...
+                SignatureScheme::PasskeyAuthenticator => {
+                    let pk = PasskeyPublicKey::new(Secp256r1PublicKey::from_bytes(tail)?);
+                    Ok(Self::Passkey(pk))
+                }
+                _ => Err(PublicKeyError::InvalidInput),
+            },
+            None => Err(PublicKeyError::InvalidInput),
+        }
+
+        // Self::from_bytes()
```
Alex6323: remove?

crates/iota-sdk-types/src/crypto/multisig.rs
Alex6323: but you shipped a commit that exactly fixes this issue. This snippet uses the `bitmap` itself to prevent any duplicates.

Alex6323: this commit changed the logic to what the snippet above also does:

https://github.com/iotaledger/iota-rust-sdk/pull/1137/changes/e4962fca41f4a3023e38ee8ddc223d2bc055fa7f#diff-c2d0682e2c453041f0b9800217c55de11f28bc9a87dc9f35e2731ad9532c51faR366-R372

thibault-martinez: I don't really understand what you want then, I already fixed this

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: You're commenting on a weird line, where was this supposed to be ?

crates/iota-sdk-types/src/crypto/public_key.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/ac86cf771829f061e97caa4cfd111eba375ffef1

crates/iota-sdk-types/src/crypto/passkey.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/ac86cf771829f061e97caa4cfd111eba375ffef1

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: It was claude explaining why something was not working but it's been fixed https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/ac86cf771829f061e97caa4cfd111eba375ffef1

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: I couldn't care less about tests consistency but sure https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/ac86cf771829f061e97caa4cfd111eba375ffef1

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/ac86cf771829f061e97caa4cfd111eba375ffef1

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -712,22 +960,161 @@ mod tests {
...
+            "expected SignaturesOutOfOrder, got {err:?}"
+        );
+
+        // Adjacent duplicates are reported as duplicates, not as ordering errors.
+        let err =
+            MultisigAggregatedSignature::new(vec![sig(pk0), sig(pk0)], committee).unwrap_err();
+        assert!(
+            matches!(err, MultisigError::DuplicatePublicKey),
+            "expected DuplicatePublicKey, got {err:?}"
+        );
```
Alex6323, line 1033: this is not needed? IIUC adjacent duplicates were never flagged as out-of-order anyways, only the non-adjancent ones?

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: I think it's correct tbh. `has_signature_scheme` would be confusing as it would mean the Multisig itself has this scheme which doesn't make sense

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: Pretty sure they were at some point, anyway it's just an assert, it's fine to keep

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -712,22 +960,161 @@ mod tests {
...
+        assert!(
+            result.is_err(),
+            "validate() must reject bitmap bits outside committee.members.len(), got Ok"
+        );
+    }
+
+    #[test]
+    fn new_labels_non_adjacent_duplicates_as_duplicates() {
+        use crate::{Ed25519PublicKey, Ed25519Signature, SimpleSignature};
+
```
Alex6323, line 1099: Can probably be renamed to smth like `new_rejects_duplicates`? And have the cases `pk0, pk0`, or `pk0, pk1, pk0`? The whole adjacent thingy only made sense before the change in e4962fc.

crates/iota-sdk-types/src/crypto/multisig.rs
Alex6323: yeah, it moved unfortunately. around L838 for me currently

crates/iota-sdk-types/src/crypto/multisig.rs
Alex6323: same block of doc comment as the comment above, so around L838

crates/iota-sdk-types/src/crypto/multisig.rs
Alex6323: okay, one last suggestion from my side before I let it go: `contains_signature_scheme`? You know the plural is really a bit off-putting when we just give it ONE scheme to test for.

crates/iota-sdk-types/src/crypto/multisig.rs
Alex6323: I thought your OCD applies to everything :trollface:

crates/iota-sdk-types/src/crypto/multisig.rs
Alex6323: all good, I think you fixed it before my comment. I didn't review the latest state at that time.

Alex6323:
Feel free the resolve my last comments if you don't want to address them. The PR in general looks good to me.

Alex6323 (approved):

Thoralf-M (approved):

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/e4da970dfcd50c8c06eb1316c7b047f379ed8b12

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/af4b0c4881be89875fc5f4eb373544de5b6df7e5

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
```diff
@@ -355,8 +259,8 @@ impl MultisigCommittee {
     ///  - the sum of the weights of all members must be larger than the
```
Thoralf-M: `>=`

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -2,85 +2,62 @@
...
+pub enum MultisigError {
+    #[error("{0}")]
+    TryFromSlice(#[from] std::array::TryFromSliceError),
+    #[error("{0}")]
+    Base64(#[from] base64ct::Error),
+    #[cfg(feature = "serde")]
+    #[error("{0}")]
+    SignatureFromBytes(#[from] SignatureFromBytesError),
+    #[error("Invalid multisig committee")]
+    InvalidCommittee,
```
Thoralf-M: Unused variant now

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -288,16 +419,39 @@ impl MultisigAggregatedSignature {
         self.bitmap
     }
 
+    pub fn get_indices(&self) -> Result<Vec<u8>, MultisigError> {
```
Thoralf-M: Missing doc comment

crates/iota-sdk-ffi/src/types/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/8f7228b486c9715c31d3ef1b285a270ae5e0bfc2

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/8f7228b486c9715c31d3ef1b285a270ae5e0bfc2

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/8f7228b486c9715c31d3ef1b285a270ae5e0bfc2

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/8faee491bdd8b9638c4e7847c2f5112405dc4f1d

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/8faee491bdd8b9638c4e7847c2f5112405dc4f1d

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/8faee491bdd8b9638c4e7847c2f5112405dc4f1d

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/8faee491bdd8b9638c4e7847c2f5112405dc4f1d

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/8faee491bdd8b9638c4e7847c2f5112405dc4f1d

crates/iota-sdk-types/src/crypto/multisig.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1137/commits/8faee491bdd8b9638c4e7847c2f5112405dc4f1d

crates/iota-sdk-types/src/crypto/multisig.rs
```diff
@@ -241,42 +276,136 @@ impl MultisigCommittee {
...
     /// A bitmap that indicates the position of which public key the signature
     /// should be authenticated with.
     bitmap: BitmapUnit,
     /// The public key encoded with each public key with its signature scheme
     /// used along with the corresponding weight.
     committee: MultisigCommittee,
+    /// A bytes representation of this aggregated signature. This helps with
+    /// implementing [trait AsRef<[u8]>].
+    #[cfg(feature = "serde")]
+    bytes: OnceLock<Vec<u8>>,
```
DaughterOfMars, line 294: Do we actually need this or did you just take it because it's that way in the monorepo?

thibault-martinez: It's currently used by the AsRef implementation, I didn't really investigate if it could easily be removed. Can open an issue if you want.

PR #1138, opened 2026-05-11
chore: rename enum opt/mut getters

thibault-martinez:
## Summary

Renames the getters generated by the `def_is_as_into_opt!` macro in [crates/iota-sdk-types/src/lib.rs](crates/iota-sdk-types/src/lib.rs) so the `opt` / `mut` qualifier comes before the variant name rather than after. This puts the modifier next to the verb (`as_` / `into_`) and groups related accessors together alphabetically.

- `as_<variant>_opt` → `as_opt_<variant>`
- `as_<variant>_mut` → `as_mut_<variant>`
- `as_<variant>_mut_opt` → `as_opt_mut_<variant>`
- `into_<variant>_opt` → `into_opt_<variant>`

Call sites updated:

- [crates/iota-sdk-types/src/object.rs](crates/iota-sdk-types/src/object.rs) — `PartialEq` impls on `Owner`.
- [crates/iota-sdk-ffi/src/types/crypto/multisig.rs](crates/iota-sdk-ffi/src/types/crypto/multisig.rs), [crates/iota-sdk-ffi/src/types/object.rs](crates/iota-sdk-ffi/src/types/object.rs), [crates/iota-sdk-ffi/src/types/signature.rs](crates/iota-sdk-ffi/src/types/signature.rs) — FFI wrappers exposing the renamed getters.
- [crates/iota-sdk/examples/package_inspect.rs](crates/iota-sdk/examples/package_inspect.rs).

Breaking at the Rust and FFI API level: any caller of `as_X_opt`, `as_X_mut`, `as_X_mut_opt`, or `into_X_opt` on enums using this macro must move the `opt` / `mut` qualifier in front of the variant name.

https://github.com/iotaledger/iota/issues/11459


DaughterOfMars:
I don't really like this because it means that when a users looks for a specific field the intellisense will have worse suggestions. For example, when you search `is_t` now you might get all the `as_opt` fns instead of `as_transaction`.

thibault-martinez:
> I don't really like this because it means that when a users looks for a specific field the intellisense will have worse suggestions. For example, when you search `is_t` now you might get all the `as_opt` fns instead of `as_transaction`.

Honestly I don't think plugins should be a concern here. This was a request from last spring and I kinda agree with it, this is rust naming convention, e.g. `as_mut_slice`, etc

DaughterOfMars:
I don't believe that there is actually any consistent convention which covers all of the various functions we generate. At the end of the day it's always up to the crate owner and IMO it is much more usable when they are suffixes.

thibault-martinez:
> I don't believe that there is actually any consistent convention which covers all of the various functions we generate. At the end of the day it's always up to the crate owner and IMO it is much more usable when they are suffixes.

I would have to disagree here, after changing them I think it reads better, it actually read as proper english + it matches the return type. Let's see what more people think.

Alex6323:
TBH my first instinct was: "I hate it" (this PR), but then I realized it was due to being so used to the suffix already. 

Then I wanted to hate on having the `opt` infix/suffix at all trying to make an argument that for `mut` it's fine to be an infix, bc that's a keyword, not a type, but very often `Option`s become `Result`s and vice versa, and then the method would have to be renamed as well, which is breaking if it's a public API.

I then thought, `Option` should be consistent with `Result`. One could make an argument that for `Result`s one uses the `try_` prefix, so for `Option`s the consistent thing to do would be to have a `maybe_` or `opt_` prefix, so for instance `maybe_as_v1`/`maybe_as_mut_v1` (analogous to `try_as_v1`/`try_as_mut_v1` if that were a thing). 

Finally I think one can make valid arguments for either of the available options, and none would be entirely superior to the others, so, unfortunately, in the end it boils down to what most people like. The human brain will get used to almost anything eventually.

I personally would prefer to have `Result` and `Option` treated consistently, either with the prefix as above, or not prefixed/infixed/suffixed at all. So there you have it. My 2 cents!

crates/iota-sdk-types/src/object.rs
```diff
@@ -173,14 +173,14 @@ pub enum ObjectData {
...
-    pub fn object_type(&self) -> Option<&MoveObjectType> {
+    pub fn opt_object_type(&self) -> Option<&MoveObjectType> {
         match self {
             Self::Struct(m) => Some(m.object_type()),
             Self::Package(_) => None,
         }
     }
 
-    pub fn struct_tag(&self) -> Option<StructTag> {
+    pub fn opt_struct_tag(&self) -> Option<StructTag> {
```
Thoralf-M, line 180: Moved my comment from the iota repo PR https://github.com/iotaledger/iota/pull/11502#discussion_r3220596562 here Should there also be a `as_` prefix here or should it get removed from the others?

thibault-martinez: I intentionally didn't put `as_` here because it doesn't do what the `as_` methods do, it doesn't simply return the variant. To be discussed.

DaughterOfMars: Maybe `get_` then?

Thoralf-M: I see, fine for me like it is or also with `get_`

Thoralf-M (approved):

Alex6323 (approved):

PR #1140, opened 2026-05-12
chore: Version::INITIAL

thibault-martinez:


crates/iota-sdk-types/src/version.rs
```diff
@@ -58,8 +58,8 @@ impl Version {
     /// is assigned to, does not appear in a cancelled transaction.
     pub const MIN_VALID_INCL: Self = Self(u64::MIN);
 
-    /// The initial shared version for shared system objects.
-    pub const INITIAL_SHARED_VERSION: Self = Self(1);
+    /// The initial version for objects.
```
Thoralf-M, line 61: Not sure how to phrase it best, but should probably somehow mention that not all objects have this as the initial version (objects created in transactions later will just get the highest version +1 as their initial version...)

PR #1147, opened 2026-05-13
feat(gRPC)!: improve client ergonomics & read-mask discoverability

### LINKED ISSUE #1142: Add additional grpc helper fns
## Description

## Alternatives

## Referencing

DaughterOfMars:
Follow-up to the initial gRPC client landing (#1062). This PR sharpens the
high-level Rust gRPC API around two recurring rough edges — every read method
having to take an `Option<ReadMask>`, and callers having to know by convention
which `XField` namespace pairs with which method — and adds a handful of
examples that exercise the result.

## Summary

- **Split every masked endpoint into a default-mask method + a `_masked`
  variant.** `client.get_objects(refs)` uses the canonical default mask;
  `client.get_objects_masked(refs, mask)` takes a mask directly (no
  `Option`, no wrapper). Applied to all 17 masked methods across
  `get_objects` / `get_objects_with_versions` / `get_transactions` /
  `get_service_info` / `get_epoch` / `get_checkpoint_*` / `stream_checkpoints*`
  / `list_owned_objects` / `list_dynamic_fields` / `get_coins` /
  `simulate_transaction(s)` / `execute_transaction(s)`.
- **Typed scoped read masks per endpoint.** Each `XField` (e.g.
  `ObjectField`, `TransactionField`, `EpochField`, …) is now a typed
  `Cow<'static, str>` newtype instead of a bag of `&str` consts, and each
  masked endpoint accepts a paired `XReadMask` (`ObjectReadMask`,
  `TransactionReadMask`, …) via `impl Into<…>`. Call sites become:
  ```rust
  client.get_objects_masked([id], ObjectField::BCS).await?;
  client.get_objects_masked([id], [ObjectField::REFERENCE, ObjectField::BCS]).await?;
  ```
  No `ReadMask::from(…)` wrapper. Passing `TransactionField::EFFECTS` to
  `get_objects_masked` is now a compile error. Pre-computed mask constants
  in `read_masks::*` still flow through via `From<&'static str>` /
  `From<String>` escape hatches on each scoped mask.
- **New `Client::get_coins(_masked)`** plus `GetCoinsQuery` builder in
  `state/coins.rs` — paginated coin listing with optional `StructTag`
  filter, returning `iota_types::framework::Coin` items (proto `Object` →
  SDK `Coin` conversion happens internally).
- **`CheckpointResponse` accessors refactored to return SDK types directly.**
  `response.summary()` / `.signature()` / `.contents()` now yield the
  deserialized SDK types; the proto variants are renamed to `_grpc`
  (`summary_grpc`, etc.). Same pattern as the `executed_transactions` /
  `executed_transactions_grpc` split.
- **Network constructors on `Client`**: `Client::new_localnet()`,
  `new_devnet()`, `new_testnet()`, `new_mainnet()` plus pub-crate host
  constants.
- **TLS features forwarded from `iota-sdk`'s `grpc` feature.** Previously
  enabling `iota-sdk/grpc` pulled in `iota-grpc-client` with
  `default-features = false`, so `Client::new_testnet()` failed at runtime
  with `"HTTPS requires the tls-ring feature"`. The `grpc` feature now
  forwards `tls-ring` + `tls-native-roots`.
- **Four new gRPC examples** in `crates/iota-sdk/examples/`:
  `chain_id_grpc`, `get_object_grpc`, `owned_objects_grpc`,
  `stream_checkpoints_grpc` — the last has no GraphQL counterpart and
  exists specifically to showcase the streaming RPC.

## Migration

This is a breaking change to the gRPC client surface. The mechanical
rewrite at call sites:

| Before | After |
|---|---|
| `get_objects(refs, None)` | `get_objects(refs)` |
| `get_objects(refs, Some(ReadMask::from(ObjectField::BCS)))` | `get_objects_masked(refs, ObjectField::BCS)` |
| `get_objects(refs, Some(ReadMask::from(&[ObjectField::BCS, ObjectField::REFERENCE])))` | `get_objects_masked(refs, [ObjectField::BCS, ObjectField::REFERENCE])` |
| `cp.summary()?.summary()?` | `cp.summary()?` |
| `cp.signature()?.signature()?` | `cp.signature()?` |
| `cp.contents()?.contents()?` | `cp.contents()?` |

`XField::Y` used to evaluate to `&'static str`; it's now an `XField`
struct. Use `.as_str()` if you need the underlying path string.

`ReadMask<'_>` is retained as a low-level string holder for advanced
composition but is no longer accepted by client methods directly — pass a
scoped mask (or a `&str` for the escape hatch) instead.


crates/iota-sdk-grpc-client/src/api/mod.rs
```diff
@@ -179,12 +204,20 @@ impl CheckpointResponse {
...
     /// [`CHECKPOINT_RESPONSE_EXECUTED_TRANSACTIONS`])
-    pub fn executed_transactions(
+    pub fn executed_transactions_grpc(
         &self,
     ) -> &Vec<iota_grpc_types::v1::transaction::ExecutedTransaction> {
         &self.executed_transactions
     }
 
+    /// Get the executed transactions in this checkpoint.
+    pub fn executed_transactions(&self) -> Result<Vec<iota_types::Transaction>> {
```
Thoralf-M: Wondering how interesting that is without the tx effects, why did you add it?

DaughterOfMars: I think I meant to revert this file and forgot

crates/iota-sdk/examples/chain_id_grpc.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2025 IOTA Stiftung
```
Thoralf-M: 2026 also in the other examples

crates/iota-sdk-grpc-types/src/read_mask_fields.rs
```diff
@@ -1,305 +1,469 @@
...
-    /// Timestamp.
-    pub const TIMESTAMP: &str = "executed_transaction.timestamp";
-    /// Input objects (all sub-fields).
-    pub const INPUT_OBJECTS: &str = "executed_transaction.input_objects";
-    /// The full BCS-encoded input object.
-    pub const INPUT_OBJECTS_BCS: &str = "executed_transaction.input_objects.bcs";
-    /// Output objects (all sub-fields).
-    pub const OUTPUT_OBJECTS: &str = "executed_transaction.output_objects";
-    /// The full BCS-encoded output object.
-    pub const OUTPUT_OBJECTS_BCS: &str = "executed_transaction.output_objects.bcs";
```
Thoralf-M, line 237: Why are they removed from SimulateExecutedTransactionField?

crates/iota-sdk-grpc-client/src/api/state/coins.rs
```diff
@@ -0,0 +1,306 @@
...
+        field_mask_with_default, saturating_usize_to_u32,
+    },
+};
+
+/// Builder for listing coins owned by an address.
+///
+/// Created by [`Client::get_coins`]. Await directly for a single page
+/// (with access to `next_page_token`), or call
+/// [`.collect(limit)`](Self::collect) to auto-paginate.
+pub struct GetCoinsQuery {
```
Thoralf-M: Modify and make use of `define_list_query!` macro for this? It's almost the same and a lot duplication right now

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
```diff
@@ -374,22 +478,51 @@ impl Client {
...
+    /// or any slice/array/vec of fields — conversion is automatic.
+    #[allow(clippy::too_many_arguments)]
+    pub async fn stream_checkpoints_filtered_masked(
+        &self,
+        start_sequence_number: impl Into<Option<CheckpointSequenceNumber>>,
+        end_sequence_number: impl Into<Option<CheckpointSequenceNumber>>,
+        read_mask: impl Into<CheckpointResponseReadMask>,
+        transactions_filter: impl Into<Option<grpc_filter::TransactionFilter>>,
+        events_filter: impl Into<Option<grpc_filter::EventFilter>>,
+        progress_interval_ms: impl Into<Option<u32>>,
```
Thoralf-M, line 515: Instead of all the individual fields, how about a `StreamCheckpointsBuilder`?

crates/iota-sdk-grpc-client/Cargo.toml
```diff
@@ -21,7 +21,7 @@ tonic = { workspace = true, features = ["zstd"] }
 
 # internal dependencies
 iota-grpc-types.workspace = true
-iota-types = { workspace = true, features = ["serde", "hash"] }
+iota-types = { workspace = true, features = ["serde", "hash", "rand"] }
```
Thoralf-M: You added rand dependency only for the two doc examples that are no_run?
Should remove it again, fetching txs by random digest won't work anyways, the `todo!()` was fine there I think

Thoralf-M:
Overall not sure if this PR is really worth it, just providing `None` for the read mask if not needed or finding the correct things to mask doesn't seem to be a big deal to me

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
DaughterOfMars: Not a bad idea, but could you create an issue for it?

DaughterOfMars:
> Overall not sure if this PR is really worth it, just providing `None` for the read mask if not needed or finding the correct things to mask doesn't seem to be a big deal to me

I think it's a pretty big usability increase, if not for the extra helper fns than definitely for the read mask discoverability. Before you basically had no way to know what read masks to use other than reading the doc. Now it's implicit.

crates/iota-sdk-grpc-types/src/read_mask_fields.rs
DaughterOfMars: Somehow Claude missed these, but they are fixed now

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/issues/1187

Thoralf-M (approved):

muXxer:
Thank you very much for all the improvements on the grpc client.

I don't want to be picky, but I think the scope of this PR is way too big. There are so many changes that deserve their own PRs and discussions. I think we should rather create a feature branch and have scoped PRs for changes.

What I wouldn't do is to double the API surface with the new "masked" functions. Improvements in mask discoverability are welcome, but we should avoid doubling the API surface just because of an Option. Users should think about the mask they pass, to adapt it to the information they really need, because it reduces the network bandwidth and also the resource consumption on the server side.

DaughterOfMars:
> Thank you very much for all the improvements on the grpc client.
> 
> I don't want to be picky, but I think the scope of this PR is way too big. There are so many changes that deserve their own PRs and discussions. I think we should rather create a feature branch and have scoped PRs for changes.
> 
> What I wouldn't do is to double the API surface with the new "masked" functions. Improvements in mask discoverability are welcome, but we should avoid doubling the API surface just because of an Option. Users should think about the mask they pass, to adapt it to the information they really need, because it reduces the network bandwidth and also the resource consumption on the server side.

I could split this into two PRs, but regardless I do believe both changes should happen. It's no cost to have extra functions that make calling the API easier for users. Using the masks is a power-user feature, IMO, regardless of this PR.

Alex6323 (approved):

DaughterOfMars:
Replaced by #1253 and #1254

Issue #1160, opened 2026-05-19
Auto gas-selection picks every gas coin, blowing past the GraphQL 5000-byte `query part` limit


Thoralf-M:
## Summary

When the transaction builder is used without an explicit `gas(...)` pin, it auto-selects gas coins by **paging through every gas coin the sender owns and pinning all of them as gas-payment inputs** — there is no "stop once budget is covered" check. For an address that owns ~40 gas coins, the resulting `dryRunTransactionBlock` request exceeds the GraphQL server's 5000-byte `query part` cap and the call fails:

```
Query error: [Query part too large: 6992 bytes. Requests are limited to 174763 bytes or fewer on transaction payloads (all inputs to executeTransactionBlock or dryRunTransactionBlock) and the rest of the request (the query part) must be 5000 bytes or fewer.]
```

This is reproducible with a trivial `send_iota` transaction once the sender has accumulated enough gas coins — something that happens easily on a shared development address (e.g. the address used by the SDK examples after a few faucet calls).

## Where

`crates/iota-sdk-transaction-builder/src/builder/mod.rs` — `ClientTransactionBuilder::resolve_ptb`, lines ~1088-1107:

```rust
for coin in self
    .client
    .objects(
        Some(StructTag::new_gas_coin().into()),
        Some(self.data.sponsor.unwrap_or(self.data.sender)),
        None, true, None, None,
    )
    .await
    .map_err(Error::client)?
{
    if !unusable_object_ids.contains(&coin.object_id()) {
        self.set_input(
            InputKind::Input(iota_types::Input::ImmutableOrOwned(coin.object_ref())),
            true,
        );
    }
}
```

There's no upper bound on the number of coins added, no preference for fewer/larger coins, and no balance-vs-budget check.

## Reproduction

Self-contained reproduction (`Cargo.toml` + `src/main.rs`). Requires a fresh localnet (`./run_localnet.sh start ./iota-localnet`).

`Cargo.toml`:
```toml
[package]
name = "repro-query-too-large"
version = "0.1.0"
edition = "2024"

[dependencies]
iota-sdk = { path = "../iota-rust-sdk/crates/iota-sdk", features = ["graphql", "txn-builder", "types", "crypto"] }
tokio = { version = "1", features = ["full"] }
eyre = "0.6"
```

`src/main.rs`:
```rust
use std::str::FromStr;
use eyre::{OptionExt, Result};
use iota_sdk::{
[...]

PR #1161, opened 2026-05-19
feat(examples): switch all to localnet to not rely on testnet

Thoralf-M:
Testnet indexer is not always 100% reliable reachable and caused CI failures


Thoralf-M:
Blocked by https://github.com/iotaledger/iota-rust-sdk/issues/1160

PR #1162, opened 2026-05-19
fix(txn-builder): paginate auto gas-coin selection

### LINKED ISSUE #1160: Auto gas-selection picks every gas coin, blowing past the GraphQL 5000-byte `query part` limit

Thoralf-M:
Auto gas-payment selection used to fetch every gas coin owned by the sender and pin all of them, which on a wallet with many coins blew past the GraphQL server's 5000-byte `query part` limit.

Changes:

* `ClientMethods::objects` now returns a single page plus an opaque continuation cursor (`Option<Vec<u8>>`) instead of an unbounded list, so callers paginate explicitly.
* The trait gains a `protocol_config` hook returning a transport-neutral `ProtocolConfig { attributes: BTreeMap<String, String> }` — both GraphQL (`Vec<ProtocolConfigAttr>`) and gRPC (`map<string, string>`) collapse into the same shape, so the type does not enumerate every on-chain field. The default impl returns an empty config so existing implementors don't have to grow protocol-config plumbing.
* Auto gas selection now pages through the owner's gas coins, keeping a running top-K by balance (K = the protocol cap). It stops as soon as the selected coins cover the requested budget, or the pages run out; with no budget set it walks every page. The selection deliberately over-includes — in the common case a wallet owns fewer coins than the cap, so the whole set is pinned — and gas smashing during execution consolidates the balances into a single coin.
* The cap (K) comes from the chain's `max_gas_payment_objects` protocol attribute via the new `protocol_config` hook, falling back to a hardcoded 255 (256 exclusive at the time of writing) when the implementation doesn't expose protocol config.

Integration tests:

* `test_auto_gas_selection_with_many_coins` funds an address with ~2044 small coins (4 × `split_coins(.., 511)` — one argument slot per command goes to the coin, leaving 511 for amounts) and runs a plain `send_iota`, exercising the pagination path.
* `test_auto_gas_pins_full_first_page_for_consolidation` mints 50 coins of 1 IOTA each — any single one trivially covers the gas budget — and asserts the resolved transaction pins all 50, proving the selection over-includes for consolidation rather than picking the minimum.
* `test_manual_gas_pin_consolidates_255_coins` pins all 255 gas coins (the protocol cap) and asserts the gas-smashing prologue consolidates them into a single coin during execution.

Fixes #1160

---
_Generated by [Claude Code](https://claude.ai/code/session_012iDqaCndUm7frUzsvrihQj)_


crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -37,6 +37,34 @@ pub trait ClientMethods {
...
+    /// Used by the transaction builder's automatic gas-payment selection to
+    /// page through gas coins, stopping once cumulative balance meets the
+    /// target budget. The default implementation falls back to
+    /// [`Self::objects`] and reports no next cursor; clients with native
+    /// pagination support should override it.
+    fn gas_coins_page(
+        &self,
+        owner: Address,
+        cursor: Option<String>,
+    ) -> impl std::future::Future<Output = Result<(Vec<Object>, Option<String>), Self::Error>> {
```
Alex6323: all those desugared trait methods can use the `async` keyword, no?

Thoralf-M: Tried that — `async fn` in a `pub trait` declaration warns about auto-trait bounds (`Future: Send`) being unspecifiable, which the rest of `ClientMethods` deliberately works around with the `impl Future` form. I'll keep `gas_coins_page` consistent with the surrounding methods here; the impls (`&T`, `Arc<T>`, the GraphQL client, the FFI client) already use `async fn` since the warning only applies to the trait declaration.

---

Alex6323: okay Claude :trollface:

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -11,6 +13,19 @@ use iota_types::{
     UserSignature, Version,
 };
 
+/// One page of objects plus an optional cursor for the next page. See
+/// [`ClientMethods::objects`].
+pub type ObjectsPage = (Vec<Object>, Option<Vec<u8>>);
```
DaughterOfMars: Why not a struct?

Thoralf-M: No strong reason, changed to struct now https://github.com/iotaledger/iota-rust-sdk/pull/1162/commits/cfc507f5796a2296c049176e1edb036081679545

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -1085,25 +1096,74 @@ impl<C: ClientMethods, L> TransactionBuilder<C, L> {
...
+                for obj in page.data {
+                    if unusable_object_ids.contains(&obj.object_id()) {
+                        continue;
+                    }
+                    let Ok(coin) = Coin::try_from_object(&obj) else {
+                        continue;
+                    };
+                    selected.push((coin.balance(), obj.object_ref()));
+                }
+                selected.sort_by_key(|c| std::cmp::Reverse(c.0));
```
DaughterOfMars, line 1144: the sort seems to be in the wrong place, shouldn't it be after the full loop?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -1085,25 +1096,74 @@ impl<C: ClientMethods, L> TransactionBuilder<C, L> {
...
+                selected.truncate(max_gas_payment_objects);
```
DaughterOfMars, line 1145: Is this actually necessary?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
Thoralf-M: It's here related to your next comment with the truncate, it makes sure that we only keep the top coins with most amount in memory and limit to what the protocol allows as gas payments
I guess for most users it will run only once anyways since the first page will have enough coins

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -26,16 +49,34 @@ pub trait ClientMethods {
...
         limit: Option<usize>,
-    ) -> impl std::future::Future<Output = Result<Vec<Object>, Self::Error>>;
+    ) -> impl std::future::Future<Output = Result<ObjectsPage, Self::Error>>;
+
+    /// Fetch the chain's protocol configuration. Implementations should
+    /// populate [`ProtocolConfig::attributes`] with the protocol config
+    /// attributes; callers look up the keys they care about.
+    ///
+    /// The default impl returns an empty [`ProtocolConfig`] so existing
+    /// implementors do not have to grow protocol-config plumbing.
```
thibault-martinez: "to grow" is a bit weird

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: But why wouldn't this be done after the complete loop?

Thoralf-M: It could be, but then the vec size could get much larger. This way it's limited to max protocol inputs limit + page size

Thoralf-M: And you still need to go over the vec in the loop to check if you already have enough funds

DaughterOfMars: But there are two problems
1. The sort here means the final vec will not be sorted
2. If there are a lot of gas objects which do not cover the cost then this will keep looping anyway for no reason

Thoralf-M: 1. the final vec doesn't need to be sorted
2. How can you know that there is not a gas object in the last page that will have sufficient gas? That's exactly what this is for, that it will go through potentially many pages and find coins that are suitable. You can only know that there are none, if you checked all of them

DaughterOfMars: But am I wrong in thinking that those objects aren't going to be kept after truncating? You aren't sorting by balance...

Thoralf-M: What do you mean?
```Rust
                    selected.push((coin.balance(), obj.object_ref()));
                }
                selected.sort_by_key(|c| std::cmp::Reverse(c.0));  
```

DaughterOfMars: Ah, I see. I misread it. In this case we could use a BTreeMap instead of sorting, couldn't we?

DaughterOfMars: Ah, actually, nevermind, the map won't work. But maybe we could insert sorted to avoid the sorting afterwards?

Thoralf-M: I don't think that's worth it, the vec size is always small (<=305 if pagesize is 50) and performance shouldn't really be noticeable
I actually did some benchmark and at this size the current way should be faster, sorted insert only wins at >1000 coins

## Top-K gas-coin selection benchmark (random balances)

K = 255, page size = 50. min of 500–5000 iters, release + LTO.

| coins | A sort-after-page | B sorted-insert | C heap | B vs A |
|------:|------:|------:|------:|:--|
| 305 | 4.61 µs | 6.60 µs | 2.73 µs | A wins (1.4×) |
| 1 000 | 17.3 µs | 18.1 µs | 6.42 µs | tie |
| 20 000 | 342 µs | 183 µs | 42.5 µs | B wins (1.9×) |
| 100 000 | 1.39 ms | 0.80 ms | 0.13 ms | B wins (1.7×) |

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -26,16 +49,34 @@ pub trait ClientMethods {
...
         owner: Option<Address>,
         object_ids: Option<Vec<ObjectId>>,
         ascending: bool,
-        cursor: Option<String>,
+        cursor: Option<Vec<u8>>,
         limit: Option<usize>,
-    ) -> impl std::future::Future<Output = Result<Vec<Object>, Self::Error>>;
+    ) -> impl std::future::Future<Output = Result<ObjectsPage, Self::Error>>;
+
+    /// Fetch the chain's protocol configuration. Implementations should
```
Alex6323: Can you line break after the first sentence. Instructions on what implementors should do should be regarded as additional information. Also ... I don't really like what Claude wrote afterwards. Seems more like it leaked its reasoning into the docs of why it implemented it like this, not something that a user of this trait would need. It's enough to mention that if this method is not implemented it returns a default ProtocolConfig.

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -272,13 +345,19 @@ impl<T: ClientMethods> ClientMethods for std::sync::Arc<T> {
...
-        cursor: Option<String>,
+        cursor: Option<Vec<u8>>,
         limit: Option<usize>,
-    ) -> impl std::future::Future<Output = Result<Vec<Object>, Self::Error>> {
+    ) -> impl std::future::Future<Output = Result<ObjectsPage, Self::Error>> {
         self.as_ref()
             .objects(type_tag, owner, object_ids, ascending, cursor, limit)
     }
 
+    fn protocol_config(
```
Alex6323, line 352: I know the other methods don't use it either, but why don't we use the `async` keyword here? Don't think it's worth an extra PR, so maybe change it?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: Alrighty, well proven 😆

DaughterOfMars (approved):

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
Thoralf-M: Still the same as https://github.com/iotaledger/iota-rust-sdk/pull/1162#discussion_r3266317792

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
Thoralf-M: Sure, changed in https://github.com/iotaledger/iota-rust-sdk/pull/1162/changes/daa0346fe048cb4797e34e69ce55ca1c097db92e

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
Alex6323: no, that's the impl part, not the trait definition

Alex6323: seem it's my bad.

Alex6323 (approved):

PR #1164, opened 2026-05-19
chore: warn `clippy::unused_async`

thibault-martinez:
CI will pass when https://github.com/iotaledger/iota-rust-sdk/pull/580 is merged

### DIFF
--- a/crates/iota-sdk-grpc-client/src/client.rs
+++ b/crates/iota-sdk-grpc-client/src/client.rs
@@ -39,7 +39,7 @@ pub struct Client {
 
 impl Client {
     /// Create a new Client instance for the given gRPC server URI.
-    pub async fn new<T>(uri: T) -> Result<Self>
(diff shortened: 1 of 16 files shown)


Alex6323 (approved):

crates/iota-sdk-grpc-client/src/client.rs
```diff
@@ -227,7 +227,7 @@ mod tests {
     async fn https_without_tls_ring_returns_failed_precondition() {
```
Thoralf-M: Now this also doesn't need to be async anymore?

thibault-martinez: Good catch, `tokio::test` was consuming the async keyword so clippy couldn't detect it, make it a simple test https://github.com/iotaledger/iota-rust-sdk/pull/1164/commits/f629d68a753a27efdc660f124bc5d9a267b1d4f6

Thoralf-M (approved):

PR #1170, opened 2026-05-21
feat: expose Error::kind() with HTTP status and decode-target context

Thoralf-M:
## Summary

`develop` already surfaces the URL, HTTP status, and (truncated) response body in GraphQL error **messages**. This PR makes that failure information **programmatically inspectable** and keeps the message **self-describing**:

- Add a public `Error::kind()` accessor — the `Kind` enum previously had no getter and was effectively display-only.
- Change `Kind::Http` to `Kind::Http { status }` so callers can read the `StatusCode` and branch on it (e.g. retry on 429/503).
- Record the decode target type via a backward-compatible `Error::while_decoding` wither, so the message also shows `` (while decoding `T`) `` — the client tags errors with `std::any::type_name::<R>()` where the generic type is known.

`Kind` is `#[non_exhaustive]` and the existing `Error::http`/`Error::decode` constructor signatures are unchanged, so this is additive (no breaking changes).

## How it looks now

Non-2xx HTTP response (`Kind::Http { status }`):
```
HTTP 504 Gateway Timeout: GraphQL request to https://graphql.testnet.iota.cafe/ failed, body="Gateway Timeout" (while decoding `cynic::result::GraphQlResponse<…::PackageQuery>`)
```

2xx response whose body is not valid JSON (`Kind::Deserialization`):
```
Deserialization error: GraphQL request to https://graphql.testnet.iota.cafe/ returned HTTP 200 OK but the body could not be parsed as JSON (body="<html>Bad Gateway</html>"): expected value at line 1 column 1 (while decoding `cynic::result::GraphQlResponse<…::PackageQuery>`)
```

## Programmatic access

```rust
match client.chain_id().await {
    Err(e) if matches!(e.kind(), Kind::Http { status } if status.is_server_error()) => {
        // retry on 5xx
    }
    _ => { /* ... */ }
}
```

## Test plan
- [x] `cargo check -p iota-sdk-graphql-client`
- [x] `cargo test -p iota-sdk-graphql-client` — `error::` (6) and `response_to_err` (3) unit tests pass
- [x] `cargo clippy -p iota-sdk-graphql-client --all-targets` clean
- [x] `cargo +nightly fmt -p iota-sdk-graphql-client -- --check` clean


crates/iota-sdk-graphql-client/src/error.rs
```diff
@@ -39,6 +39,10 @@ pub enum Kind {
     Query,
     Missing,
     Other,
+    /// The HTTP response carried a non-success status code.
+    Http {
+        status: reqwest::StatusCode,
+    },
```
Alex6323, line 66: This variant should probably come before `Other`?

crates/iota-sdk-graphql-client/src/error.rs
```diff
@@ -107,6 +117,7 @@ impl std::fmt::Display for Kind {
             Kind::Query => write!(f, "Query error:"),
             Kind::Missing => write!(f, "Missing:"),
             Kind::Other => write!(f, "Error:"),
+            Kind::Http { status } => write!(f, "HTTP {status}:"),
```
Alex6323, line 185: here as well. Reads a bit weird that after `Other` there's a regular variant again.

crates/iota-sdk-graphql-client/src/client.rs
```diff
@@ -180,6 +176,38 @@ impl Client {
...
+    let url = response.url().clone();
+    let status = response.status();
+    let body = response.bytes().await?;
+    let body_str = String::from_utf8_lossy(&body);
+    let type_name = std::any::type_name::<T>();
+
+    if !status.is_success() {
+        return Err(Error::from_message(
+            Kind::Http { status },
+            format!("{url} while decoding `{type_name}`; body: {body_str:?}"),
```
Alex6323: is the message intended like this with the `url` first? Shouldn't it say something like "Request failed for {url} while ...."?

crates/iota-sdk-graphql-client/src/error.rs
Thoralf-M: Moved

crates/iota-sdk-graphql-client/src/error.rs
Thoralf-M: Moved

crates/iota-sdk-graphql-client/src/client.rs
Thoralf-M: Changed

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1171, opened 2026-05-21
feat: `PasskeyAuthenticator` changes

thibault-martinez:
## Summary

A few small, mostly-ergonomic additions around `PasskeyAuthenticator` and the key/address types in `iota-sdk-types`:

- **`PasskeyAuthenticator` is now hashable.** Adds `Hash` and `AsRef<[u8]>`, backed by a lazily-initialized `OnceLock` cache of the serialized bytes, so the authenticator can be used in hash-based collections (and compared as bytes) without re-serializing on every access.
- **`new` reports errors instead of dropping them.** `PasskeyAuthenticator::new` now returns `Result<_, SignatureFromBytesError>` rather than `Option`, surfacing why construction failed.
- **`From<…> for Address`.** Added `From` conversions to `Address` for every key type (`Ed25519`/`Secp256k1`/`Secp256r1`/`Passkey` public keys, the `PublicKey` enum, and `MultisigCommittee`), each forwarding to `derive_address()`.
- **`Digest` ⇄ `&[u8]` equality.** `PartialEq` between `Digest` and `&[u8]` in both directions.

## Test plan

- `make test`
- `make clippy`


crates/iota-sdk-types/src/crypto/passkey.rs
```diff
@@ -33,12 +38,14 @@ use super::{Secp256r1PublicKey, Secp256r1Signature, SimpleSignature};
...
+    /// Compact r1 public key for this passkey.
     pub(crate) public_key: Secp256r1PublicKey,
-    /// The secp256r1 signature from the passkey.
+    /// Normalized r1 signature from the passkey.
     pub(crate) signature: Secp256r1Signature,
     /// Parsed base64url decoded challenge bytes from
-    /// `client_data_json.challenge`.
+    /// `client_data_json.challenge`, which is expected to be the signing
+    /// message `hash(Intent | bcs_message)`
+    // TODO  challenge: [u8; DefaultHash::OUTPUT_SIZE] ?
```
Thoralf-M: TODO when?

crates/iota-sdk-types/src/crypto/passkey.rs
```diff
@@ -230,6 +252,12 @@ mod serialization {
                 origin: _,
             } = serde_json::from_str(&client_data_json).map_err(SignatureFromBytesError::new)?;
 
+            // if client_data_json_parsed.ty != ClientDataType::Get {
+            //     return Err(IotaError::InvalidSignature {
+            //         error: "Invalid client data type".to_string(),
+            //     });
+            // };
```
Thoralf-M: Delete if not needed

crates/iota-sdk-types/src/crypto/passkey.rs
thibault-martinez: I'm currently investigating

Thoralf-M: Why did you remove instead of changing to have a fixed size?

thibault-martinez: Because the passkey standard accepts any kind of challenge, it just happens that we check against a digest. Parsing accepts any kind of size but it's at verification time that eventually the size is checked to be 32.

Thoralf-M: But if we check during verification for the length, what does someone gain if he can construct a challenge that won't pass this step?

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1171/changes/3ce592164f64409b0a41389955a4b49f0aa977b3

crates/iota-sdk-types/src/crypto/passkey.rs
```diff
@@ -50,15 +59,19 @@ pub struct PasskeyAuthenticator {
     /// See [CollectedClientData](https://www.w3.org/TR/webauthn-2/#dictdef-collectedclientdata)
     /// for more information on this field.
     pub(crate) client_data_json: String,
+    /// Initialization of bytes for passkey in serialized form.
+    bytes: OnceLock<Vec<u8>>,
```
Thoralf-M, line 65: That's now also used when comparing with `PartialEq`

crates/iota-sdk-types/src/crypto/passkey.rs
```diff
@@ -91,6 +105,18 @@ impl PasskeyAuthenticator {
     }
 }
 
+impl Hash for PasskeyAuthenticator {
```
Thoralf-M, line 120: Needs to be behind the serde feature? Because `.as_ref()` calls `self.to_bytes()` which is defined in the serialization block

crates/iota-sdk-types/src/crypto/passkey.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1171/changes/59ad11d9f3747c91dd2b1b466836a2109d0077f3

crates/iota-sdk-types/src/crypto/passkey.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1171/commits/3f6f875e4518c442c30eb8bfdb773661b26f6a48

crates/iota-sdk-types/src/crypto/passkey.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1171/changes/b7708d03c50069d762f74622a1ba8eb6f9b75498

crates/iota-sdk-types/src/crypto/passkey.rs
```diff
@@ -50,15 +60,20 @@ pub struct PasskeyAuthenticator {
...
 
 impl PasskeyAuthenticator {
     /// The passkey public key.
+    // TODO maybe split this in public_key and passkey_public_key?
     pub fn public_key(&self) -> PasskeyPublicKey {
         PasskeyPublicKey::new(self.public_key)
     }
 
     /// The passkey signature.
+    // TODO maybe split this in secp256r1_signature and simple signature?
```
Thoralf-M: Want to do that or not?

thibault-martinez: No https://github.com/iotaledger/iota-rust-sdk/pull/1171/commits/16a4400c8e403152d47d8a2916d1fbdc2ebdcae8

Thoralf-M (approved):

crates/iota-sdk-types/src/crypto/passkey.rs
```diff
@@ -50,6 +60,9 @@ pub struct PasskeyAuthenticator {
     /// See [CollectedClientData](https://www.w3.org/TR/webauthn-2/#dictdef-collectedclientdata)
     /// for more information on this field.
     pub(crate) client_data_json: String,
+    /// Initialization of bytes for passkey in serialized form.
+    #[cfg(feature = "serde")]
+    bytes: OnceLock<Vec<u8>>,
```
DaughterOfMars, line 65: is this actually necessary

thibault-martinez: For now, yes. We need this type to impl AsRef. Removing it is outside of the scope of this PR.

DaughterOfMars: But why do we need AsRef?

thibault-martinez: Because `GenericSignature` in the monorepo is AsRef. Eventually it will be replaced by the SDK enum and we can consider removing it, but for now I'm just replacing the variants. I can open an issue for us to see if we can actually remove AsRef at some point if you want.

DaughterOfMars: Please do 👍

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/issues/1219

DaughterOfMars (approved):

Dkwcs (approved):
LGTM

PR #1178, opened 2026-05-28
feat(graphql)!: remove transaction signer filters

thibault-martinez:
## Summary

https://github.com/iotaledger/iota/issues/9255

The GraphQL schema dropped the deprecated `signAddress` transaction filter (and `SIGN` relationship enum) in favor of `sentAddress` / `SENT`. This refreshes the bundled schema and renames the corresponding Rust field on `TransactionsFilter` from `sign_address` to `sent_address`, plus the same rename in the FFI mirror and the `polling-indexer` example.

Breaking change: callers using `TransactionsFilter::sign_address` must rename to `sent_address`. Semantics are unchanged — both always filtered by sender only.

The schema sync also brings in `startAfter` parameters on the `transactions` / `events` subscriptions, but those are not surfaced through any Rust API in this PR.

## Test plan
- [x] `make ci` passes

### DIFF
--- a/crates/iota-sdk-graphql-client-build/schema.graphql
+++ b/crates/iota-sdk-graphql-client-build/schema.graphql
@@ -149,18 +149,6 @@ The possible relationship types for a transaction block: sent or received.
 """
 enum AddressTransactionBlockRelationship {
   """
-  Transactions this address has sent. NOTE: this input filter has been
-  deprecated in favor of `SENT` which behaves identically but is named
-  more clearly. Both filters restrict transactions by their sender,
-  only, not signers in general.
-
-  This filter will be removed after 6 months with the 1.24.0 release.
-  """
-  SIGN
-    @deprecated(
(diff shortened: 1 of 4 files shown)


Thoralf-M (approved):

crates/iota-sdk-graphql-client-build/schema.graphql
```diff
@@ -4796,16 +4784,46 @@ type Subscription {
...
+
+  On reconnect, clients should pass the digest of the last fully
+  received transaction as `start_after`. The stream resumes from the
+  next transaction, so no partial transaction is ever observed.
   """
-  events(filter: SubscriptionEventFilter): EventSubscriptionPayload!
+  events(
+    startAfter: String
+    filter: SubscriptionEventFilter
+  ): EventSubscriptionPayload!
```
thibault-martinez, line 4826: This will be addresses in another PR

kodemartin (approved):

PR #1184, opened 2026-06-01
feat!: separate Event and GraphQlEvent types

### LINKED ISSUE #1159: FFI `Event` type: lossy/misleading conversion between GraphQL event and core `iota_sdk::types::Event`
## Summary

The uniffi-exported `Event` struct in `crates/iota-sdk-ffi/src/types/events.rs` conflates two distinct concepts:

- A **GraphQL-sourced event** (has `data`, `json`, `timestamp` — optional sender/sending module).
- A **core chain event** (`iota_sdk::types::Event` — all fields required, no `data`/`json`/`timestamp`).

The FFI struct's fields are now `Option<…>` (matching the GraphQL query type), but it still implements `From<Event> for iota_sdk::types::Event` with silent fallback to `ZERO`/empty defaults, and `event_to_bcs` / `event_from_bcs` round-trip uses that lossy path.

## Where it shows up

`crates/iota-sdk-ffi/src/types/events.rs`:

```rust
pub struct Event {
    pub package_id: Option<Arc<ObjectId>>,
    pub module: Option<String>,
    pub sender: Option<Arc<Address>>,
    pub type_: String,
    pub contents: Vec<u8>,
    pub timestamp: Option<String>,
    pub data: String,
    pub json: String,
}

impl From<Event> for iota_sdk::types::Event {
    fn from(value: Event) -> Self {
        Self {
            package_id: value.package_id.map(|id| **id)
                .unwrap_or(iota_sdk::types::ObjectId::ZERO),
            module: Identifier::from_str(value.module.as_deref().unwrap_or("")).unwrap(),
            sender: value.sender.map(|s| **s)
                .unwrap_or(iota_sdk::types::Address::ZERO),
            type_: StructTag::from_str(&value.type_).unwrap(),
            contents: value.contents,
        }
    }
}
```

And in `macros.rs`:

```rust
pub fn event_to_bcs(data: Event) -> Result<Vec<u8>> {
    let data: iota_sdk::types::Event = data.into();
    Ok(bcs::to_bytes(&data)?)
}
```

## Why it matters

The optional fields were added because some on-chain events (e.g. `0x3::validator::StakingRequestEvent`) are emitted by the system and have `sender = null` / `sending_module = null` in the GraphQL response. The previous code did `.unwrap()` on those Options and panicked when the bindings were used to query such events on localnet (`called Option::unwrap() on a None value`). See e.g. `bindings/kotlin/examples/PackageEvents.kt` running against a fresh localnet querying `0x3::validator::StakingRequestEvent`.

The Option-ification fixed the panic for read paths, but introduced two new footguns:

1. `Identifier::from_str(\"\")` is invalid, so the `.unwrap()` in `From<Event> for iota_sdk::types::Event` will panic if `module` is `None`.
2. `event_to_bcs` of a GraphQL-sourced event with a missing sender or module silently produces a BCS-encoded core event with `Address::ZERO` / `ObjectId::ZERO`. That's not equivalent to the original chain event and a downstream consumer can't tell.

The deeper issue is that one FFI struct is doing two jobs:
- mirror of `graphql_client::query_types::Event` (which legitimately has `Option<sender>`, `Option<sending_module>`, plus `data`, `json`, `timestamp`); and
- proxy for `iota_sdk::types::Event` (the BCS-serializable chain type with required `package_id`, `module`, `sender`, `type_`, `contents`).

## Suggested fix

Either:

**(a) Split the types.** Keep the GraphQL-flavored type for `client.events()` results (rename to e.g. `GraphQlEvent` or `EventQuery`) and have a separate FFI `Event` that mirrors `iota_sdk::types::Event` 1:1 with required fields for BCS/JSON round-trip. This matches how the rest of the FFI surface treats these two concerns separately.

**(b) Minimum fix.** At least change `From<Event> for iota_sdk::types::Event` to return a `Result` (so `event_to_bcs` fails cleanly on a GraphQL-sourced event with missing required fields), and drop the `.unwrap()` on `Identifier::from_str(\"\")`.

(a) is the cleaner end state since the FFI `Event` already carries GraphQL-only fields (`data`, `json`, `timestamp`) that have no place on the chain type.

## Repro

Run any binding's `PackageEvents` example against a fresh localnet querying `0x3::validator::StakingRequestEvent`:
- Before the Option-ification: panics with `called Option::unwrap() on a None value`.
- After: works for the read path, but calling `event_to_bcs` on the returned event produces a BCS-encoded core event with zeroed `sender`/`package_id`/empty `module`, which is silently incorrect — and `module = None` triggers a fresh `.unwrap()` panic on `Identifier::from_str(\"\")`.

Thoralf-M:
Fixes #1159

## Summary

The FFI `Event` struct conflated two things: the GraphQL `events()` result (optional sender/module, plus `data`/`json`/`timestamp`) and the chain `iota_sdk::types::Event` (all fields required, BCS-serializable). It `.unwrap()`'d the optional GraphQL fields and panicked on system/genesis events (e.g. `0x3::validator::StakingRequestEvent`, whose sender/module are `null`).

This splits the two concerns:

- **`Event`** (`types::events`) — the chain type, all fields required.
- **`GraphQlEvent`** (`graphql::query_types`, alongside `EventFilter` and the other query-result types) — the GraphQL view, with `package_id`/`module`/`sender`/`timestamp` optional and `type_`/`contents`/`data`/`json` always present.

Both conversions into the chain type are now `TryFrom`, so there are no panics and no silent `0x0`/empty defaults: `event_to_bcs`/`event_to_json` and a malformed GraphQL `bcs` now fail cleanly instead of producing an incorrect/empty result.

## Breaking changes

Generated binding surface (the FFI crate is unpublished; bindings are regenerated):

- `Event` drops the GraphQL-only `timestamp`/`data`/`json` fields.
- `client.events()` now returns `GraphQlEvent`.
- `TransactionEvents::new()` and `GenesisTransaction::new()` are now fallible (`Result`).

## Test plan

- clippy / build / fmt green.
- Go, Python, Swift, C#, and Kotlin `package_events` examples updated to handle the optional `sender`/`module` and verified to build.

https://claude.ai/code/session_01PVTttCUHhsgKhFhyDzR8kg

### DIFF
--- a/bindings/python/examples/package_events.py
+++ b/bindings/python/examples/package_events.py
@@ -18,9 +18,13 @@ async def main():
     )
 
     for event in events.data:
+        # Sender and module are optional: system events (e.g.
+        # 0x3::validator::StakingRequestEvent) have neither.
+        sender = event.sender.to_hex() if event.sender is not None else "none"
+        module = event.module if event.module is not None else "none"
         print(f"Type: {event.type}")
-        print(f"Sender: {event.sender.to_hex()}")
-        print(f"Module: {event.module}")
+        print(f"Sender: {sender}")
+        print(f"Module: {module}")
         print(f"JSON: {event.json}")
 
 
--- a/bindings/python/examples/package_events.py
+++ b/bindings/python/examples/package_events.py
@@ -18,8 +18,8 @@ async def main():
     )
 
     for event in events.data:
-        # Sender and module are optional: system events (e.g.
-        # 0x3::validator::StakingRequestEvent) have neither.
+        # Sender and module are optional: events emitted by the system or at
+        # genesis (sender 0x0, e.g. the genesis StakingRequestEvents) have neither.
         sender = event.sender.to_hex() if event.sender is not None else "none"
         module = event.module if event.module is not None else "none"
         print(f"Type: {event.type}")
--- a/bindings/python/examples/package_events.py
+++ b/bindings/python/examples/package_events.py
@@ -18,8 +18,8 @@ async def main():
     )
 
     for event in events.data:
-        # Sender and module are optional: events emitted by the system or at
-        # genesis (sender 0x0, e.g. the genesis StakingRequestEvents) have neither.
+        # Sender and module are optional: some events (such as system- or
+        # genesis-emitted ones) carry neither.
         sender = event.sender.to_hex() if event.sender is not None else "none"
         module = event.module if event.module is not None else "none"
         print(f"Type: {event.type}")
--- a/crates/iota-sdk-ffi/src/graphql/query_types.rs
+++ b/crates/iota-sdk-ffi/src/graphql/query_types.rs
@@ -432,6 +432,73 @@ impl From<EventFilter> for iota_sdk::graphql_client::query_types::EventFilter {
     }
 }
 
+/// An event as returned by the GraphQL `events` query.
+///
+/// This is a faithful view of the GraphQL response, distinct from the chain
+/// [`Event`](crate::types::events::Event). Three fields are optional, for two
+/// different reasons:
+///
+/// - `package_id`, `module` and `sender` are absent for events emitted by the
+///   system or at genesis (e.g. the genesis validator
+///   `0x3::validator::StakingRequestEvent`s): on chain their sender is the zero
+///   address and their emitting module can't be resolved, both of which the
+///   GraphQL server reports as `null`.
+/// - `timestamp` is absent for events not yet included in a checkpoint (e.g.
+///   from a dry run or a just-executed transaction).
+///
+/// `type_`, `contents`, `data` and `json` are always present (non-null in the
+/// GraphQL schema). Unlike the chain `Event`, this type is not
+/// BCS/JSON-serializable as a chain event.
+#[derive(uniffi::Record)]
+pub struct GraphQlEvent {
+    /// Package id of the top-level function invoked by a MoveCall command which
+    /// triggered this event to be emitted. `None` for system events.
+    pub package_id: Option<Arc<ObjectId>>,
+    /// Module name of the top-level function invoked by a MoveCall command
+    /// which triggered this event to be emitted. `None` for system events.
+    pub module: Option<String>,
+    /// Address of the account that sent the transaction where this event was
+    /// emitted. `None` for system events.
+    pub sender: Option<Arc<Address>>,
+    /// The type of the event emitted
+    pub type_: String,
+    /// BCS serialized bytes of the event
+    pub contents: Vec<u8>,
+    /// UTC timestamp in milliseconds since epoch (1/1/1970)
+    pub timestamp: Option<String>,
+    /// Structured contents of a Move value
+    pub data: String,
+    /// Representation of a Move value in JSON
+    pub json: String,
+}
+
+impl TryFrom<iota_sdk::graphql_client::query_types::Event> for GraphQlEvent {
+    type Error = crate::error::SdkFfiError;
+
+    fn try_from(value: iota_sdk::graphql_client::query_types::Event) -> crate::error::Result<Self> {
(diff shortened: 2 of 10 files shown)


bindings/python/examples/package_events.py
```diff
@@ -18,9 +18,13 @@ async def main():
...
 
     for event in events.data:
+        # Sender and module are optional: some events (such as system- or
+        # genesis-emitted ones) carry neither.
+        sender = event.sender.to_hex() if event.sender is not None else "none"
+        module = event.module if event.module is not None else "none"
         print(f"Type: {event.type}")
-        print(f"Sender: {event.sender.to_hex()}")
-        print(f"Module: {event.module}")
+        print(f"Sender: {sender}")
```
thibault-martinez, line 26: Would this still be hex?

thibault-martinez: I'm blind

crates/iota-sdk-ffi/src/types/events.rs
```diff
@@ -107,9 +65,62 @@ impl From<iota_sdk::types::Event> for Event {
...
+///
+/// Unlike the core chain [`Event`], events emitted by the system or at genesis
+/// have a zero (`0x0`) sender and an unresolvable emitting module — the GraphQL
+/// server returns `null` for both (e.g. the genesis validator
+/// `0x3::validator::StakingRequestEvent`s) — so `package_id`, `module`,
+/// `sender` and `timestamp` are optional. This type is a faithful, non-lossy
+/// view of the GraphQL response and is not BCS/JSON-serializable as a chain
+/// event.
+#[derive(uniffi::Record)]
+pub struct GraphQlEvent {
```
thibault-martinez: I know that's an event type but all other graphql types are in the graphql module, I'd rather have it there

Thoralf-M: Moved in https://github.com/iotaledger/iota-rust-sdk/pull/1184/changes/06cc32c44665d4abc7eb7a1fc2be9db782232d9f

crates/iota-sdk-ffi/src/graphql/query_types.rs
```diff
@@ -432,6 +432,73 @@ impl From<EventFilter> for iota_sdk::graphql_client::query_types::EventFilter {
...
+                Some(sending_module.name),
+            ),
+            None => (None, None),
+        };
+        Ok(Self {
+            package_id,
+            module,
+            sender: value.sender.map(|s| Arc::new(Address(s.address))),
+            type_: value.type_.repr,
+            contents: base64ct::Base64::decode_vec(&value.bcs.0)
```
thibault-martinez, line 493: Is this correct?

thibault-martinez: Ah yeah it is, all good

crates/iota-sdk-ffi/src/types/events.rs
```diff
@@ -37,65 +40,19 @@ pub struct Event {
...
+    type Error = crate::error::SdkFfiError;
 
-impl From<Event> for iota_sdk::types::Event {
-    fn from(value: Event) -> Self {
-        Self {
+    fn try_from(value: Event) -> Result<Self> {
+        Ok(Self {
             package_id: (**value.package_id),
-            module: Identifier::from_str(&value.module).unwrap(),
+            module: Identifier::from_str(&value.module)?,
```
thibault-martinez, line 51: Wondering if these are really fallible

thibault-martinez (approved):

crates/iota-sdk-ffi/src/types/events.rs
Thoralf-M: Yes, `Event` here is the ffi type, anyone could construct it with an invalid string

DaughterOfMars (approved):

PR #1186, opened 2026-06-02
feat(iota-sdk-types): update `Object`

Alex6323:
## Summary

Derive `Hash` for the `Object` type so it can be used as a key in hash-based collections (`HashMap`, `HashSet`) and nested inside other hashable types. This is a non-breaking, additive change — `Object` already derives `Eq`/`PartialEq`, and all of its fields are themselves `Hash`.

## Test plan

- `make clippy`
- `make test`


Thoralf-M (approved):

thibault-martinez (approved):

DaughterOfMars (approved):

crates/iota-sdk-types/src/object.rs
```diff
@@ -578,6 +578,70 @@ impl Object {
...
+    /// Returns true if the object is a system package.
+    pub fn is_system_package(&self) -> bool {
+        self.is_package() && self.object_id().is_system_package()
+    }
+
+    /// Returns the object's owner and id together, if it is owned by a single
+    /// owner.
+    pub fn get_owner_and_id(&self) -> Option<(Owner, ObjectId)> {
+        Some((self.owner, self.object_id()))
+    }
```
DaughterOfMars: There's no check here?

crates/iota-sdk-types/src/object.rs
```diff
@@ -578,6 +578,70 @@ impl Object {
...
+
+    /// Returns the address of the single owner of this object (address- or
+    /// object-owned), or `None` if it is shared or immutable.
+    pub fn get_single_owner(&self) -> Option<Address> {
+        self.owner.address_or_object().copied()
+    }
+
+    /// Changes the owner of this object to `new_owner`.
+    pub fn transfer(&mut self, new_owner: Address) {
```
DaughterOfMars: Can we just call this `set_owner`?

Alex6323: I guess we can but in the monorepo this derefs to this API currently unless we create a wrapper method. Saying that because then we need to get approval again from the ones affected.

crates/iota-sdk-types/src/object.rs
Alex6323: no there wasn't

Alex6323: maybe validity is checked on construction, will check

crates/iota-sdk-types/src/object.rs
```diff
@@ -578,6 +578,70 @@ impl Object {
...
+    }
+
+    /// Returns true if the object is a system package.
+    pub fn is_system_package(&self) -> bool {
+        self.is_package() && self.object_id().is_system_package()
+    }
+
+    /// Returns the object's owner and id together, if it is owned by a single
+    /// owner.
+    pub fn get_owner_and_id(&self) -> Option<(Owner, ObjectId)> {
```
thibault-martinez: Let's drop the `get_` prefix

crates/iota-sdk-types/src/object.rs
```diff
@@ -578,6 +578,70 @@ impl Object {
...
+
+    /// Returns the coin's type parameter if this object is a coin.
+    pub fn coin_type_opt(&self) -> Option<&TypeTag> {
+        self.as_struct_opt()
+            .and_then(|move_object| move_object.struct_tag().coin_type_opt())
+    }
+
+    /// Returns the address of the single owner of this object (address- or
+    /// object-owned), or `None` if it is shared or immutable.
+    pub fn get_single_owner(&self) -> Option<Address> {
```
thibault-martinez: same

crates/iota-sdk-types/src/object.rs
Alex6323: can we just kill this method? It's not even used in the monorepo

DaughterOfMars: 🔪

DaughterOfMars (approved):

PR #1189, opened 2026-06-03
fix(graphql-client): disable idle connection pool to avoid LB 429s

### LINKED ISSUE #1181: flaky doctests: SDK reqwest keepalive pool reuses dropped load-balancer connections
## Symptom

The doctests in `iota-sdk-transaction-builder` flake intermittently in CI with:

```
error decoding response body
caused by: expected value at line 1 column 1
```

Recent CI failures:

- https://github.com/iotaledger/iota-rust-sdk/actions/runs/26551023836
- https://github.com/iotaledger/iota-rust-sdk/actions/runs/26560822340

The tests that have surfaced the failure are:

- `crates/iota-sdk-transaction-builder/src/builder/mod.rs::unstake`
- the crate-level example at `crates/iota-sdk-transaction-builder/src/lib.rs:24`

## Root cause (verified by experiment)

The SDK's GraphQL client (`crates/iota-sdk-graphql-client/src/client.rs:53`)
builds its reqwest client with default settings, which enables the HTTP/2 idle
connection pool and keep-alive. When running a burst of sequential GraphQL
requests through one `Client`, reqwest reuses the same TCP/TLS connection to
the devnet GraphQL load balancer (`lb-0.h.devnet.iota.cafe`). After roughly
~9 sequential requests the LB drops the connection silently; reqwest then
dispatches the next request on the dead socket, the response body comes back
empty, and `.json()` decode fails with `expected value at line 1 column 1` —
exactly the error in CI.

## Repro

A minimal reproducer was added at
`crates/iota-sdk/examples/devnet_pool_repro.rs`. It loops 20 sequential
builder transactions through one `Client::new_devnet()` (the same pattern as
the transaction-builder doctests) and counts where it fails:

```rust
#[tokio::main(flavor = "current_thread")]
async fn main() -> eyre::Result<()> {
    let client = Client::new_devnet();
    // ... fixed sender / coin / recipient addresses ...
    const N: usize = 20;
    for i in 1..=N {
        let mut b = TransactionBuilder::new(sender).with_client(&client);
        b.send_coins([coin], to, 100u64);
        match b.finish().await {
            Ok(_)  => println!("[{i:02}/{N}] ok"),
            Err(e) => { println!("[{i:02}/{N}] FAILED: {e:?}"); std::process::exit(1); }
        }
    }
    Ok(())
}
```

`cargo run --example devnet_pool_repro` against the unmodified SDK:

```
endpoint: https://graphql.devnet.iota.cafe
[01/20] ok
[02/20] ok
[03/20] ok
[04/20] ok
[05/20] ok
[06/20] ok
[07/20] ok
[08/20] ok
[09/20] ok
[10/20] ok
[11/20] FAILED: Client(Error { inner: InnerError { kind: Other, query_errors: None,
    source: Some(reqwest::Error { kind: Decode,
        source: Error("expected value", line: 1, column: 1) }) } })
```

i.e. 10 sequential requests succeed on the pooled connection, the 11th hits a
dead socket and the body decode fails — identical to the CI error.

## Fix (verified)

Disable the idle connection pool for the SDK's GraphQL client so each request
opens a fresh connection:

```rust
// crates/iota-sdk-graphql-client/src/client.rs
inner: reqwest::Client::builder()
    .user_agent(USER_AGENT)
    .pool_max_idle_per_host(0)
    .build()?,
```

After applying this patch, `cargo run --example devnet_pool_repro` succeeds
20/20:

```
[01/20] ok
[02/20] ok
... (every iteration ok) ...
[20/20] ok
```

The trade-off is one fresh TCP+TLS handshake per request (~50–100 ms on
devnet). For an SDK GraphQL client that fans out short, sequential queries
this is acceptable and far preferable to the silent decode failure.

## Related

An in-flight PR migrates the transaction-builder doctests from testnet to
devnet to sidestep testnet overload. That migration does **not** fix this
issue — the same keepalive bug reproduces on devnet, as shown above. This SDK
fix is the actual root cause for the doctest flakes regardless of which
network they target.

## Out of scope

- The doctest network migration (testnet → devnet) is tracked in a separate PR.

Thoralf-M:
## Why

The devnet/testnet GraphQL load balancers apply a per-connection rate limit and return HTTP 429 once a burst of requests reuses the same pooled keep-alive connection. This is the root cause of the flaky transaction-builder doctests in CI (Fixes https://github.com/iotaledger/iota-rust-sdk/issues/1181).

## What

Disable reqwest's idle connection pool (`pool_max_idle_per_host(0)`) so every request opens a fresh connection. No connection ever accumulates enough requests to trip the per-connection cap, so the 429 is *prevented* rather than reacted to — this is the fix verified in the issue. Gated to non-wasm32, since reqwest's wasm backend has no pool controls (the browser manages connections).

The trade-off is one TCP+TLS handshake per request (~50–100ms), acceptable for the SDK's short, sequential GraphQL queries.

> An earlier revision tried keeping the pool and retrying once on a fresh connection. CI showed that's insufficient: the first attempt still rides the pooled connection that saturates the cap, and reactive retries under parallel doctests can themselves catch a 429. Disabling the pool prevents the 429 entirely and is simpler.

## Test plan

- `cargo clippy -- -Dwarnings`, nightly fmt, native + `wasm32-unknown-unknown` builds, offline unit tests pass locally.
- [run-ci] to exercise the previously-flaky doctests against the live network.

https://claude.ai/code/session_01Rg44DHJmk4sRcrFNCsgNrv

### DIFF
--- a/crates/iota-sdk-graphql-client/src/client.rs
+++ b/crates/iota-sdk-graphql-client/src/client.rs
@@ -48,9 +48,19 @@ impl Client {
     pub fn new(server: &str) -> Result<Self> {
         let rpc = reqwest::Url::parse(server)?;
 
+        let builder = reqwest::Client::builder().user_agent(USER_AGENT);
+        // The GraphQL load balancers apply a per-connection rate limit and
+        // return HTTP 429 once a burst of requests reuses the same pooled
+        // keep-alive connection. Disabling the idle connection pool makes every
+        // request use a fresh connection, so no connection accumulates enough
+        // requests to trip the limit. `pool_max_idle_per_host` is not available
+        // on the wasm32 reqwest backend, where the browser manages connections.
+        #[cfg(not(target_arch = "wasm32"))]
+        let builder = builder.pool_max_idle_per_host(0);
+
         let client = Client {
             rpc,
-            inner: reqwest::Client::builder().user_agent(USER_AGENT).build()?,
+            inner: builder.build()?,
             service_config: Default::default(),
         };
         Ok(client)


Thoralf-M:
Server side was just too limited, not needed anymore

PR #1190, opened 2026-06-03
feat: implement ClientMethods for the gRPC client

### LINKED ISSUE #1139: Implement `ClientMethods` for the GrpcClient
Can you impl `ClientMethods` for your grpc client? Otherwise we won't be able to use it with the TransactionBuilder.

_Originally posted by @thibault-martinez in https://github.com/iotaledger/iota-rust-sdk/pull/1062#pullrequestreview-4263286036_

DaughterOfMars:
Implements `ClientMethods` for the gRPC `Client` so it can drive the `TransactionBuilder`, just like the GraphQL client already does.

Closes #1139.

### What
- New `ClientMethods` impl for the gRPC client, wiring each method to the existing high-level gRPC APIs (`get_objects`, `get_transactions`, `get_epoch`, `simulate_transaction`, `execute_transaction`, `list_owned_objects`). Missing objects/transactions map to `Ok(None)`; `wait_for_tx` polls until the tx is indexed (`IndexedOnNode`) or in a checkpoint (`Finalized`).
- Simplifies the `ClientMethods::objects` signature to take `Option<struct_tag: StructTag>` + `owner: Address` directly, dropping the unused `object_ids`, optional `owner`, and `ascending` params — the builder only ever lists an owner's coins by type. Updated the GraphQL and FFI impls accordingly.
- Adds `tokio` (`time` feature) to `iota-sdk-grpc-client` for the `wait_for_tx` poll loop (tonic already pulls tokio in; the crate isn't built for WASM).

### Test plan
- `cargo clippy` clean across grpc-client, transaction-builder, graphql-client, and ffi
- `cargo nextest run` for transaction-builder + grpc-client
- `cargo check --target wasm32-unknown-unknown` for the touched WASM crates


crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -80,10 +80,8 @@ pub trait ClientMethods {
     /// pass the cursor returned by a previous call to advance.
     fn objects(
         &self,
-        type_tag: Option<TypeTag>,
-        owner: Option<Address>,
-        object_ids: Option<Vec<ObjectId>>,
-        ascending: bool,
+        struct_tag: StructTag,
```
Thoralf-M: Why make it required? If we want to add something like `send_all_objects()` it wouldn't really work this way, because we would need to know all the StructTags the user owns

crates/iota-sdk-grpc-client/src/client_methods.rs
```diff
@@ -0,0 +1,253 @@
...
+                    Ok(envelope) => {
+                        if let Some(tx) = envelope.into_inner().first() {
+                            let ready = match wait_for {
+                                WaitForTx::IndexedOnNode => true,
+                                WaitForTx::Finalized => tx.checkpoint_sequence_number().is_ok(),
+                                _ => unreachable!("checked above"),
+                            };
+                            if ready {
+                                return Ok(());
+                            }
```
Thoralf-M, line 242: Feel free to ignore, but could change it to a single match like this
```Rust
// Single source of truth: each variant maps to (field to request, readiness predicate).
let (mask, is_ready): (ReadMask, fn(&ExecutedTransaction) -> bool) = match wait_for {
    WaitForTx::IndexedOnNode => (ReadMask::from(TRANSACTION_DIGEST), |_| true),
    WaitForTx::Finalized => (
        ReadMask::from(EXECUTED_TRANSACTION_CHECKPOINT),
        |tx| tx.checkpoint_sequence_number().is_ok(),
    ),
    _ => unimplemented!("a new WaitForTx enum variant was added and needs to be handled"),
};

let poll = async {
    let mut interval = tokio::time::interval(WAIT_FOR_TX_POLL_INTERVAL);
    loop {
        interval.tick().await;
        match self.get_transactions(&[digest], Some(mask.clone())).await {
            Ok(envelope) => {
                if let Some(tx) = envelope.into_inner().first() {
                    if is_ready(tx) {
                        return Ok(());
                    }
                }
            }
            Err(e) if is_not_found(&e) => {}
            Err(e) => return Err(e),
        }
    }
};

```

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
DaughterOfMars: Fair point. I can make it optional again. The owner, however, must be provided to gprc

crates/iota-sdk-grpc-client/src/client_methods.rs
DaughterOfMars: feels a bit messier tbh

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
Thoralf-M: Yes owner is also fine since we can't build txs that touch objects from random owners

Thoralf-M: Please update the PR description for this change

Thoralf-M (approved):

crates/iota-sdk-transaction-builder/src/builder/client_methods.rs
```diff
@@ -391,10 +384,8 @@ pub(crate) mod test_client {
 
         async fn objects(
             &self,
-            _type_tag: Option<TypeTag>,
-            owner: Option<Address>,
-            _object_ids: Option<Vec<ObjectId>>,
-            _ascending: bool,
```
thibault-martinez, line 397: The description doesn't really explain the why of the removed fields, why don't we ever make use of them?

DaughterOfMars: The transaction builder simply doesn't need them and in some clients they are unusable

crates/iota-sdk-grpc-client/src/client_methods.rs
```diff
@@ -0,0 +1,253 @@
...
+            Err(e) => Err(e),
+        }
+    }
+
+    async fn reference_gas_price(
+        &self,
+        epoch: impl Into<Option<u64>>,
+    ) -> Result<Option<u64>, Self::Error> {
+        let epoch = self
+            .get_epoch(epoch.into(), Some(ReadMask::from("reference_gas_price")))
```
thibault-martinez: Use the constant?

crates/iota-sdk-grpc-client/src/client_methods.rs
```diff
@@ -0,0 +1,253 @@
...
+            .iter()
+            .map(|obj| obj.object().map_err(Error::from))
+            .collect::<Result<Vec<_>, _>>()?;
+        let next_cursor = page.next_page_token.map(|token| token.to_vec());
+        Ok(ObjectsPage { data, next_cursor })
+    }
+
+    async fn protocol_config(&self) -> Result<ProtocolConfig, Self::Error> {
+        let epoch = self
+            .get_epoch(None, Some(ReadMask::from("protocol_config.attributes")))
```
thibault-martinez: Use the constant?

crates/iota-sdk-grpc-client/src/client_methods.rs
```diff
@@ -0,0 +1,253 @@
...
+        let epoch = self
+            .get_epoch(epoch.into(), Some(ReadMask::from("reference_gas_price")))
+            .await?
+            .into_inner();
+        Ok(epoch.reference_gas_price)
+    }
+
+    async fn estimate_tx_budget(&self, tx: &Transaction) -> Result<Option<u64>, Self::Error> {
+        // Simulate with relaxed checks and read the gas used from the resulting
+        // effects, mirroring the GraphQL client's estimate.
```
thibault-martinez: Not sure it's worth mentioning graphql here

crates/iota-sdk-grpc-client/src/client_methods.rs
```diff
@@ -0,0 +1,253 @@
...
+
+        Ok(effects)
+    }
+
+    async fn wait_for_tx(&self, digest: Digest, wait_for: WaitForTx) -> Result<(), Self::Error> {
+        // Request only the field that proves the desired condition: any
+        // transaction field confirms it is indexed on the node, while
+        // `checkpoint` is only populated once it has been finalized.
+        let mask = match wait_for {
+            WaitForTx::IndexedOnNode => ReadMask::from(TRANSACTION_DIGEST),
```
thibault-martinez, line 221: Are we 100% sure this is enough?

crates/iota-sdk-grpc-client/src/client_methods.rs
```diff
@@ -0,0 +1,253 @@
...
+            WaitForTx::Finalized => ReadMask::from(EXECUTED_TRANSACTION_CHECKPOINT),
+            _ => {
+                unimplemented!("a new WaitForTx enum variant was added and needs to be handled")
+            }
+        };
+
+        let poll = async {
+            let mut interval = tokio::time::interval(WAIT_FOR_TX_POLL_INTERVAL);
+            loop {
+                interval.tick().await;
```
thibault-martinez, line 231: Does this mean waiting a first time already at the beginning?

DaughterOfMars: No, this is the same way the graphql one works

crates/iota-sdk-grpc-client/src/client_methods.rs
DaughterOfMars: I can check with max

thibault-martinez: He's off, ask someone else from node 🌴

Alex6323 (approved):

crates/iota-sdk-grpc-client/src/client_methods.rs
DaughterOfMars: According to Bing it seems this is correct

thibault-martinez (approved):

PR #1191, opened 2026-06-04
refactor(types)!: align readable serialization with base types

### LINKED ISSUE #440: Can we get rid of the serialization modules?

DaughterOfMars:
## Why

The JSON / human-readable serialization format is no longer owned by the SDK. Readable forms should simply mirror the base type instead of carrying SDK-specific variant/field renames and a denormalized `Object` shape. The manual `Readable`/`Binary` split only needs to exist where BCS requires an encoding a plain derive can't express.

## What

- **`Owner`, `Object`** — drop the manual `Readable*` proxies and derive directly (BCS is byte-identical, so no custom encoding is needed). `Object`'s readable form is no longer denormalized — no more hoisted `object_id`/`version`/`type` or flattened data.
- **`GenesisObject`** — keeps the manual split (BCS needs the `RawObject` `%d00` wrapper) but its readable proxy now mirrors the base `{data, owner}`.
- **`TransactionEffectsV1`** — drop the `gas_cost_summary` → `gas_used` rename.
- **`SharedObjectReference`** — drop `rename_all = "camelCase"`.
- **`contents` / `modules`** — kept as base64 in JSON via `IfIsHumanReadable<Base64Encoded, Bytes>`.

**Breaking:** changes JSON output for these types. BCS wire format is unchanged.

## Test plan

- `bcs-schema.abnf` regenerates with no diff; `object_fixture` round-trips byte-for-byte against mainnet fixtures.
- Full `iota-sdk-types` suite (280 tests) passes; clippy clean; WASM + workspace build.


crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -339,7 +338,6 @@ mod serialization {
         status: ExecutionStatus,
         #[serde(with = "crate::_serde::ReadableDisplay")]
```
thibault-martinez: Are these things necessary still?

crates/iota-sdk-types/src/object.rs
```diff
@@ -469,6 +466,7 @@ pub struct Object {
     /// The amount of IOTA we would rebate if this object gets deleted.
     /// This number is re-calculated each time the object is mutated based on
     /// the present storage gas price.
+    #[cfg_attr(feature = "serde", serde(with = "crate::_serde::ReadableDisplay"))]
```
Thoralf-M, line 474: `storage_rebate` keeps `ReadableDisplay`, but package `version` doesn't: `Object` → `ObjectData::Package(MovePackage)` → `version: Version` (`#[serde(transparent)]` u64, no `ReadableDisplay`) ⇒ bare JSON **number**, while struct objects keep it a **string** via `ReadableMoveStructRef`. Inconsistent, and JS-unsafe once `version > 2^53` (`MAX_VALID_EXCL = 2^63-1`).

Fix: `#[serde(with = "crate::_serde::ReadableDisplay")]` on `MovePackage.version` (BCS no-op).

Repro:

```rust
#[test]
fn package_version_json_number_vs_struct_string() {
    // package version → bare JSON number; struct version → string
    const UNSAFE: u64 = (1 << 53) + 1;
    let v = Version::from_u64(UNSAFE);

    let pkg = Object::new(
        ObjectData::Package(MovePackage {
            id: ObjectId::ZERO, version: v,
            modules: Default::default(),
            type_origin_table: Vec::new(),
            linkage_table: Default::default(),
        }),
        Owner::Immutable, Digest::ZERO, 0,
    );
    let strct = Object::new(
        ObjectData::Struct(
            MoveStruct::new(
                MoveObjectType::new(StructTag::new(
                    Address::FRAMEWORK,
                    Identifier::new("bar").unwrap(),
                    Identifier::new("foo").unwrap(),
                    Vec::new(),
                )),
                v, ObjectId::ZERO.into(),
            ).unwrap(),
        ),
        Owner::Immutable, Digest::ZERO, 0,
    );

    let pv = &serde_json::to_value(&pkg).unwrap()["data"]["Package"]["version"];
    let sv = &serde_json::to_value(&strct).unwrap()["data"]["Struct"]["version"];
    assert!(pv.is_number());                       // package: number (the bug)
    assert!(sv.is_string());                       // struct: string
    assert!(pv.as_u64().unwrap() > (1 << 53) - 1); // > JS MAX_SAFE_INTEGER
    assert_eq!(sv.as_str().unwrap(), UNSAFE.to_string());
}
```

crates/iota-sdk-types/src/object.rs
```diff
@@ -1214,7 +902,7 @@ mod serialization {
         use wasm_bindgen_test::wasm_bindgen_test as test;
 
         use super::*;
-        use crate::object::Object;
+        use crate::{Identifier, object::Object};
```
Thoralf-M: `obj()` below only `println!`s, never asserts; `object_fixture` only checks round-trips (which pass despite the number-vs-string `version` split) — nothing pins the JSON shape. Add an `assert_snapshot!` over a **package** object to catch it. Also the commented owner alternatives at 923-927 are stale (`Shared { initial_shared_version }` → tuple `Shared(Version)`).

crates/iota-sdk-types/src/effects/v1.rs
thibault-martinez: You answered in the huddle

crates/iota-sdk-types/src/object.rs
DaughterOfMars: I dunno. I think it's fine to remove these tags

crates/iota-sdk-types/src/object.rs
DaughterOfMars: I didn't write that fn 🤷‍♀️

crates/iota-sdk-types/src/object.rs
Thoralf-M: Same issue as always, if a big number is serialized as number and this JSON will be read in JS it's unsafe there

crates/iota-sdk-types/src/object.rs
Thoralf-M: I know, but I think you should modify it now, because you're doing a change that breaks something and is not obvious if this is not modified

crates/iota-sdk-types/src/object.rs
DaughterOfMars: yeah but that's a json problem and we no longer care specifically about json

Thoralf-M: We still have to/from json methods on the types for the bindings, so I'd say we still care about json, just not about compatibility with the JSON-RPC

thibault-martinez: > we still care about json, just not about compatibility with the JSON-RPC

I agree

thibault-martinez (commented):
What about BinaryObjectIn, BinaryObjectOut, BinaryUnchangedSharedKind, BinaryArgument, and BinaryTransactionKind?

DaughterOfMars:
> BinaryArgument

Sorry, those are fixed now

crates/iota-sdk-types/src/object.rs
DaughterOfMars: Yeah but that's not what I said. What about other formats that can serialize u64 just fine?

crates/iota-sdk-types/src/object.rs
DaughterOfMars: I don't even understand what you want to add. Can you just push it?

crates/iota-sdk-types/src/object.rs
Thoralf-M: Both ways work for them I guess, but we should use what works with all formats if possible

crates/iota-sdk-types/src/object.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1191/commits/6802b68b666c36a91ab254f62cc9c19329a0a59f

crates/iota-sdk-types/src/object.rs
DaughterOfMars: Pretty sure serde_json handles all values of u64 anyway

thibault-martinez (commented):
What about VersionAssignment, CancelledTransaction, ConsensusDeterminedVersionAssignments?
They may diverge in tuple/struct but it doesn't affect BCS

DaughterOfMars:
> What about VersionAssignment, CancelledTransaction, ConsensusDeterminedVersionAssignments? They may diverge in tuple/struct but it doesn't affect BCS

Looks like you're right. Removing them

crates/iota-sdk-types/src/checkpoint.rs
```diff
@@ -116,23 +117,28 @@ pub struct EndOfEpochData {
...
 /// ```
 #[derive(Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
 #[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
 pub struct CheckpointSummary {
     /// Epoch that this checkpoint belongs to.
+    #[cfg_attr(feature = "serde", serde(with = "crate::_serde::ReadableDisplay"))]
     #[cfg_attr(feature = "bcs-schema", bcs_schema(as_type = "u64"))]
     pub epoch: EpochId,
```
thibault-martinez, line 127: Do we want to open an issue to make these proper types and have the attributes on it? Or do you think it's not worth it?

crates/iota-sdk-types/src/checkpoint.rs
```diff
@@ -141,17 +147,25 @@ pub struct CheckpointSummary {
...
     /// CheckpointSummary is not an evolvable structure - it must be readable by
     /// any version of the code. Therefore, in order to allow extensions to
     /// be added to CheckpointSummary, we allow opaque data to be added to
     /// checkpoints which can be deserialized based on the current
     /// protocol version.
+    #[cfg_attr(feature = "serde", serde(default))]
+    #[cfg_attr(
+        feature = "serde",
+        serde(with = "crate::_serde::ReadableBase64Encoded")
+    )]
```
thibault-martinez: merge them?

crates/iota-sdk-types/src/checkpoint.rs
DaughterOfMars: not worth it

thibault-martinez (approved):
Assuming Thoralf's point is addressed

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -953,15 +961,10 @@ impl Input {
...
-#[cfg_attr(
-    feature = "serde",
-    derive(serde::Deserialize, serde::Serialize),
-    serde(rename_all = "camelCase")
-)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
 pub struct SharedObjectReference {
     pub object_id: ObjectId,
-    #[cfg_attr(feature = "serde", serde(with = "crate::_serde::ReadableDisplay"))]
```
Thoralf-M, line 964: Shouldn't this also be reverted?

Thoralf-M: Version itself takes care of the serialization

Thoralf-M (approved):

PR #1192, opened 2026-06-04
feat(types): add SplitSerde derive for the BCS/JSON serialization split

### LINKED ISSUE #440: Can we get rid of the serialization modules?

DaughterOfMars:
## Why

A lot of core types hand-roll `Serialize`/`Deserialize` impls that branch on `is_human_readable()` so JSON can have a different shape than BCS — serde's `#[derive]` can't express "this representation only in JSON / that one only in BCS". It's repetitive boilerplate, and the binary form is easy to accidentally drift from the type (the `Owner` bug).

Follow-up to the readable-serialization cleanup. Plan A from that discussion: a standalone derive, owned-only (no borrowed mirrors yet).

## What

New `iota-sdk-serde-derive` proc-macro crate with `#[derive(SplitSerde)]`. From one definition it generates two private shadow types + the dispatching impls. The **binary shadow is always a plain serde derive of the base type**, so the BCS wire format can't drift; only the readable shadow carries JSON behaviour.

Attributes:
- type/variant/field `json(<serde meta>)` → applied to the readable shadow only (e.g. `json(tag = "kind", rename_all = "snake_case")`, or `json(skip_serializing_if = "...", default)`).
- field `with = "path"` → `#[serde(with = "path")]` on both shadows; the adapter must be `IfIsHumanReadable`-based (e.g. `crate::_serde::ReadableDisplay`, `crate::_serde::ReadableBase64Encoded`) so the binary form is unaffected.

Migrated as proof of concept:
- `CheckpointCommitment`, `PackageUpgradeError` — JSON-tagged enums.
- `CheckpointSummary` — `ReadableDisplay` + `ReadableBase64Encoded` + JSON-only `skip`/`default`.

## Scope / follow-ups

Owned-only (clones on serialize); generics and tuple/unit structs are rejected with a clear error. Borrowed mirrors, deserialize validation hooks, and unifying the binary shadow with the `bcs-schema` derive are left for later. Structurally bespoke encodings (envelopes, transpositions, reserved indices) stay hand-written.

## Test plan

- Serialization is unchanged: `bcs-schema.abnf` regenerates with **no diff**; the `signed_checkpoint_fixture` / `contents_fixture` (pin real JSON+BCS) and all round-trip proptests pass.
- Full `iota-sdk-types` suite (280 tests), clippy (`-Dwarnings`), no-serde build (attr gating), WASM, and workspace build all pass.


scripts/cargo_sort/cargo_sort.py
thibault-martinez, line 1: Bit annoying to review, did anything change or did you just format it?

DaughterOfMars: I just formatted it (accidentally)

DaughterOfMars: Oh sorry, I did add the new crate tho

PR #1193, opened 2026-06-08
fix: correct wasm getrandom configuration for downstream consumers

DaughterOfMars:
## Why

`iota-sdk` declared `getrandom_3` as a real (non-dev) wasm dependency. That made it the **only** thing pulling getrandom 0.3 into a downstream consumer's wasm dependency tree — and getrandom 0.3 requires the `--cfg getrandom_backend="wasm_js"` rustflag, which is a property of the final compiled artifact and **can't** be supplied by a library dependency. So downstream wasm builds hit getrandom's `compile_error!`. The SDK's own builds pass only because its `.cargo/config.toml` sets that flag locally, which masks the bug (and doesn't propagate to consumers).

## What

Mirror the `iota-sdk-crypto` / `iota-sdk-types` crates: scope `getrandom_3` to `[target.'cfg(target_arch = "wasm32")'.dev-dependencies]` (adding `getrandom_2` for consistency). getrandom 0.3 is unused at runtime on wasm — the runtime RNG is getrandom 0.2 (via graphql-client and rand_core 0.6) — so this removes no functionality, only the unsatisfiable runtime dep imposed on consumers.

## Test plan

- `cargo tree -p iota-sdk --target wasm32-unknown-unknown -e no-dev -i getrandom@0.3` → empty after the change (was the sole `getrandom 0.3 → iota-sdk` path before).
- `cargo check -p iota-sdk --target wasm32-unknown-unknown` and `make wasm` (`wasm-pack build`) both pass.

### DIFF
--- a/crates/iota-sdk-crypto/Cargo.toml
+++ b/crates/iota-sdk-crypto/Cargo.toml
@@ -88,9 +88,8 @@ test-strategy.workspace = true
 iota-types = { workspace = true, features = ["proptest"] }
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
-# `getrandom` renamed the WebAssembly support feature from "js" to "wasm_js"
-# Since both versions are in our dependency tree, we need to overwrite them with the proper feature
-getrandom_2.workspace = true
+# getrandom 0.2's "js" feature is inherited from iota-types (a real wasm dep there).
+# getrandom 0.3 needs the `getrandom_backend="wasm_js"` rustflag too (.cargo/config.toml); proptest pulls it into our tests.
 getrandom_3.workspace = true
 wasm-bindgen-test.workspace = true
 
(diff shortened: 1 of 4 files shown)


crates/iota-sdk-crypto/Cargo.toml
```diff
@@ -88,9 +88,8 @@ test-strategy.workspace = true
 iota-types = { workspace = true, features = ["proptest"] }
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
-# `getrandom` renamed the WebAssembly support feature from "js" to "wasm_js"
-# Since both versions are in our dependency tree, we need to overwrite them with the proper feature
-getrandom_2.workspace = true
+# getrandom 0.2's "js" feature is inherited from iota-types (a real wasm dep there).
```
Thoralf-M, line 91: Not sure if there is really value in mentioning this here

Thoralf-M (approved):

thibault-martinez (approved):

crates/iota-sdk-crypto/Cargo.toml
DaughterOfMars: I think it is actually important because if it wasn't inherited then it would break the compilation

PR #1194, opened 2026-06-08
feat!: use GraphQL subscriptions for events_stream and transactions_stream

### LINKED ISSUE #1179: Add `startAfter` to graphql event stream
> This will be addresses in another PR 

 _Originally posted by @thibault-martinez in [#1178](https://github.com/iotaledger/iota-rust-sdk/pull/1178/changes#r3315994988)_

DaughterOfMars:
## Why

[iotaledger/iota#9365](https://github.com/iotaledger/iota/pull/9365) added GraphQL subscription support for live events and transactions. This switches `Client::events_stream` / `transactions_stream` from historical HTTP pagination to live `graphql-transport-ws` WebSocket subscriptions.

## What

- New WebSocket transport (`tokio-tungstenite` + `graphql-ws-client`) with an auto-reconnecting stream that tracks the `startAfter` resume cursor and backs off on disconnect.
- New cynic types for the `Subscription` root, the `@oneOf` `SubscriptionEventFilter` / `SubscriptionTransactionFilter`, and the `Event | Lagged` / `TransactionBlock | Lagged` payload unions.
- The two methods drop `Direction` and take `start_after`; the paginated `events()` / `transactions()` page methods are unchanged.

**Breaking** (acceptable pre-1.0): filter types and stream semantics change.

## Heads-up for reviewers

- **WASM**: the two stream helpers are now non-wasm only (the WebSocket transport isn't built for `wasm32`). WASM consumers keep the page-based methods; WASM subscriptions (via `ws-stream-wasm`) can follow.
- **Server support**: as of 2026-08-04 subscriptions are deployed and enabled on all public networks, not just devnet/localnet. `serviceConfig.enabledFeatures` lists `SUBSCRIPTIONS` on mainnet (GraphQL 1.28.3), testnet (1.28.2-rc) and devnet (1.29.0-beta), and a live `subscription { transactions { ... on TransactionBlock { digest } } }` against `wss://graphql.mainnet.iota.cafe/graphql/subscriptions` streams payloads within seconds. Reviewers can therefore point the stream methods at any public endpoint; the new integration test still runs against localnet, so it needs a subscription-capable localnet.

## Test plan

- [x] `cargo clippy --all-features --all-targets`
- [x] `cargo +nightly fmt --check`
- [x] `make wasm` (graphql-client builds for wasm32 with subscriptions gated out)
- [x] `make test`
- [ ] `make test-with-localnet` — runs `test_transactions_subscription` (needs a subscription-capable localnet)

Closes #1179


crates/integration-tests/tests/graphql_ptbs.rs
```diff
@@ -461,3 +461,36 @@ async fn test_auto_gas_pins_full_first_page_for_consolidation() {
         resolved.gas_payment.objects.len(),
     );
 }
+
+/// Open a live transactions subscription and assert it delivers a transaction.
+/// A transfer is executed on a background task once the subscription has had a
+/// moment to connect, so the live stream observes it (localnet also produces
+/// system transactions continuously). This exercises the WebSocket transport,
```
Thoralf-M: `(localnet also produces system transactions continuously)` shouldn't be mentioned here, that's obvious

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
```diff
@@ -0,0 +1,217 @@
...
+
+/// Filter incoming transactions in a subscription. Exactly one field may be
+/// set (the GraphQL input is `@oneOf`).
+#[derive(Clone, cynic::InputObject, Debug, Default)]
+#[cynic(schema = "rpc", graphql_type = "SubscriptionTransactionFilter")]
+pub struct SubscriptionTransactionFilter {
+    /// Filter incoming transactions by kind.
+    pub kind: Option<TransactionBlockKindInput>,
+    /// Filter incoming transactions by signing address.
+    pub signing_address: Option<Address>,
```
Thoralf-M: Does that work if the signing address is sender or sponsor? Would be good to clarify in the comment

Thoralf-M (commented):
What about FFI? Can they not have streams? If it's just more complicated maybe open an issue

crates/iota-sdk-graphql-client/src/subscription.rs
```diff
@@ -0,0 +1,266 @@
+// Copyright (c) Mysten Labs, Inc.
```
Thoralf-M: It's not from Mysten Labs, is it?

DaughterOfMars: Indeed not

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
```diff
@@ -0,0 +1,221 @@
+// Copyright (c) Mysten Labs, Inc.
+// Modifications Copyright (c) 2026 IOTA Stiftung
```
Thoralf-M: Same here

crates/iota-sdk-graphql-client/src/subscription.rs
```diff
@@ -0,0 +1,265 @@
...
+    }
+}
+
+/// Wrap a connect-and-subscribe closure in an auto-reconnecting stream.
+///
+/// `connect` is called with the current resume cursor on every (re)connection
+/// and must yield a stream of decoded [`Outcome`]s. Connection and transport
+/// errors are surfaced to the consumer and then trigger a backed-off reconnect;
+/// the stream itself never terminates.
+fn reconnecting_subscription<'a, T, C, Fut, S>(
```
Thoralf-M, line 227: There is nothing like this already in tokio_tungstenite?

DaughterOfMars:
> What about FFI? Can they not have streams? If it's just more complicated maybe open an issue

Created https://github.com/iotaledger/iota-rust-sdk/pull/1303 to investigate.

crates/iota-sdk-graphql-client/src/subscription.rs
DaughterOfMars: no

crates/iota-sdk-graphql-client/src/subscription.rs
```diff
@@ -0,0 +1,265 @@
...
+impl Client {
+    /// Subscribe to a live stream of events matching the (optional) filter.
+    ///
+    /// The stream yields events as they arrive and reconnects automatically on
+    /// disconnect. `start_after` optionally resumes the stream from the
+    /// transaction immediately following the given transaction digest;
+    /// thereafter the stream tracks its own resume point.
+    ///
+    /// Note: subscriptions are served over a WebSocket and are currently
+    /// supported on devnet and localnet only.
```
Thoralf-M: Also mentions devnet and localnet only, still right?

crates/integration-tests/tests/graphql_ptbs.rs
```diff
@@ -461,3 +461,36 @@ async fn test_auto_gas_pins_full_first_page_for_consolidation() {
...
     );
 }
+
+/// Open a live transactions subscription and assert it delivers a transaction.
+/// A transfer is executed on a background task once the subscription has had a
+/// moment to connect, so the live stream observes it. This exercises the
+/// WebSocket transport, payload decoding, and resume-cursor tracking shared by
+/// `events_stream`.
+#[tokio::test]
+async fn test_transactions_subscription() {
```
Thoralf-M, line 472: Please modify or add a new test to also cover the events part. A staking transaction can be used as this will emit an event `0x3::validator::StakingRequestEvent`

DaughterOfMars: added a test

crates/iota-sdk-graphql-client/src/subscription.rs
DaughterOfMars: fixed

Thoralf-M (approved):

thibault-martinez (commented):
Is this doable in FFI?

thibault-martinez (commented):
Please add a rust example

DaughterOfMars:
> Is this doable in FFI?

Yes, see the above PR

thibault-martinez:
> This switches Client::events_stream / transactions_stream from historical HTTP pagination to live graphql-transport-ws WebSocket subscriptions.

Are we sure we want to do that? Why not just add it? Bit of a shame to lose WASM support?

DaughterOfMars:
> > This switches Client::events_stream / transactions_stream from historical HTTP pagination to live graphql-transport-ws WebSocket subscriptions.
> 
> Are we sure we want to do that? Why not just add it? Bit of a shame to lose WASM support?

We can still re-add wasm support. Either with a fake stream cfg gated or with actual wasm support which does exist

thibault-martinez:
> > > This switches Client::events_stream / transactions_stream from historical HTTP pagination to live graphql-transport-ws WebSocket subscriptions.
> > 
> > 
> > Are we sure we want to do that? Why not just add it? Bit of a shame to lose WASM support?
> 
> We can still re-add wasm support. Either with a fake stream cfg gated or with actual wasm support which does exist

That does not really answer the question, why isn't it something we are adding? Why replace the HTTP pagination?

DaughterOfMars:
> > > > This switches Client::events_stream / transactions_stream from historical HTTP pagination to live graphql-transport-ws WebSocket subscriptions.
> > > 
> > > 
> > > Are we sure we want to do that? Why not just add it? Bit of a shame to lose WASM support?
> > 
> > 
> > We can still re-add wasm support. Either with a fake stream cfg gated or with actual wasm support which does exist
> 
> That does not really answer the question, why isn't it something we are adding? Why replace the HTTP pagination?

Because it's weird and streaming is better

thibault-martinez:
> Because it's weird and streaming is better

Pagination is weird? How?

Anyway, it's fine if you don't want to do WASM support in this PR but then please open an issue.

crates/iota-sdk/examples/subscriptions_graphql.rs
```diff
@@ -0,0 +1,98 @@
...
+    let client = Client::new_testnet();
+
+    // Subscriptions are served over a WebSocket that the node has to have
+    // enabled
+    let enabled = client
+        .service_config()
+        .await?
+        .enabled_features
+        .iter()
+        .any(|feature| matches!(feature, Feature::Subscriptions));
```
thibault-martinez: Want to use your new `supports_feature`?

DaughterOfMars:
> > Because it's weird and streaming is better
> 
> Pagination is weird? How?
> 
> Anyway, it's fine if you don't want to do WASM support in this PR but then please open an issue.

https://github.com/iotaledger/iota-rust-sdk/issues/1336 opened

crates/iota-sdk/examples/subscriptions_graphql.rs
DaughterOfMars: yes, thanks for the reminder

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
```diff
@@ -0,0 +1,220 @@
...
+#[cynic(schema = "rpc", graphql_type = "SubscriptionEventFilter")]
+pub struct SubscriptionEventFilter {
+    /// Filter incoming events by emitting module, e.g. `"0x02"` (package) or
+    /// `"0x02::coin"` (module).
+    pub emitting_module: Option<String>,
+}
+
+/// Filter incoming transactions in a subscription. Exactly one field may be
+/// set (the GraphQL input is `@oneOf`).
+#[derive(Clone, cynic::InputObject, Debug, Default)]
```
thibault-martinez, line 79: Do we really want to derive Default on the filters? An all None filter is invalid and ends with an error

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
```diff
@@ -0,0 +1,220 @@
...
+            .transpose()?
+            .map(|bcs| bcs::from_bytes::<SenderSignedTransaction>(&bcs))
+            .transpose()?;
+
+        let signatures = if let Some(sigs) = value.signatures {
+            sigs.iter()
+                .map(|s| UserSignature::from_base64(&s.0))
+                .collect::<Result<Vec<_>, _>>()?
+        } else {
+            vec![]
```
thibault-martinez: I don't know what to think about this, can't really have a `SignedTransaction` without signature. How/when would the subscription actually return None?

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
DaughterOfMars: We still wouldn't be preventing that. Plus, what if you want
```rust
SubscriptionTransactionFilter {
  kind: Some(...),
  ..Default::default()
}
```

thibault-martinez: Just feels a bit wrong to give a shorthand that simply does not work, loosing the `..` syntax feels like a small price to pay to me. But keep it if you want, it's fine.

DaughterOfMars: I actually think it's one of the best ways to use structs like this

thibault-martinez: I would agree in most cases but not this one. We're making it easier for the user to misuse it. Feel free to resolve though.

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
```diff
@@ -0,0 +1,220 @@
...
+pub struct TransactionsSubscriptionArgs {
+    pub start_after: Option<String>,
+    pub filter: Option<SubscriptionTransactionFilter>,
+}
+
+// ===========================================================================
+// Subscription filters
+// ===========================================================================
+
+/// Filter incoming events in a subscription. Exactly one field may be set
```
thibault-martinez: Maybe change may to must?

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
```diff
@@ -0,0 +1,208 @@
...
+
+// ===========================================================================
+// Payload unions
+// ===========================================================================
+
+#[derive(cynic::InlineFragments, Debug)]
+#[cynic(schema = "rpc", graphql_type = "EventSubscriptionPayload")]
+#[non_exhaustive]
+pub enum EventSubscriptionPayload {
+    Event(Box<SubscriptionEvent>),
```
thibault-martinez, line 104: Bit annoying that this is not the same word order. I guess it comes from the graphQL but we could technically change the rust name. Would cynic be unhappy about it?

crates/iota-sdk-graphql-client/src/query_types/transaction.rs
```diff
@@ -228,19 +226,8 @@ impl TryFrom<TransactionBlock> for SignedTransaction {
...
-        } else {
-            vec![]
-        };
-
         if let Some(transaction) = transaction {
-            Ok(SignedTransaction {
-                transaction: transaction.0.transaction,
-                signatures,
-            })
+            Ok(transaction.into())
```
thibault-martinez, line 229: Was there a reason we were doing this? @Thoralf-M do you know?

thibault-martinez (approved):

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
DaughterOfMars: May means in this context that only one field may be set

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
DaughterOfMars: What do you mean?

crates/iota-sdk-graphql-client/src/query_types/transaction.rs
Thoralf-M: So we also get the signatures and can verify them?

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
thibault-martinez: Yeah but to me it also suggests than none can be set, which is not the case

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
thibault-martinez: EventSubscription vs SubscriptionEvent

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
DaughterOfMars: That's not the case though. It says exactly one field

crates/iota-sdk-graphql-client/src/query_types/transaction.rs
Thoralf-M: This can be deleted now?

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
DaughterOfMars: But it's not that. It's an Event from an EventSubscription, thus, SubscriptionEvent

crates/iota-sdk-graphql-client/src/query_types/transaction.rs
Thoralf-M: Checked and at some point bcs didn't contain the signature, then it got changed to include it, but the field was not removed, so this is just cleaning up now

crates/iota-sdk-graphql-client/src/query_types/transaction.rs
DaughterOfMars: Looks like you're right, even though claude intentionally did not

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
thibault-martinez: I still think this is very confusing, and `must` would eliminate any doubt.

Thoralf-M (approved):

crates/iota-sdk-graphql-client/src/query_types/subscriptions.rs
DaughterOfMars: alright, fine, I changed it

PR #1195, opened 2026-06-09
refactor: remove deprecated zklogin variants from public signature enums

### LINKED ISSUE #1076: Can we remove ZkLogin from public enums?
They are kept for now to avoid shifting BCS discriminants but they could be kept only on a serde-specific struct.

Thoralf-M:
Removes the deprecated `ZkLoginAuthenticatorDeprecated` variant from the public `UserSignature` and `SignatureScheme` enums (and the FFI mirror). Follow-up to #1076.

Neither enum is serialized by variant index — `UserSignature` uses a custom flag-byte encoding, `SignatureScheme` an explicit `#[repr(u8)]` — so this changes no BCS bytes for any supported signature; `0x05` is now a reserved/rejected flag. The index-locked placeholders (`BinaryPublicKey`, `MemberSignature`, `TransactionKind`) are deliberately kept, since removing those *would* shift the on-chain wire format.

**Tests:** added hardcoded-tag anchors (round-trips can't catch a discriminant shift): `signature_scheme_flag_values` pins every flag byte and that `0x05` is rejected; `user_signature_rejects_zklogin_flag`; `member_signature_bcs_tags` pins the multisig member tags (`passkey = 0x04`).

https://claude.ai/code/session_01GKkvevz4vSZike1ryqCQMz


muXxer (approved):

DaughterOfMars (approved):

thibault-martinez:
let's huddle quickly about that, I'm not too sure

thibault-martinez (approved):

PR #1201, opened 2026-06-09
feat: add grpc to FFI/bindings

### LINKED ISSUE #1200: Add GRPC to FFI/bindings

thibault-martinez:
## Summary

Exposes the gRPC client through the FFI crate and all five language bindings (Python, Go, Kotlin, Swift, C#), which so far only had GraphQL.

- New `grpc` module in `iota-sdk-ffi` (on by default): `GrpcClient` with all non-streaming convenience methods — checkpoints, epoch, gas price, health, service info, objects, transactions, coin info, owned objects, dynamic fields, package versions, and execute/simulate. Proto responses are converted eagerly into the core FFI types; read masks are plain string lists, and fields not covered by the mask come back as `None`.
- `TransactionBuilder` gains `with_grpc_client()`, so transactions can be resolved, dry-run, and executed entirely over gRPC. The existing `ClientTransactionBuilder` API is unchanged (internally dispatches over either client).
- Fixes the umbrella `iota-sdk` `grpc` feature, which pulled the gRPC client without any TLS, making the HTTPS constructors fail at runtime. It now forwards `tls-ring` + `tls-webpki-roots` (webpki rather than native roots so the published Android/iOS artifacts work, matching reqwest's behavior for GraphQL).
- One `grpc_service_info` example per language.

Checkpoint streaming is left for a follow-up (needs a uniffi callback/iterator design).

## Test plan

- `make ci` passes; clippy/check also clean with `--no-default-features`.
- All five binding examples run against testnet.
- Smoke-tested the whole surface through the Python bindings against testnet, including read-mask variants, pagination, and builder `finish`/`dry_run`/`execute` over gRPC.

Note: gRPC pulls `zstd-sys` (C) into every binding artifact; iOS compile-checked locally, but Android/manylinux only build in nightly/publish CI — a `make kotlin-android` smoke run before merging would de-risk that.


Thoralf-M:
Didn't verify if this actually works, but looks like with some changes we could get gRPC working in wasm32 https://github.com/iotaledger/iota-rust-sdk/commit/c67fc6c46cef39ceb5ddb994ec45cd6c01ff4718 (requires a gRPC-Web-capable reverse proxy — e.g. Envoy or Traefik — in front of the gRPC endpoints to translate gRPC-Web to gRPC and add CORS)

thibault-martinez:
split it in smaller PRs

PR #1218, opened 2026-06-15
feat: `MoveAuthenticator` improvements

thibault-martinez:


thibault-martinez:
Need to think about it a bit more

PR #1228, opened 2026-06-18
fix: accept optional 0x prefix in FFI hex_decode

### LINKED ISSUE #1213: wasm bindings: accept 0x prefix in hexDecode
## Summary

Feature request: the WASM/TS bindings export `base64Encode`/`base64Decode` and `hexEncode`/`hexDecode`, but **no `base58Encode`/`base58Decode`** — even though base58 is the canonical display encoding for digests on IOTA.

`Digest.fromBase58`/`toBase58` exist but only handle exactly 32 bytes, so any tool that needs general base58 (parsing user input, rendering arbitrary byte strings the way explorers do) has to ship its own implementation alongside the SDK.

## Suggestion

Export `base58Encode(input: ArrayBuffer): string` and `base58Decode(input: string): ArrayBuffer` from `iota-sdk-ffi` next to the existing base64/hex helpers (`bs58` is already in the dependency tree via `iota-sdk-types`).

## Related ergonomic nit

`hexDecode("0x...")` throws `Invalid character 'x'` — callers must strip the `0x` prefix manually. Since `0x`-prefixed hex is the dominant on-chain format, accepting (or at least documenting) the prefix would remove a footgun. Happy to split this into its own issue if preferred.

## Environment

iota-rust-sdk @ 73e2db6c, `make wasm` output (`dist/types/iota_sdk_ffi.d.ts` contains no base58 function exports).

Thoralf-M:
## Why

`hexDecode("0x…")` threw `Invalid character 'x'`, forcing callers to strip the prefix manually — awkward given `0x`-prefixed hex is the dominant on-chain format. (Ergonomic nit noted in #1213.)

## What

`hex_decode` in `iota-sdk-ffi` now strips an optional leading `0x` before decoding. Added a unit test covering both prefixed and unprefixed input.

[run-ci]

https://claude.ai/code/session_012DeyuE6Qx9ADpchRPTqSdf

---
_Generated by [Claude Code](https://claude.ai/code/session_012DeyuE6Qx9ADpchRPTqSdf)_

### DIFF
--- a/crates/iota-sdk-ffi/src/lib.rs
+++ b/crates/iota-sdk-ffi/src/lib.rs
@@ -38,8 +38,21 @@ pub fn hex_encode(input: &[u8]) -> String {
 
 #[uniffi::export]
 pub fn hex_decode(input: String) -> crate::error::Result<Vec<u8>> {
+    let input = input.strip_prefix("0x").unwrap_or(&input);
     Ok(hex::decode(input)?)
 }
 
 crate::export_primitive_types_bcs_conversion!(u8, u16, u32, u64, i8, i16, i32, i64, bool, String);
 crate::export_primitive_types_json_conversion!(u8, u16, u32, u64, i8, i16, i32, i64, bool, String);
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn hex_decode_accepts_optional_0x_prefix() {
+        let expected = vec![0xde, 0xad, 0xbe, 0xef];
+        assert_eq!(hex_decode("deadbeef".to_owned()).unwrap(), expected);
+        assert_eq!(hex_decode("0xdeadbeef".to_owned()).unwrap(), expected);
+    }
+}


DaughterOfMars (approved):

crates/iota-sdk-ffi/src/lib.rs
thibault-martinez, line 37: What do we think about this not having a prefix?

Thoralf-M: I'd lean toward keeping `hex_encode` unprefixed. This is a generic bytes↔hex primitive (sitting next to the equally bare `base64_encode`), not an address/ID formatter — the `0x` convention belongs to the typed values like `Address`/`ObjectId` that already render it via their own `Display`.

The asymmetry is intentional: lenient on input, canonical on output (Postel's law). `hex_decode` now accepts `0x` or bare, while `hex_encode` emits the minimal form and leaves prefixing to the caller. Round-trips still hold since decode strips the prefix. Happy to add a prefix if you feel the symmetry is worth it, though.

---

thibault-martinez (approved):

PR #1230, opened 2026-06-18
fix: apply PaginationFilter direction default in wasm bindings

### LINKED ISSUE #1209: wasm bindings: PaginationFilter record default for 'direction' not applied (panics with enum value 65536)
## Summary

In the WASM/TS bindings, constructing a `PaginationFilter` without an explicit `direction` produces a garbage enum value, and the next call using it panics with:

```
RustPanic: Failed to convert arg 'pagination_filter': Invalid Direction enum value: 65536
```

The generated record factory doesn't apply the Rust-side default for enum fields — `PaginationFilter.defaults()` returns `{ cursor: undefined, limit: undefined }` with **no `direction` key at all**, so the lowering serializes uninitialized memory.

## Repro (Node, against the `make wasm` output)

```js
import { initAsync, GraphQlClient, PaginationFilter } from "@iota/sdk-wasm";
await initAsync();

console.log(PaginationFilter.defaults()); // { cursor: undefined, limit: undefined } — direction missing

await client.activeValidators(undefined, PaginationFilter.new({ limit: 2 }));
// → RustPanic: Invalid Direction enum value: 65536
```

Passing `direction: Direction.Forward` explicitly works fine.

## Expected

`PaginationFilter.new({ limit: 2 })` should fall back to the `#[uniffi(default)]` direction (Forward), matching the Rust API.

## Environment

- iota-rust-sdk @ 73e2db6c, `make wasm`, Node v26, macOS arm64
- Likely a uniffi-bindgen-react-native record-defaults issue (enum field defaults), so may need an upstream report or a generator patch in `fix-generated-typescript.mjs`.

Thoralf-M:
Constructing `PaginationFilter` in JS without `direction` left the field `undefined`; the enum converter lowered it to a garbage ordinal and the next call panicked (`Invalid Direction enum value: 65536`).

uniffi record metadata can't encode an enum-variant default, so ubrn omits `direction` from the generated `defaults()`. This injects the Rust-side default (`Direction.Forward`) into the `PaginationFilter` factory in `fix-generated-typescript.mjs`, so omitting `direction` now matches the Rust API and the field becomes optional in `.new({…})`. The patch is idempotent.

Upgrading the toolchain doesn't fix it: the latest `uniffi-bindgen-react-native` still renders an enum type-default as `undefined`, reproducing the same bug.

Closes #1209

https://claude.ai/code/session_01XbXPUrsd7PuttHeHCyNMa5

---
_Generated by [Claude Code](https://claude.ai/code/session_01XbXPUrsd7PuttHeHCyNMa5)_

### DIFF
--- a/bindings/wasm/scripts/fix-generated-typescript.mjs
+++ b/bindings/wasm/scripts/fix-generated-typescript.mjs
@@ -40,6 +40,27 @@ fixed = fixed.replace(
   'from "./wasm-bindgen/index_bg.js"'
 );
 
+// Restore the default for `PaginationFilter.direction`.
+// uniffi record metadata can only encode literal defaults (numbers, strings,
+// bools, None, Some(...), []), never an enum variant, so the `#[default]`
+// `Direction::Forward` on the Rust field cannot be carried into the bindings.
+// As a result ubrn emits `direction` as a required field and omits it from the
+// record's `defaults()`. In TypeScript that surfaces as a type error, but in
+// plain JS the field is silently `undefined`, the enum converter lowers it to a
+// garbage ordinal, and the next call panics ("Invalid Direction enum value").
+// Upgrading does not help: uniffi's bare `#[uniffi(default)]` keyword exists
+// (>=0.30) but ubrn deliberately renders an enum type-default as `undefined`
+// ("no canonical first variant"), reproducing the same bug. So we inject the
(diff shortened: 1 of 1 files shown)


thibault-martinez (approved):
Ewwww

DaughterOfMars (approved):

PR #1231, opened 2026-06-18
feat: changes to `MoveAuthenticator`

thibault-martinez:
## Summary

Rounds out the `MoveAuthenticator` API surface ahead of its use in the Account Abstraction flow.

- Adds version-agnostic accessors on `MoveAuthenticator` (`version`, `address`, `call_args`, `type_args`, `object_to_authenticate`) that delegate to the inner version, so callers don't have to match on the enum.
- Derives `From` and implements `Hash` (over the canonical serialized bytes), and adds a `digest()` helper consistent with the other type digests.
- Tidies the `to_bytes`/`from_bytes` round-trip.

## Test plan

- New proptest covers the `to_bytes`/`from_bytes` round-trip.
- `make test`


crates/iota-sdk-types/src/crypto/move_authenticator.rs
```diff
@@ -187,3 +220,142 @@ mod serialization {
...
+        assert!(MoveAuthenticator::from_bytes(&[]).is_err());
+    }
+
+    #[test]
+    fn from_bytes_rejects_flag_only() {
+        let flag = SignatureScheme::MoveAuthenticator as u8;
+        assert!(MoveAuthenticator::from_bytes(&[flag]).is_err());
+    }
+
+    // // ---- Signable / SignableBytes round-trip tests ----
```
thibault-martinez: Waiting for some comments reply on the monorepo to decide what to do with these

crates/iota-sdk-types/src/crypto/move_authenticator.rs
```diff
@@ -18,6 +20,43 @@ pub enum MoveAuthenticator {
...
+        match self {
+            Self::V1(v1) => v1.type_args(),
+        }
+    }
+
+    pub fn object_to_authenticate(&self) -> &Input {
+        match self {
+            Self::V1(v1) => v1.object_to_authenticate(),
+        }
+    }
```
Thoralf-M: How sure are we that other versions will have the same method?

crates/iota-sdk-types/src/crypto/move_authenticator.rs
```diff
@@ -18,6 +20,43 @@ pub enum MoveAuthenticator {
...
 impl MoveAuthenticator {
     crate::def_is_as_into_opt!(V1(MoveAuthenticatorV1));
+
+    pub fn version(&self) -> u64 {
+        match self {
+            Self::V1(_) => 1,
+        }
+    }
+
+    pub fn address(&self) -> Address {
```
Thoralf-M: Should add a doc comment

crates/iota-sdk-types/src/crypto/move_authenticator.rs
```diff
@@ -187,3 +220,142 @@ mod serialization {
...
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    use crate::{Digest, SignatureScheme};
+
+    #[cfg(feature = "proptest")]
+    #[test_strategy::proptest]
+    fn to_from_bytes_roundtrip(authenticator: MoveAuthenticator) {
```
Thoralf-M, line 189: Can we also add a test with a fixture and not only roundtrips?

crates/iota-sdk-types/src/crypto/move_authenticator.rs
thibault-martinez: Already like that in the monorepo, @iotaledger/vm-language ?

crates/iota-sdk-types/src/hash.rs
Thoralf-M, line 365: The old node `digest()` hashed `"MoveAuthenticator::" ‖ bcs(inner)`; the SDK version hashes the enum through its custom `Serialize` (which emits `serialize_bytes(to_bytes())`), i.e. `"MoveAuthenticator::" ‖ ULEB(len) ‖ flag ‖ bcs(inner)`. The extra length-prefix + scheme-flag bytes mean a different preimage, hence a different digest.
Verified against PR #1231's crate for a simple authenticator:

```
OLD digest : 06403c2123b204e3f93dade02c8b091e7e03b4c46883ff6d85054af81c2017d9
NEW digest : 1fa801e88b674bcab0211fa01654324b3341604b6e8c91b1e609ccfd2bca6ece
```

Thoralf-M: Also add a test with fixture for this

crates/iota-sdk-types/src/crypto/move_authenticator.rs
Thoralf-M: I think we said we wanted to change it when we move to the SDK to have it more future proof here. For usability in the monorepo we can add them with an extension trait

crates/iota-sdk-types/src/crypto/move_authenticator.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1231/commits/40a600b3279702e5537d3822216bd45e09d18883

crates/iota-sdk-types/src/crypto/move_authenticator.rs
thibault-martinez: Sure, but if they guarantee other versions will be only additions and this is the basic API, maybe we can keep them

Thoralf-M: For object_to_authenticate() I could imagine that at some point we would have something incompatible, could be multiple objects, could be new kind of account without object if we have account balances...

crates/iota-sdk-types/src/hash.rs
thibault-martinez: This should fix it https://github.com/iotaledger/iota-rust-sdk/pull/1231/commits/b09623c3584aee73ad66ddb7f72ea088c8009080
But please try again

Thoralf-M:
```
// now (broken): emits bare enum, no flag/length
UserSignature::MoveAuthenticator(move_auth) => move_auth.serialize(serializer),
// fix: mirror the other variants' on-wire form
UserSignature::MoveAuthenticator(move_auth) => serializer.serialize_bytes(&move_auth.to_bytes()),
```

Thoralf-M: But digest is fine now

crates/iota-sdk-types/src/crypto/move_authenticator.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1231/changes/167bf860e8b5329a9d9fcb23c774bbee8cefda8c

crates/iota-sdk-ffi/src/types/crypto/move_authenticator.rs
```diff
@@ -42,32 +41,34 @@ pub struct MoveAuthenticatorV1(pub iota_sdk::types::MoveAuthenticatorV1);
 impl MoveAuthenticatorV1 {
     /// Create a new move authenticator from an immutable object.
     #[uniffi::constructor]
-    pub fn new_immutable(
+    pub fn with_immutable_account_object(
```
DaughterOfMars: I don't really like the `with_` prefix, as it implies that this is a chainable function and not a constructor

crates/iota-sdk-types/src/crypto/move_authenticator.rs
miker83z: Currently we plan to keep the MoveAuthenticator always tied to an on-chain object. If you think it's not future proof it's ok to remove (even though with the same approach we could remove call_args and the rest 😅).

crates/iota-sdk-ffi/src/types/crypto/move_authenticator.rs
thibault-martinez: vm-lang requested the name change so please check with them before reverting (I see it's been done already)

thibault-martinez: I see it's only been renamed in the FFI though, why?

crates/iota-sdk-types/src/crypto/move_authenticator.rs
Thoralf-M: I checked with the other types in the SDK and decided to remove them so it's also aligned with them

DaughterOfMars (approved):

crates/iota-sdk-ffi/src/types/crypto/move_authenticator.rs
Thoralf-M: Renamed everywhere with new_ prefix now https://github.com/iotaledger/iota-rust-sdk/pull/1231/changes/7321222d188f390afa5482c2f626286bb59d1a7f

DaughterOfMars (approved):

crates/iota-sdk-types/src/crypto/move_authenticator.rs
```diff
@@ -1,42 +1,98 @@
...
+    prop_oneof![
+        any::<ObjectReference>().prop_map(Input::ImmutableOrOwned),
+        any::<SharedObjectReference>().prop_map(Input::Shared),
+    ]
+}
+
 impl MoveAuthenticatorV1 {
     /// Create a new move authenticator from an immutable object.
-    pub fn new_immutable(
+    pub fn new_with_immutable_account_object(
```
Alex6323, line 95: Nit: should this be called `new_from_immutable_account_object` instead?. The docs seem to suggest that this would be the more fitting name.

Thoralf-M: I'd rather say the docs should be changed to `with`. The object is just referenced here, we don't get anything `from` the object to construct it
What do you think?

Alex6323: fine with it

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1231/commits/738b45aba322f7a68e7f2a50a16aba34eebf64ce

Alex6323 (approved):

PR #1232, opened 2026-06-22
feat: add distinct digest newtype wrappers

### LINKED ISSUE #11349: Address context specific Digest concerns
Maybe I am missing something, but it feels wrong to replace all the 12 old tuple-struct wrappers with aliases to a single type for several reasons:

- All new aliases are now serializable, while old `CertificateDigest` and `SenderSignedDataDigest` were not.
- Many did not have the `Default` impl; now all new aliases have and return `ZERO`.
- Many did not have the `ZERO` const, `random()`, or `generate()` constructors; now all have them.
- `GENESIS_MARKER` was a transaction-digest-only concept; now all have it. `MisbehaviorReportDigest::GENESIS_MARKER` is now perfectly valid but makes no sense.
- `OBJECT_{DELETED,WRAPPED,CANCELLED}` and `is_object_{deleted,wrapped,alive}()` were object-digest-only concepts. Now `TransactionDigest::OBJECT_DELETED` is perfectly valid but makes no sense.
- Functions like `fn(digest: TransactionDigest)` and `fn(digest: ObjectDigest)` will not be treated by the compiler as different.
- `Debug` impl for some types like `ObjectDigest` are changed. So now debugging `Digest(...)` becomes harder; tracing logs with `?digest` lose per-domain prefix, potentially breaking grep pipelines.
- None of the aliases have docstrings, while some of the previous wrappers had them, properly explaining what each per-domain digest means.

Some of these are simply mentioned in the PR description, but none is properly explained or argued. Can you please give more context here?

Also, the `is_alive` -> `is_object_alive` rename is missing in the release notes.

_Originally posted by @roman1e2f5p8s in https://github.com/iotaledger/iota/pull/11338#pullrequestreview-4183349841_

DaughterOfMars:
## Why

We currently use a single `Digest` type everywhere a 32-byte Blake2b hash appears, so nothing stops a `TransactionDigest` being passed where an `ObjectDigest` is expected. This brings the per-domain digest newtypes that the monorepo used to own (removed in iotaledger/iota#11338) into the SDK.

## What

- Adds 12 newtype wrappers over `Digest`: `CheckpointDigest`, `CheckpointContentsDigest`, `CertificateDigest`, `SenderSignedDataDigest`, `TransactionDigest`, `TransactionEffectsDigest`, `TransactionEventsDigest`, `EffectsAuxDataDigest`, `ObjectDigest`, `ConsensusCommitDigest`, `MisbehaviorReportDigest`, `MoveAuthenticatorDigest`. The shared API (construction, encoding, ordering) is generated by a macro.
- Moves domain-specific behaviour off the bare `Digest`: object-state markers (`OBJECT_DELETED`/`WRAPPED`/`CANCELLED`) and `is_alive`/`is_deleted`/`is_wrapped` now live on `ObjectDigest`; the genesis marker lives on `TransactionDigest`.
- Rewires core type fields (checkpoint, object, effects, transaction) and `*::digest()` methods to the precise wrapper types, and propagates the change through the FFI / gRPC / GraphQL / transaction-builder crates.

Each wrapper is a transparent newtype, so **BCS and JSON encodings are byte-for-byte identical to a bare `Digest`** — there is no wire-format change. A proptest asserts this for all 12. The regenerated `bcs-schema.abnf` is purely more descriptive (e.g. `transaction-digest = digest`).

> [!NOTE]
> This is an intentional API-breaking change for downstream consumers (struct fields now carry distinct types). The four wrappers with no SDK field yet (`CertificateDigest`, `SenderSignedDataDigest`, `MisbehaviorReportDigest`, `MoveAuthenticatorDigest`) are exported so the monorepo can re-point its aliases.

## Test plan

- `make test` (451 unit tests), workspace doc tests, `cargo clippy --workspace --all-features --all-targets`, `make wasm32`, `make check-features`, `make check-sort-derives`, and `make bcs-schema` (committed) all pass.

Closes #11349


crates/iota-sdk-graphql-client/src/api/checkpoints.rs
```diff
@@ -96,7 +96,10 @@ impl Client {
 
     /// The total number of transaction blocks in the network by the end of the
     /// provided checkpoint digest.
-    pub async fn total_transaction_blocks_by_digest(&self, digest: Digest) -> Result<Option<u64>> {
+    pub async fn total_transaction_blocks_by_digest(
+        &self,
+        digest: CheckpointContentsDigest,
```
Thoralf-M: I think this should be CheckpointDigest

crates/iota-sdk-ffi/src/types/digest.rs
Thoralf-M, line 35: Do we still need this generic digest then?

DaughterOfMars: Yes, it is still used in some places

Thoralf-M (approved):

Alex6323 (approved):

PR #1235, opened 2026-06-24
fix(graphql): reconstruct CheckpointSummary from bcs

### LINKED ISSUE #1211: wasm bindings: checkpoint(digest) returns null for existing digests
## Summary

In the WASM/TS bindings, `GraphQlClient.checkpoint(digest, undefined)` returns `null/undefined` for digests that demonstrably exist. Lookup by sequence number works.

## Repro (Node, against the `make wasm` output, mainnet)

```js
import { initAsync, GraphQlClient, Digest } from "@iota/sdk-wasm";
await initAsync();

const seq = await client.latestCheckpointSequenceNumber();
const cp = await client.checkpoint(undefined, seq - 5n);   // works
const digest = cp.digest().toBase58();                      // e.g. 9MrqpfBs46hg92hB6TGR6gT6cmE1Qc3wQt51885sgMsb

const byDigest = await client.checkpoint(Digest.fromBase58(digest), undefined);
console.log(byDigest); // → undefined  (expected: the same checkpoint)
```

The same digest resolves fine through raw GraphQL (`checkpoint(id: { digest: $digest })` via `client.runQuery`), so the digest itself is valid and indexed — the typed path loses or mis-encodes the digest argument.

## Expected

`checkpoint(digest)` returns the checkpoint summary.

## Notes

May share a root cause with #1210 (optional args in the first parameter slots of this same API file). Workaround in use: resolve digest → sequence number via `runQuery`, then call `checkpoint(undefined, seq)`.

## Environment

iota-rust-sdk @ 73e2db6c, `make wasm`, Node v26, macOS arm64.

Thoralf-M:
Addresses #1211. Replaces the lossy `Checkpoint → CheckpointSummary` conversion — which defaulted `content_digest`/`version_specific_data` and so recomputed a wrong digest, making `checkpoint(digest)` lookups miss — with deserializing the full summary from the new `Checkpoint.bcs` field (iotaledger/iota#12020, released in v1.29.0).

`bcs` is optional, so a checkpoint indexed before the field existed surfaces a clear error instead of a wrong digest.

## Test plan
Localnet CI jobs exercise the `bcs` round-trip (`checkpoint(seq) → digest() → checkpoint(digest)`) against the v1.29.0 node. Also verified live against devnet, testnet, and mainnet: recent checkpoints return `bcs` and the digest round-trip passes; checkpoints indexed before the v1.29.0 deployment return `null` and surface the intended error.

https://claude.ai/code/session_01P3MFxXr93dq6zkrS9yXUzB


Thoralf-M:
Blocked until Mainnet 1.29.x/August 12th

crates/iota-sdk-graphql-client/src/api/checkpoints.rs
```diff
@@ -266,10 +266,8 @@ mod tests {
             .unwrap()
             .unwrap();
 
-        // TODO: https://github.com/iotaledger/iota-rust-sdk/issues/1211
-        let digest = chckp.content_digest.into_inner().into();
         let total_transaction_blocks_by_digest = client
-            .total_transaction_blocks_by_digest(digest)
+            .total_transaction_blocks_by_digest(chckp.digest())
```
Alex6323: PR is fine, it's just that this local variable lit up and I wondered, no I didn't wonder, I feel its name is abstruse. You can change it, you can leave it, but one thing is certain. I DO mind.

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1235/changes/1c6552a54c826f64744b499e9ecf61749481d03e

DaughterOfMars (approved):

crates/iota-sdk-graphql-client/src/query_types/checkpoint.rs
```diff
@@ -82,106 +78,19 @@ pub struct CheckpointId {
...
-#[derive(cynic::QueryFragment, Debug)]
-#[cynic(schema = "rpc", graphql_type = "GasCostSummary")]
-pub struct GasCostSummary {
-    pub computation_cost: Option<BigInt>,
-    pub computation_cost_burned: Option<BigInt>,
-    pub non_refundable_storage_fee: Option<BigInt>,
-    pub storage_cost: Option<BigInt>,
-    pub storage_rebate: Option<BigInt>,
+    /// BCS serialization of the `CheckpointSummary`, Base64-encoded. `None` for
+    /// checkpoints indexed before the field was available.
```
thibault-martinez: Is it really worth saying the second sentence if checkpoints are kept 30 days? Wouldn't be a matter in a month anymore?

Thoralf-M: It stays relevant beyond pruned networks: devnet still serves checkpoints from genesis with `bcs: null`, and any indexer that isn't re-indexed keeps the null rows indefinitely. Happy to drop it if you still prefer.

---

thibault-martinez (approved):

crates/iota-sdk-graphql-client/src/query_types/checkpoint.rs
thibault-martinez: And for such instances, wouldn't it be nice to have a poor man's solution like we had before? Now we can't fetch CheckpointSummary for these at all?

Thoralf-M: Depends on the indexer, for our public indexers yes if there are others that don't prune it might still be relevant, but I guess you're right that it doesn't belong here

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1235/changes/63ad96a33bd7b819a81c3e5a87c1c528d1aea58d

PR #1239, opened 2026-06-30
feat: implement ClaimAccount transaction kind

valeriyr:
TODO

### DIFF
--- a/crates/iota-sdk-types/src/transaction/mod.rs
+++ b/crates/iota-sdk-types/src/transaction/mod.rs
@@ -6,7 +6,7 @@ use super::{
     Address, CheckpointTimestamp, Digest, EpochId, Event, GenesisObject, Identifier, ObjectId,
     ObjectReference, ProtocolVersion, RandomnessRound, TypeTag, UserSignature, Version,
 };
-use crate::utils::write_sep;
+use crate::{crypto::PublicKey, utils::write_sep};
 
 #[cfg(feature = "serde")]
 #[cfg_attr(doc_cfg, doc(cfg(feature = "serde")))]
@@ -183,6 +183,7 @@ pub struct RandomnessStateUpdate {
 ///                     =/ %d03                                        ; AuthenticatorStateUpdateV1Deprecated
 ///                     =/ %d04 (vector end-of-epoch-transaction-kind) ; EndOfEpoch
 ///                     =/ %d05 randomness-state-update                ; RandomnessStateUpdate
+///                     =/ %d06 claim-account-transaction              ; ClaimAccount
 /// ```
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
@@ -206,6 +207,8 @@ pub enum TransactionKind {
     EndOfEpoch(Vec<EndOfEpochTransactionKind>),
     /// Randomness update
     RandomnessStateUpdate(RandomnessStateUpdate),
+    /// User transaction for claiming/creating an Account object
+    ClaimAccount(ClaimAccountTransaction),
 }
 
 impl TransactionKind {
@@ -218,6 +221,7 @@ impl TransactionKind {
         Programmable(ProgrammableTransaction),
         Genesis(GenesisTransaction),
         EndOfEpoch(Vec<EndOfEpochTransactionKind>),
+        ClaimAccount(ClaimAccountTransaction),
     }
 
     /// Create a [`TransactionKind::Programmable`].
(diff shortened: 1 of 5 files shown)


crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -1555,3 +1566,248 @@ pub struct MoveCall {
...
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "proptest", derive(test_strategy::Arbitrary))]
+#[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
+pub struct SmartAccountClaim {
+    /// Signature-scheme flag of the public key for the address being claimed:
+    /// `0x00` Ed25519, `0x01` Secp256k1, `0x02` Secp256r1, `0x03` MultiSig or
+    /// `0x06` Passkey.
+    pub public_key_scheme: u8,
+    /// Raw public key bytes, without the scheme flag prefix. For `MultiSig`
+    /// this is a BCS-encoded multisig public key.
```
Thoralf-M: The multisig address? Since there is no single multisig public key

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -1555,3 +1566,248 @@ pub struct MoveCall {
...
+    /// Whether the created account object is mutable or immutable.
+    pub build_kind: SmartAccountBuildKind,
+}
+
+impl SmartAccountClaim {
+    /// Creates a claim of the address derived from `public_key`, which is the
+    /// address a transaction signed by the corresponding private key is sent
+    /// from.
+    pub fn new(public_key: &PublicKey, build_kind: SmartAccountBuildKind) -> Self {
+        Self::new_unchecked(
```
Thoralf-M, line 1661: Why have `new` and `new_unchecked` if there is no check?

crates/iota-sdk-types/src/transaction/mod.rs
valeriyr: It is meant `MultisigCommittee`, which is serialized as bytes. It was previously named `MultiSigPublicKey`.

crates/iota-sdk-types/src/transaction/mod.rs
valeriyr: In `new_unchecked`, there is no check of public key bytes; yes, you can put arbitrary data

crates/iota-sdk-types/src/transaction/mod.rs
valeriyr: @Thoralf-M One more clarification. Our feature branch is super outdated, so it may not be fully aligned with the current SDK state. We'll need to review it one more time when it is ready, I suppose.

crates/iota-sdk-types/src/transaction/mod.rs
valeriyr: In `new` it is checked by design with the `PublicKey` type.

Thoralf-M: I see you want tos hare it between PublicKey and MultisigCommittee, any other reason to have the unchecked variant? If not I would make it private

Thoralf-M:
Please add in the PR description why we're adding this as new tx kind instead of using a normal PTB (consensus/sequenzer reasons)

crates/iota-sdk-types/src/transaction/mod.rs
valeriyr: Are you strongly against having this function?
I can't say that it is required since all the fields in the struct are public, but it can be useful for FFI conversion(it is used there) and for testing.
I would leave it if it doesn't break any SDK rules or conventions.
I will follow your suggestion if you think we should remove it.

valeriyr: We can also consider whether we need this function: `smart_account_claim_new_unchecked`.

Thoralf-M: No strong reason, just better to not expose functions if they are not really needed
I think in the FFI the unchecked fn can also be removed, users can just construct the types from bytes

valeriyr: done in cfe6b949

crates/iota-sdk-types/src/transaction/mod.rs
valeriyr: renamed in 1a55ed4a517787c01bec5b6e2bf4204e84ab6b75

Thoralf-M (approved):

miker83z (approved):

PR #1240, opened 2026-07-01
feat: add getters and constructor to CheckpointSummary

DaughterOfMars:
Adds a `new` constructor, per-field getters, and two convenience accessors (`next_epoch_committee`, `is_last_checkpoint_of_epoch`) to `CheckpointSummary`, matching the accessor style already used by types like `Coin` and `MoveStruct`.

Fields stay public, so this is a non-breaking, additive change. It gives consumers a stable read/construct API ahead of a later field-privatization step, and lets the node adopt these accessors instead of maintaining its own extension-trait helpers.

## Test plan

- `cargo nextest run -p iota-sdk-types --all-features` — 329 pass.
- `cargo check -p iota-sdk-types --target wasm32-unknown-unknown` — builds.
- clippy clean; formatted with nightly rustfmt.


crates/iota-sdk-types/src/checkpoint.rs
```diff
@@ -169,6 +169,102 @@ pub struct CheckpointSummary {
     pub version_specific_data: Vec<u8>,
 }
 
+impl CheckpointSummary {
+    /// Construct a `CheckpointSummary` from its constituent parts.
+    #[expect(clippy::too_many_arguments)]
+    pub fn new(
```
Thoralf-M, line 179: Could also update the ffi to make use of this, but not much of a difference so feel free to resolve without doing

Thoralf-M (approved):

crates/iota-sdk-types/src/checkpoint.rs
DaughterOfMars: done

Thoralf-M (approved):

crates/iota-sdk-types/src/checkpoint.rs
```diff
@@ -169,6 +169,102 @@ pub struct CheckpointSummary {
...
+        }
+    }
+
+    /// The epoch that this checkpoint belongs to.
+    pub fn epoch(&self) -> EpochId {
+        self.epoch
+    }
+
+    /// The height of this checkpoint.
+    pub fn sequence_number(&self) -> CheckpointSequenceNumber {
```
thibault-martinez, line 211: Wondering if keeping the SequenceNumber name, since we renamed everything to version, is really wise

thibault-martinez (approved):

crates/iota-sdk-types/src/checkpoint.rs
DaughterOfMars: But it's a different type

thibault-martinez: So what? Same concept

DaughterOfMars: I mean it might make sense to re-use the type but does it make sense to rename the methods? I dunno

thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/issues/1250

PR #1241, opened 2026-07-01
fix: debug-print byte containers as base64 in iota-sdk-types

### LINKED ISSUE #1220: Debug-print `Vec<u8>` fields as base64 instead of per-byte numeric lists
## Summary

Types in `iota-sdk-types` that hold byte containers (`Vec<u8>`, `Vec<Vec<u8>>`, `BTreeMap<_, Vec<u8>>`) use `#[derive(Debug)]`, which renders **every byte on its own line as a decimal number**. For anything non-trivial this dwarfs the rest of the output and makes `{:?}` practically unusable.

Concrete example: debug-printing a value that contains a framework `MovePackage` (e.g. `0x2`) expands to **80,000+ lines**, almost all of it the per-byte rendering of compiled module bytecode. (This surfaced downstream while adding `Debug` to `ExecutionResult` in the `iota-vm-sdk` crate of the `iota` monorepo: a single dev-inspect result printed ~84k lines, ~82k of which were package bytecode bytes.)

## Request

Render byte-container fields in `Debug` as a **base64 string** — no byte-length suffix, no truncation:

```text
contents: "oRzrCwYAAAAL..."
```

instead of:

```text
contents: [
    161,
    28,
    235,
    ...        // tens of thousands of lines
]
```

## Why this is consistent, and not a breaking change

- The crypto key/signature types **already do exactly this**: their manual `impl Debug` delegates to the `Base64Display{32,33,48,64,96}` helpers (`crates/iota-sdk-types/src/crypto/mod.rs`; e.g. `Ed25519PublicKey` Debug at `crypto/ed25519.rs:115`). Variable-length byte fields are simply the gap.
- Serde already encodes these fields as base64 in the human-readable format (`ReadableBase64Encoded` / `IfIsHumanReadable`). `Debug` should match that intent.
- `Debug` output is **not part of the stable API surface**, so changing the format is not a breaking change (relevant given the SDK's backward-compat policy — this is additive in spirit).

## Suggested mechanism

`derive_more` (feature `full`) and `base64ct` are already dependencies. Two options:

1. Add a variable-length base64 `Debug` wrapper for `&[u8]` (mirroring the fixed-size `Base64DisplayN` pattern in `crypto/mod.rs`), then switch the affected types to `#[derive(derive_more::Debug)]` and annotate byte fields, e.g. `#[debug("{}", Base64Display(field))]`. Keeps the impls declarative.
2. Hand-write `impl Debug` for the affected types, following the existing crypto precedent.

Exact approach is up to maintainers.

## Affected types / fields (`iota-sdk-types`)

All of these `#[derive(Debug)]` and hold byte containers:

| Type | Field | Location |
|---|---|---|
| `Event` | `contents: Vec<u8>` | `events.rs:35` |
| `MoveStruct` | `contents: Vec<u8>` | `object.rs:285` |
| `MovePackage` | `modules: BTreeMap<Identifier, Vec<u8>>` | `move_package.rs:144` |
| `MovePackageData` | `modules: Vec<Vec<u8>>` | `move_package.rs:58` |
| `CheckpointSummary` | `version_specific_data: Vec<u8>` | `checkpoint.rs:123` |
| `RandomnessStateUpdate` | `random_bytes: Vec<u8>` | `transaction/mod.rs:157` |
| `SystemPackage` | `modules: Vec<Vec<u8>>` | `transaction/mod.rs:780` |
| `Publish` | `modules: Vec<Vec<u8>>` | `transaction/mod.rs:1315` |
| `Upgrade` | `modules: Vec<Vec<u8>>` | `transaction/mod.rs:1369` |
| `Input` | `Pure(Vec<u8>)` variant | `transaction/mod.rs:874` |
| `PasskeyAuthenticator` | `challenge`, `authenticator_data: Vec<u8>` | `crypto/passkey.rs:35` |

Note: `Input::Pure` and the `PasskeyAuthenticator` fields don't have base64 serde today either, so their derived `Debug` is the only thing rendering them and they're the most affected.

## Related (other crates, lower priority)

- `iota-sdk-graphql-client` output types with raw `bcs: Vec<u8>` fields that derive `Debug`: `DryRunMutation`, `DryRunReturn`, `DynamicFieldName`, `DynamicFieldValue` (`crates/iota-sdk-graphql-client/src/output_types/mod.rs`).
- Internal aggregator working buffers (`message: Vec<u8>`) in `iota-sdk-crypto` (`multisig.rs`, `validator.rs`) — implementation details, optional.

## Out of scope

Fixed-size crypto newtypes (`[u8; N]` keys/signatures) already have custom base64 `Debug` — no change needed there.

Thoralf-M:
## Summary

`Debug`-printing types that hold byte containers (`Vec<u8>`, `Vec<Vec<u8>>`, `BTreeMap<_, Vec<u8>>`) rendered every byte on its own line as a decimal number, which is impractical for anything non-trivial (e.g. a `MovePackage`'s compiled module bytecode can dwarf everything else with tens of thousands of lines).

This renders those fields as base64 strings instead, matching how `serde` already encodes them in the human-readable format. `Debug` output is not part of the stable API surface, so this isn't a breaking change.

Adds a small `base64_debug` module with `Base64Debug`/`Base64DebugList`/`Base64DebugMap` helpers, and switches the affected types from `#[derive(Debug)]` to a hand-written `impl Debug` (mirroring the existing pattern used for the fixed-size crypto key/signature types) so the byte fields can be routed through the new helpers.

Affected types: `Event`, `MoveStruct`, `MovePackage`, `MovePackageData`, `CheckpointSummary`, `RandomnessStateUpdate`, `SystemPackage`, `Publish`, `Upgrade`, `Input::Pure`, and `PasskeyAuthenticator`.

Fixes #1220.

### Example

`{:#?}` on an `Event` with 48 bytes of `contents`, before:

```text
Event {
    ...
    contents: [
        0,
        1,
        2,
        3,
        ...
        47,
    ],
}
```

after:

```text
Event {
    ...
    contents: "AAECAwQFBgcICQoLDA0ODxAREhMUFRYXGBkaGxwdHh8gISIjJCUmJygpKissLS4v",
}
```

## Test plan

- [x] `cargo test --all-features -p iota-sdk-types` (unit + doc tests) passes
- [x] `cargo clippy --all-features --all-targets -p iota-sdk-types` is clean
- [x] `cargo +nightly fmt`
- [x] Manually verified `{:?}`/`{:#?}` output before/after on an `Event`
[run-ci]

---
_Generated by [Claude Code](https://claude.ai/code/session_015998FVDvYonMFGT7WADeXi)_


crates/iota-sdk-types/src/crypto/passkey.rs
```diff
@@ -59,6 +59,24 @@ pub struct PasskeyAuthenticator {
     pub(crate) client_data_json: String,
 }
 
+impl std::fmt::Debug for PasskeyAuthenticator {
```
DaughterOfMars: We can use `derive_more` to avoid these I think

Thoralf-M: Nice, forgot about this, will try

Thoralf-M: Done — switched to `#[derive(derive_more::Debug)]` with per-field `#[debug(...)]` attributes that inline the base64 encoding call, and removed the `base64_debug` helper module entirely.

---

DaughterOfMars (approved):

thibault-martinez (approved):

PR #1244, opened 2026-07-01
fix: remove FaucetClient::new_testnet() as the testnet faucet is web-only

### LINKED ISSUE #1217: fix: testnet faucet endpoint is deprecated, breaking FaucetClient::new_testnet()
## Summary

Requesting gas from the **testnet** faucet through the SDK no longer works. The testnet faucet endpoint (`https://faucet.testnet.iota.cafe/v1/gas`) is deprecated and now only answers with a deprecation notice, so `FaucetClient::new_testnet()` fails for every request.

`FaucetClient::new_testnet()` already targets the new host (`crates/iota-sdk-graphql-client/src/faucet.rs`):

```rust
pub const FAUCET_TESTNET_HOST: &str = "https://faucet.testnet.iota.cafe";
pub fn new_testnet() -> Self { Self::new(FAUCET_TESTNET_HOST) }
```

so the constant itself is correct — the upstream endpoint it points at is the one that has been deprecated.

> **devnet and localnet are unaffected** — only the testnet faucet endpoint is deprecated. `FaucetClient::new_devnet()` and `FaucetClient::new_localnet()` continue to work.

## Reproduction

Using the shipped `faucet` example (`crates/iota-sdk/examples/faucet.rs`) pointed at testnet:

```rust
use iota_sdk::{graphql_client::faucet::FaucetClient, types::Address};

#[tokio::main]
async fn main() -> eyre::Result<()> {
    let address =
        Address::from_hex("0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900")?;
    FaucetClient::new_testnet().request_and_wait(address).await?;
    Ok(())
}
```

```
$ cargo run --example faucet
Error: Faucet request was unsuccessful: This faucet is deprecated. Please use the faucet at https://faucet.testnet.iota.cafe
```

The error originates from the faucet server, which answers with a JSON error body instead of provisioning gas:

```
$ curl -s -X POST https://faucet.testnet.iota.cafe/v1/gas \
    -H 'Content-Type: application/json' \
    -d '{"FixedAmountRequest":{"recipient":"0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900"}}'
{"error":"This faucet is deprecated. Please use the faucet at https://faucet.testnet.iota.cafe"}
```

## Expected vs. actual

- **Expected:** a programmatic testnet faucet request returns gas, or the SDK clearly documents how testnet gas is now obtained.
- **Actual:** every testnet request fails with the deprecation message above; there is no working programmatic testnet endpoint behind `new_testnet()`.

## Suggested direction

- Confirm whether a programmatic testnet faucet endpoint still exists and, if so, repoint `FAUCET_TESTNET_HOST` at it.
- If testnet gas is now only available via a web/CAPTCHA faucet, document that on `new_testnet()` and in the examples, and consider whether `new_testnet()` should be deprecated/removed so callers aren't pointed at a dead endpoint.
- devnet/localnet need no changes.

Alex6323:
## Why

The testnet faucet is web-only — its programmatic `/v1/gas` endpoint returns a deprecation error — so a testnet `FaucetClient` constructor can't work.

## What

Since no stable release has shipped, remove `FaucetClient::new_testnet()` outright (graphql-client + FFI) rather than shipping it deprecated (per review). `FAUCET_TESTNET_HOST` stays as a documented public host constant; the coin-stream test already skips testnet.

Closes #1217


crates/iota-sdk-graphql-client/src/faucet.rs
```diff
@@ -100,6 +106,15 @@ impl FaucetClient {
...
     /// Create a new Faucet client connected to the `testnet` faucet.
+    ///
+    /// The testnet faucet is web-only: its programmatic gas endpoint
+    /// (`https://faucet.testnet.iota.cafe/v1/gas`) responds with a deprecation
+    /// error. Request testnet gas from the web faucet at
+    /// <https://faucet.testnet.iota.cafe>. [`FaucetClient::new_devnet`] and
+    /// [`FaucetClient::new_localnet`] expose working programmatic endpoints.
+    #[deprecated(
+        note = "the testnet faucet is web-only; its programmatic gas endpoint returns a deprecation error. Request gas from the web faucet at https://faucet.testnet.iota.cafe . new_devnet and new_localnet are unaffected."
+    )]
```
Thoralf-M: Since we didn't do a stable release yet, shouldn't we rather just delete it or do you think it's good to have the deprecation message so one can find the web faucet?

Alex6323: Claude was being conservative here, and I think I rather agree with you. So removing is fine for me. @thibault-martinez do you also agree?

thibault-martinez: Yeah, no need to release with deprecations already

crates/iota-sdk-graphql-client/src/faucet.rs
```diff
@@ -13,6 +13,12 @@ use tracing::{error, info};
 use crate::WaitForTx;
 
 pub const FAUCET_DEVNET_HOST: &str = "https://faucet.devnet.iota.cafe";
+/// Host of the testnet faucet.
+///
+/// The testnet faucet is web-only: the programmatic gas endpoint at this
+/// host (`/v1/gas`) responds with a deprecation error. Request testnet gas
+/// from the web faucet at <https://faucet.testnet.iota.cafe>. The devnet and
```
thibault-martinez: Feels a bit weird to repeat the faucet address

Thoralf-M (approved):

thibault-martinez (approved):

PR #1245, opened 2026-07-01
fix: forward pagination arguments in the checkpoints GraphQL query

### LINKED ISSUE #1210: wasm bindings: checkpoints() ignores its PaginationFilter (always returns the first page)
## Summary

In the WASM/TS bindings, `GraphQlClient.checkpoints(paginationFilter)` ignores its filter entirely: regardless of `direction`, `limit`, or `cursor`, it always returns the first page (checkpoints 0..19 with the server's default page size).

Other methods taking a `PaginationFilter` in a *later* parameter position (`transactions`, `activeValidators`) honor it correctly, so this looks specific to lowering an optional record in the **first** argument slot (the method is declared with `#[uniffi::method(default(pagination_filter = None))]`).

## Repro (Node, against the `make wasm` output, mainnet)

```js
import { initAsync, GraphQlClient, PaginationFilter, Direction } from "@iota/sdk-wasm";
await initAsync();

const page = await client.checkpoints(
  PaginationFilter.new({ direction: Direction.Backward, limit: 40 }),
);
console.log(page.data.length, page.data[0].sequenceNumber(), page.data.at(-1).sequenceNumber());
// → 20 0n 19n   (expected: the newest 40 checkpoints, ~154M range)
```

For contrast, the same filter shape works through `transactions(undefined, filter)` and `activeValidators(undefined, filter)`.

## Expected

`Direction.Backward, limit: 40` → the latest 40 checkpoints.

## Workaround

Routing checkpoint listing through `client.runQuery` with an explicit `checkpoints(last: N, before: $cursor)` GraphQL query.

## Environment

iota-rust-sdk @ 73e2db6c, `make wasm`, Node v26, macOS arm64. Possibly related to #1209 (same record type), but here even fully-populated filters are dropped.

Alex6323:
## Why

`GraphQlClient.checkpoints(paginationFilter)` ignored its filter entirely — any `direction`/`limit`/`cursor` returned the same default page (#1210).

Root cause is Rust-side, not the bindings: the cynic `CheckpointsQuery` declared `variables = "CheckpointsArgs"` but never attached `#[arguments(...)]` to the `checkpoints` connection field, so the generated GraphQL omitted `first/after/last/before`. The Rust GraphQL client had the same bug; the wasm bindings just surfaced it.

## What

- Attach `#[arguments(first: $first, after: $after, last: $last, before: $before)]` to the `checkpoints` field (matches `objects.graphql` and the sibling `CheckpointQuery`).
- Add an offline regression test asserting the generated query forwards the pagination arguments — the previous network test only checked "returns data" with the default filter, so it never caught this.

## Test plan

- New test `checkpoints_query_forwards_pagination_arguments` passes; confirmed it fails if the `#[arguments]` is removed.
- End-to-end against mainnet via the wasm bindings (reporter's exact repro):
  - Before: `checkpoints({direction: Backward, limit: 40})` → 20 items (default page).
  - After: → **40** items, newest sequence numbers; `{Forward, limit: 5}` → **5** items. Filter honored.
- `cargo clippy` (all targets) + `cargo +nightly fmt --check` clean.

Closes #1210

### DIFF
--- a/crates/iota-sdk-graphql-client/src/api/checkpoints.rs
+++ b/crates/iota-sdk-graphql-client/src/api/checkpoints.rs
@@ -149,6 +149,30 @@ impl Client {
 mod tests {
     use crate::{PaginationFilter, test_utils::test_client};
 
+    /// Guards #1210: the `checkpoints` field must forward its pagination
+    /// arguments. Without them the server returns its default page and every
+    /// `PaginationFilter` (direction, limit, cursor) is silently ignored.
+    #[test]
+    fn checkpoints_query_forwards_pagination_arguments() {
+        use cynic::QueryBuilder;
+
+        use crate::query_types::{CheckpointsArgs, CheckpointsQuery};
+
+        let operation = CheckpointsQuery::build(CheckpointsArgs {
+            first: Some(10),
+            after: None,
+            last: None,
+            before: None,
+        });
+        for arg in ["first:", "after:", "last:", "before:"] {
+            assert!(
+                operation.query.contains(arg),
+                "checkpoints query is missing the `{arg}` argument:\n{}",
+                operation.query
+            );
+        }
+    }
+
     #[tokio::test]
     async fn test_checkpoint_query() {
         let client = test_client();
--- a/crates/iota-sdk-graphql-client/src/api/checkpoints.rs
+++ b/crates/iota-sdk-graphql-client/src/api/checkpoints.rs
@@ -149,9 +149,6 @@ impl Client {
 mod tests {
     use crate::{PaginationFilter, test_utils::test_client};
 
-    /// Guards #1210: the `checkpoints` field must forward its pagination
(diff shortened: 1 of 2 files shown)


Thoralf-M (approved):

crates/iota-sdk-graphql-client/src/api/checkpoints.rs
```diff
@@ -149,6 +149,30 @@ impl Client {
 mod tests {
     use crate::{PaginationFilter, test_utils::test_client};
 
+    /// Guards #1210: the `checkpoints` field must forward its pagination
```
thibault-martinez: Let's not leave stuff like that in PRs please

Alex6323: I agree. I am removing it.

thibault-martinez (approved):

PR #1246, opened 2026-07-02
feat: add try_from_object to remaining key-object mirror types

### LINKED ISSUE #1226: `try_from_object.rs` doesn't cover everything
Seems like I still missed something to resolve this comment:
https://github.com/iotaledger/iota-rust-sdk/pull/1124#discussion_r3421021655

Alex6323:
## Why

`iota-sdk-move-types` mirrors on-chain Move objects, and a mirror of a `key`-ability object should be constructible from an `Object` with type-tag validation (`try_from_object` / `TryFrom<&Object>`). Only 16 of them had one; #1226 asks whether every type that could have one does. It didn't — an audit found 21 `key`-object mirrors missing a constructor, plus `Config<WriteCap>`.

Criterion used: a mirror gets `try_from_object` iff it models a `key` object (first field `id: UID`). Nested/data/event structs (e.g. `Balance`, the `*Event` types, dynamic-field-only inner state) are correctly excluded.

## What

- **22 new constructors** following the existing `Coin`/`Clock` pattern — `TryFrom<&Object>` for all, plus `try_from_object_with_type` for the 15 generic ones (coin caps/metadata, closed-loop token & transfer-policy families, `Display`, kiosk `PurchaseCap`, coin-manager caps, `Config`, `LabelerCap`; and non-generic `DenyList`, `Random`, `AuthenticatorState`, `PackageMetadataV1`, zkLogin `VerifiedID`/`VerifiedIssuer`, `UnverifiedValidatorOperationCap`).
- **15 new `StructTag` predicates** in `struct_tag.rs` (`VerifiedID` hand-written — the `ID` acronym doesn't round-trip the macro's snake-case, same as the existing `object::ID`).
- **Tests** in `try_from_object.rs`: 16 types reuse committed `.bcs` fixtures via the existing macros; 6 use synthetic values.

Impl-only — no struct fields or derives changed, so BCS wire shapes are untouched.

## Test plan

- `try_from_object` (`--features serde`): 103 pass; `fixture_roundtrip`: 55 pass (BCS unchanged).
- Each constructor pairs with the correct predicate; generics validate their type param.
- clippy `--all-features` clean, `fmt` clean, wasm32 compiles.

Closes #1226


crates/iota-sdk-move-types/src/packages/iota_framework.rs
```diff
@@ -460,6 +460,51 @@ pub mod coin {
...
+        /// Decode a [`RegulatedCoinMetadata<T>`] from an on-chain object,
+        /// validating that the object's type tag matches
+        /// `0x2::coin::RegulatedCoinMetadata<type_param>`.
+        ///
+        /// Escape hatch for type params only known at runtime; nothing ties
+        /// `type_param` to `T`. When the type param is known at compile
+        /// time, prefer the `TryFrom` impl.
+        pub fn try_from_object_with_type(
+            object: &iota_types::Object,
+            type_param: &iota_types::TypeTag,
```
DaughterOfMars: Is this actually necessary?

DaughterOfMars: Oh nvm

DaughterOfMars (commented):
What about FFI?

Alex6323:
> What about FFI?

If we already agree that the types that now have these ctors are the final ones, I can add them to the FFI already, otherwise I would just do it in a follow-up.

DaughterOfMars (approved):

DaughterOfMars:
> > What about FFI?
> 
> If we already agree that the types that now have these ctors are the final ones, I can add them to the FFI already, otherwise I would just do it in a follow-up.

A followup is fine with me

crates/iota-sdk-move-types/src/packages/iota_framework.rs
```diff
@@ -1562,6 +1785,47 @@ pub mod display {
...
+    where
+        T: serde::de::DeserializeOwned + crate::MoveType,
+    {
+        type Error = crate::FromObjectError;
+
+        fn try_from(object: &iota_types::Object) -> Result<Self, Self::Error> {
+            Self::try_from_object_with_type(object, &T::type_tag())
+        }
+    }
+
```
Thoralf-M: Seems pretty similar for all types, just the `tag.is_` and the type names + docs differ, would it maybe be worth it to use a macro?

Alex6323: I agree. Change incoming!

DaughterOfMars (approved):

Thoralf-M (approved):

crates/iota-sdk-types/src/move_core/struct_tag.rs
```diff
@@ -461,6 +475,30 @@ impl StructTag {
...
+    }
+
+    /// Checks if this is a zkLogin verified ID type
+    /// (`0x2::zklogin_verified_id::VerifiedID`).
+    pub fn is_verified_id(&self) -> bool {
+        self.address == Address::FRAMEWORK
+            && self.module == Identifier::from_static("zklogin_verified_id")
+            && self.name == Identifier::from_static("VerifiedID")
+            && self.type_params.is_empty()
+    }
```
thibault-martinez: Why are we adding zklogin stuff?

Alex6323: `VerifiedID` and `VerifiedIssuer` are `key` objects — they carry a `UID` and persist on-chain once zklogin verification creates them — so they fall under the same criterion as the rest of this PR: every mirror of a `key` object gets a `try_from_object`. They're here because they're key objects, not as zklogin-specific handling.

That said, if we'd rather not surface zklogin verified-proof objects, I'm happy to drop both (and their predicates). Keep or remove?

Alex6323: Dropped both, as discussed — removed the `VerifiedID`/`VerifiedIssuer` `try_from_object` constructors and the `is_verified_id`/`new_verified_id` + `VerifiedIssuer` struct-tag helpers (and their tests) in https://github.com/iotaledger/iota-rust-sdk/commit/4e075682033784b927d5d7226baca94dd0774e86. The mirror struct definitions themselves stay. The matching FFI shims are removed on #1247.

thibault-martinez (approved):

Issue #1207, opened 2026-06-12
Expose remaining Move-type mirrors in the FFI


Alex6323:
> [!NOTE]
> This issue was drafted by Claude and posted on Alex's behalf, following up on [this review thread](https://github.com/iotaledger/iota-rust-sdk/pull/1124#discussion_r3371856743).
> Updated 2026-07-22 (by Claude on Alex's behalf) after #1247 merged and #1268 opened: the tally grew to **187** mirrors with the new package-metadata types, the type-parameter strategy was settled and implemented, and the exposed set grew from 21 to 51.

`iota-sdk-move-types` contains mirrors for all **187** public Move system types. This issue accounts for every one of them: which are decodable through the FFI, which remain to be exposed, and which will never need a standalone binding (and why). The lists below are generated from the code (shape-test registry and `iota-sdk-ffi/src/move_types.rs`), not maintained by hand.

## Exposed (51)

From #1124, #1247, and #1268 (`ModuleMetadata` pending merge). Generic mirrors are either fixed `<IOTA>` instantiations (constructor validates the full type tag) or take a caller-provided `TypeTag` at runtime via `try_from_object_with_type` (marked "runtime `TypeTag`"):

| Package | Mirror | FFI binding |
| --- | --- | --- |
| `0x107a` | `alias::Alias` | `Alias` |
| `0x107a` | `alias_output::AliasOutput` | AliasOutput (`<IOTA>`) |
| `0x107a` | `basic_output::BasicOutput` | BasicOutput (`<IOTA>`) |
| `0x107a` | `expiration_unlock_condition::ExpirationUnlockCondition` | ExpirationUnlockCondition (embedded) |
| `0x107a` | `irc27::Irc27Metadata` | Irc27Metadata (embedded) |
| `0x107a` | `nft::Nft` | `Nft` |
| `0x107a` | `nft_output::NftOutput` | NftOutput (`<IOTA>`) |
| `0x107a` | `storage_deposit_return_unlock_condition::StorageDepositReturnUnlockCondition` | StorageDepositReturnUnlockCondition (embedded) |
| `0x107a` | `timelock_unlock_condition::TimelockUnlockCondition` | TimelockUnlockCondition (embedded) |
| `0x2` | `authenticator_state::AuthenticatorState` | `AuthenticatorState` |
| `0x2` | `clock::Clock` | `Clock` |
[...]

PR #1247, opened 2026-07-02
feat(bindings): add FFI shims for the new key-object mirror types

### LINKED ISSUE #1226: `try_from_object.rs` doesn't cover everything

Alex6323:
## Why

#1226 added `try_from_object` constructors to the 22 `key`-object mirror types that lacked them, but the FFI/bindings layer had no shims for those types — so non-Rust consumers still couldn't decode them. (The 16 pre-existing mirror constructors already have 1:1 FFI shims.)

## What

- Two macros — `ffi_move_object!` (non-generic) and `ffi_move_object_generic!` (single phantom type parameter) — generate the `uniffi::Object` wrapper plus its `try_from_object` / `try_from_bcs` constructors; per-type field accessors are passed in and spliced into the exported impl. This follows the crate's existing macro-generated-`#[uniffi::export]` pattern.
- Shims for all 22 new mirrors. Generic shims wrap the `<IOTA>` instantiation (the parameter is phantom, so the BCS layout is identical for any coin type) and expose `try_from_object_with_type(object, type_param: TypeTag)`, mirroring the core API so a cap over an arbitrary coin type can be decoded.

## Notes

- **Stacked on #1226** — based on `fix/1226-try-from-object-constructors` and targets it; GitHub will retarget this to `develop` once #1226 merges.
- Accessors mirror the existing shims: always `id()`, plus simple scalar/id/address fields; complex `Bag`/`VecMap` fields are either surfaced as a related object id or skipped.

## Test plan

- `cargo build -p iota-sdk-ffi`, `cargo clippy -p iota-sdk-ffi --all-targets`, `cargo +nightly fmt --check` — all clean, no lint suppressions.
- `uniffi-bindgen generate --language python` runs clean; the 22 new objects and their `try_from_object` / `try_from_object_with_type` / `try_from_bcs` constructors appear in the generated bindings.

### DIFF
--- a/crates/iota-sdk-ffi/src/macros.rs
+++ b/crates/iota-sdk-ffi/src/macros.rs
@@ -1,10 +1,13 @@
 // Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
-/// Define the FFI wrapper for a non-generic `key` Move-object mirror: the
-/// `uniffi::Object` newtype plus its `try_from_object` / `try_from_bcs`
-/// constructors. Type-specific field accessors go in the trailing block and
-/// are spliced into the exported impl.
+/// Define the FFI wrapper for a `key` Move-object mirror decoded at a fixed
+/// Move type tag (a non-generic mirror, or a fixed instantiation such as
+/// `<IOTA>`): the `uniffi::Object` newtype plus its `try_from_object` /
+/// `try_from_bcs` constructors. Type-specific field accessors go in the
+/// trailing block and are spliced into the exported impl. For mirrors whose
+/// type parameter is chosen by the caller at runtime, use
+/// [`crate::ffi_move_object_generic`].
 #[macro_export]
 macro_rules! ffi_move_object {
     (
--- a/crates/iota-sdk-ffi/src/macros.rs
+++ b/crates/iota-sdk-ffi/src/macros.rs
@@ -32,6 +32,11 @@ macro_rules! ffi_move_object {
                 Ok(::bcs::from_bytes::<$core>(&bytes)?.into())
             }
 
+            /// The object's ID.
+            pub fn id(&self) -> $crate::types::object::ObjectId {
+                (*self.0.id.object_id()).into()
+            }
+
             $($accessors)*
         }
     };
@@ -68,6 +73,11 @@ macro_rules! ffi_move_object_generic {
             #[uniffi::constructor]
             pub fn try_from_bcs(bytes: Vec<u8>) -> $crate::error::Result<Self> {
                 Ok(::bcs::from_bytes::<$core>(&bytes)?.into())
+            }
+
+            /// The object's ID.
+            pub fn id(&self) -> $crate::types::object::ObjectId {
+                (*self.0.id.object_id()).into()
             }
 
             $($accessors)*
--- a/crates/iota-sdk-ffi/src/macros.rs
+++ b/crates/iota-sdk-ffi/src/macros.rs
@@ -1,6 +1,80 @@
 // Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+/// Define the FFI wrapper for a non-generic `key` Move-object mirror: the
+/// `uniffi::Object` newtype plus its `try_from_object` / `try_from_bcs`
+/// constructors. Type-specific field accessors go in the trailing block and
+/// are spliced into the exported impl.
+#[macro_export]
+macro_rules! ffi_move_object {
+    (
+        $(#[$meta:meta])*
+        $name:ident($core:ty) { $($accessors:tt)* }
+    ) => {
+        $(#[$meta])*
+        #[derive(Debug, derive_more::From, uniffi::Object)]
+        #[uniffi::export(Debug)]
+        pub struct $name(pub $core);
+
+        #[uniffi::export]
+        impl $name {
+            /// Decode from an on-chain object, validating its Move type tag.
+            #[uniffi::constructor]
+            pub fn try_from_object(
+                object: &$crate::types::object::Object,
+            ) -> $crate::error::Result<Self> {
+                Ok(<$core>::try_from(&object.0)?.into())
+            }
+
+            /// Decode from BCS bytes, without validating the on-chain type tag.
+            #[uniffi::constructor]
+            pub fn try_from_bcs(bytes: Vec<u8>) -> $crate::error::Result<Self> {
+                Ok(::bcs::from_bytes::<$core>(&bytes)?.into())
+            }
+
+            $($accessors)*
+        }
+    };
+}
+
+/// Like [`ffi_move_object`], but for a mirror with a single (phantom) type
+/// parameter. `$core` is the type instantiated at `IOTA` (a phantom marker, so
+/// the BCS layout is the same for every coin type); the object constructor
+/// validates the on-chain type parameter against a caller-provided `TypeTag`.
+#[macro_export]
+macro_rules! ffi_move_object_generic {
+    (
+        $(#[$meta:meta])*
+        $name:ident($core:ty) { $($accessors:tt)* }
+    ) => {
+        $(#[$meta])*
+        #[derive(Debug, derive_more::From, uniffi::Object)]
+        #[uniffi::export(Debug)]
+        pub struct $name(pub $core);
+
+        #[uniffi::export]
+        impl $name {
+            /// Decode from an on-chain object, validating its Move type tag,
+            /// including that its type parameter equals `type_param`.
+            #[uniffi::constructor]
+            pub fn try_from_object_with_type(
+                object: &$crate::types::object::Object,
+                type_param: &$crate::types::move_core::TypeTag,
+            ) -> $crate::error::Result<Self> {
+                Ok(<$core>::try_from_object_with_type(&object.0, &type_param.0)?.into())
+            }
+
+            /// Decode from BCS bytes, without validating the on-chain type tag.
+            #[uniffi::constructor]
+            pub fn try_from_bcs(bytes: Vec<u8>) -> $crate::error::Result<Self> {
+                Ok(::bcs::from_bytes::<$core>(&bytes)?.into())
+            }
+
+            $($accessors)*
+        }
+    };
+}
+
 #[macro_export]
 macro_rules! export_iota_types_bcs_conversion {
     ($($name:ty),+ $(,)?) => {
--- a/crates/iota-sdk-ffi/src/move_types.rs
+++ b/crates/iota-sdk-ffi/src/move_types.rs
@@ -28,11 +28,6 @@
 //!   unlock-condition records [`TimelockUnlockCondition`],
 //!   [`ExpirationUnlockCondition`], [`StorageDepositReturnUnlockCondition`].
 //!
-//! Shims for `key`-object mirrors are macro-generated: `ffi_move_object!`
-//! for mirrors decoded at a fixed Move type tag (including the `<IOTA>`
-//! instantiations), `ffi_move_object_generic!` for mirrors whose type
-//! parameter is validated at runtime against a caller-provided `TypeTag`.
-//!
 //! Generic Move types are exposed as their `<IOTA>` instantiations
 //! (`BasicOutput<IOTA>`, `NftOutput<IOTA>`, `AliasOutput<IOTA>`,
 //! `IotaCoinMetadata` wrapping `CoinMetadata<IOTA>`, and
--- a/crates/iota-sdk-ffi/src/move_types.rs
+++ b/crates/iota-sdk-ffi/src/move_types.rs
@@ -901,10 +901,6 @@ crate::ffi_move_object! {
     /// A typed view of an on-chain `0x2::deny_list::DenyList` object (the
     /// `0x403` singleton tracking regulated-coin deny lists).
     DenyList(iota_sdk::move_types::iota_framework::deny_list::DenyList) {
-        pub fn id(&self) -> ObjectId {
-            (*self.0.id.object_id()).into()
-        }
-
         /// Object ID of the `Bag` holding the per-coin deny lists.
         pub fn lists_id(&self) -> ObjectId {
             (*self.0.lists.id.object_id()).into()
@@ -916,10 +912,6 @@ crate::ffi_move_object! {
     /// A typed view of the on-chain `0x2::random::Random` object (the `0x8`
     /// singleton holding the global randomness state).
     Random(iota_sdk::move_types::iota_framework::random::Random) {
-        pub fn id(&self) -> ObjectId {
-            (*self.0.id.object_id()).into()
-        }
-
         // The `inner` field is a `Versioned` wrapper whose state lives in a
         // dynamic field; it is not surfaced across the FFI.
     }
@@ -932,10 +924,6 @@ crate::ffi_move_object! {
     AuthenticatorState(
         iota_sdk::move_types::iota_framework::authenticator_state::AuthenticatorState
     ) {
-        pub fn id(&self) -> ObjectId {
-            (*self.0.id.object_id()).into()
-        }
-
         /// Version selecting which inner state layout the dynamic field holds.
         pub fn version(&self) -> u64 {
             self.0.version
@@ -949,10 +937,6 @@ crate::ffi_move_object! {
     PackageMetadataV1(
         iota_sdk::move_types::iota_framework::package_metadata::PackageMetadataV1
     ) {
-        pub fn id(&self) -> ObjectId {
-            (*self.0.id.object_id()).into()
-        }
-
         /// Storage ID of the package this metadata describes.
         pub fn storage_id(&self) -> ObjectId {
             self.0.storage_id.bytes.into()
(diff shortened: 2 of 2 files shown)


crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -879,3 +890,351 @@ impl StorageDepositReturnUnlockCondition {
...
+
+// --- Generic `key` mirrors, instantiated at `IOTA` ---
+
+crate::ffi_move_object_generic! {
+    /// A typed view of an on-chain `0x2::coin::TreasuryCap<T>` object, the
+    /// capability controlling a coin type's supply.
+    TreasuryCap(iota_sdk::move_types::iota_framework::coin::TreasuryCap<IOTA>) {
+        pub fn id(&self) -> ObjectId {
+            (*self.0.id.object_id()).into()
+        }
```
Thoralf-M: All the objects have the same id method, just move it into `ffi_move_object_generic!()`?

Alex6323: Done — moved `id()` into both `ffi_move_object!` and `ffi_move_object_generic!` (all the macro shims share it) and dropped the 20 per-shim copies in https://github.com/iotaledger/iota-rust-sdk/commit/267043f8ea55632590bbea05a21c4e5d9d93dd7b.

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -879,3 +890,274 @@ impl StorageDepositReturnUnlockCondition {
...
+            self.0.supply_immutable
+        }
+
+        /// Whether the metadata is considered immutable.
+        pub fn metadata_immutable(&self) -> bool {
+            self.0.metadata_immutable
+        }
+
+        // The nested `treasury_cap`, `metadata`, and `immutable_metadata`
+        // objects are not surfaced across the FFI.
```
Thoralf-M: Why not?

Alex6323: Good point — these three are concrete typed objects, not opaque containers, so I've surfaced them: `treasury_cap()`, `metadata()`, and `immutable_metadata()` (the last via a new `ImmutableCoinMetadata` shim) in https://github.com/iotaledger/iota-rust-sdk/commit/0ac91ef7563df10f4f9a42142c15bb0dc60cac03.

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -879,3 +922,293 @@ impl StorageDepositReturnUnlockCondition {
...
+        pub fn runtime_id(&self) -> ObjectId {
+            self.0.runtime_id.bytes.into()
+        }
+
+        pub fn package_version(&self) -> u64 {
+            self.0.package_version
+        }
+
+        // `modules_metadata` is a `VecMap` of per-module metadata; it is not
+        // surfaced across the FFI.
```
Thoralf-M: What about these other things? There are more comments about not surfacing something, but no proper reason

DaughterOfMars (approved):

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: Fixed in https://github.com/iotaledger/iota-rust-sdk/commit/ac7526407f68b91d1507b7e85d859ea96f319b5f. Surfaced the two that are cheap and clearly useful: `Display.fields` (as a key/value map) and `TransferPolicy.rules` (as type-name strings). For the rest I replaced the bare comments with concrete reasons:

- `Random.inner` — a `Versioned` handle whose state lives in a dynamic field, not in the object's contents, so it genuinely can't be decoded here.
- `PackageMetadataV1.modules_metadata` (`VecMap<String, ModuleMetadataV1>`) and `TokenPolicy.rules` (`VecMap<String, VecSet<TypeName>>`) — nested maps; uniffi has no tuple/map-of-struct, so each needs a dedicated record type.
- `IotaSystemStateV2` validator set + treasury — large nested structures.

Do you want those last three surfaced too? I'm happy to add the record types for the two `VecMap`s, and the validator/treasury set could go here or in a dedicated follow-up. Let me know your preference.

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -879,3 +922,305 @@ impl StorageDepositReturnUnlockCondition {
...
+            self.0.runtime_id.bytes.into()
+        }
+
+        pub fn package_version(&self) -> u64 {
+            self.0.package_version
+        }
+
+        // `modules_metadata` maps each module name to a `ModuleMetadataV1`
+        // record; surfacing it over the FFI needs a dedicated record type, so
+        // it is omitted here.
```
Thoralf-M: Can you explain a bit more why this is a reason not do add it? What's the problem with a dedicated record type?

crates/iota-sdk-ffi/src/move_types.rs
Thoralf-M: We should add all of them?

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: No good reason, really. Added the record types in https://github.com/iotaledger/iota-rust-sdk/commit/cce9e9746: `modules_metadata()` now returns a map from module name to a `ModuleMetadataV1` record, which carries the module's `AuthenticatorMetadataV1` entries.

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: Sounds good. https://github.com/iotaledger/iota-rust-sdk/commit/cce9e9746 adds the two record-type ones: `PackageMetadataV1.modules_metadata` (new `ModuleMetadataV1` and `AuthenticatorMetadataV1` records) and `TokenPolicy.rules` (a plain map from action name to rule type names turned out to be enough, no record type needed).

`Random.inner` and the `Config` settings can't be decoded from the object itself (the data lives in dynamic fields), so those stay as comments.

That leaves the `IotaSystemStateV2` validator set and treasury. It's a much bigger surface, so my preference would be a dedicated PR to keep this one reviewable (tracked in #1207), but I can pull it into this PR if you'd rather have it all in one place.

Thoralf-M: How big would the change be in line count? Estimate is fine

Alex6323: Added it to this PR after all: https://github.com/iotaledger/iota-rust-sdk/commit/7cb2d1c17 surfaces the `IotaSystemStateV2` validator set (`ValidatorSetV2`, `ValidatorV1`, `ValidatorMetadataV1`, `StakingPoolV1`), storage fund, system parameters, treasury total supply, report records, and the safe-mode scalars. With that, everything from this discussion is surfaced except data that lives in dynamic fields rather than in the objects' contents.

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -879,3 +1343,364 @@ impl StorageDepositReturnUnlockCondition {
...
+                    )
+                })
+                .collect()
+        }
+    }
+}
+
+crate::ffi_move_object_generic! {
+    /// A typed view of an on-chain `0x2::config::Config<WriteCap>` object.
+    Config(iota_sdk::move_types::iota_framework::config::Config<IOTA>) {
```
Thoralf-M: That's correct here? Comment says `Config<WriteCap>` but then there is `Config<IOTA>`

Alex6323: Yes, that's intentional, but the comment invites the confusion. `WriteCap` is the name of the phantom type parameter in the Move declaration (`public struct Config<phantom WriteCap> has key`), not a concrete type. The Rust side instantiates the mirror at `IOTA` as an arbitrary placeholder: the parameter is phantom, so the BCS layout is the same for every instantiation, and `try_from_object_with_type` validates the actual on-chain type parameter. Same convention as the other generic shims. I'll reword the doc comment so it doesn't read like a concrete type.

Thoralf-M: When reword it?

Thoralf-M (approved):

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -879,3 +1343,366 @@ impl StorageDepositReturnUnlockCondition {
...
+        pub fn version(&self) -> u64 {
+            self.0.version
+        }
+    }
+}
+
+/// Metadata of a single module of a package
+/// (`0x2::package_metadata::ModuleMetadataV1`).
+#[derive(uniffi::Record)]
+pub struct ModuleMetadataV1 {
```
thibault-martinez: Why is this here? In your `// Macro-generated shims for the Move-object mirror constructors` section

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -879,3 +1343,366 @@ impl StorageDepositReturnUnlockCondition {
...
+        Self {
+            authenticator_metadata: m.authenticator_metadata.iter().map(Into::into).collect(),
+        }
+    }
+}
+
+/// An authenticator function declared by a module
+/// (`0x2::package_metadata::AuthenticatorMetadataV1`).
+#[derive(uniffi::Record)]
+pub struct AuthenticatorMetadataV1 {
```
thibault-martinez, line 822: Same

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -879,3 +1343,366 @@ impl StorageDepositReturnUnlockCondition {
         self.0.return_amount
     }
 }
+
+// =====================================================================
+// Macro-generated shims for the Move-object mirror constructors
```
thibault-martinez: So ALL the types above are not key-object?

Alex6323: No, several of them are key objects too (`StakedIota`, `Kiosk`, `Clock`, ...). The split is hand-written shims (which predate the macros) above and macro-generated ones below, not key vs non-key. Reworded the section header in https://github.com/iotaledger/iota-rust-sdk/commit/87fc8e7d7 to make that clear.

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: It's the record type returned by the `modules_metadata()` accessor of the `PackageMetadataV1` shim right below, so it lives next to its only consumer. The section header now says the section also contains these accessor record types: https://github.com/iotaledger/iota-rust-sdk/commit/87fc8e7d7

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: Same as `ModuleMetadataV1` above: it's part of the `modules_metadata()` return type, kept next to its consumer.

crates/iota-sdk-ffi/src/move_types.rs
thibault-martinez: Ok but can't we also use the macro for the currently handwritten ones?

Alex6323: Done in https://github.com/iotaledger/iota-rust-sdk/commit/1c8beef0e: all fifteen hand-written `key`-object shims now use `ffi_move_object!` (it also covers the `<IOTA>`-instantiated ones, since the mirrors provide fixed-tag `TryFrom` impls). Not everything can convert: `IotaSystemStateV2` has no object constructor or `id` (it's a dynamic-field payload), and the embedded types (`ValidatorSetV2` and friends, `Irc27Metadata`, the unlock conditions) have no constructors at all, so those stay hand-written. Verified the generated bindings are unchanged apart from docstrings by diffing the Python bindings before and after.

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -879,3 +1020,367 @@ impl StorageDepositReturnUnlockCondition {
         self.0.return_amount
     }
 }
+
+// =====================================================================
+// Remaining `key` Move-object mirrors, plus the record types their
```
thibault-martinez: Why do we have a "remaining" section containing stuff from iota_framework, iota_system, etc... we already have sections for?

Alex6323: right

Alex6323: Fair point, that section was purely historical. Dissolved it in https://github.com/iotaledger/iota-rust-sdk/commit/42e010934: every shim now lives under its package section (0x3 / 0x2 / 0x107a) and the module doc lists the types per package. The generated bindings are byte-identical.

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -244,578 +176,1141 @@ impl IotaSystemStateV2 {
...
+    pub fn parameters(&self) -> SystemParametersV1 {
+        SystemParametersV1(self.0.parameters.clone())
     }
 
-    pub fn id(&self) -> ObjectId {
-        (*self.0.id.object_id()).into()
+    /// Validator report records: each reported validator's address mapped to
+    /// the addresses of the validators reporting it, as canonical hex
+    /// strings.
+    pub fn validator_report_records(&self) -> HashMap<String, Vec<String>> {
```
thibault-martinez: Why return strings?

crates/iota-sdk-ffi/src/macros.rs
```diff
@@ -1,6 +1,90 @@
 // Copyright (c) 2025 IOTA Stiftung
 // SPDX-License-Identifier: Apache-2.0
 
+/// Define the FFI wrapper for a non-generic `key` Move-object mirror: the
```
thibault-martinez: Maybe this isn't clear enough, because it is definitely used on some generic types

Alex6323: Right, that doc predated the conversion. Reworded in https://github.com/iotaledger/iota-rust-sdk/commit/a6f0509e3: it now says the macro covers mirrors decoded at a fixed type tag, including `<IOTA>` instantiations, and points to `ffi_move_object_generic!` for runtime type parameters.

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: The keys have to stay strings (FFI map keys cannot be objects, and Address is an object over the FFI), but the values did not: typed them in https://github.com/iotaledger/iota-rust-sdk/commit/1238d0783, so it now returns `HashMap<String, Vec<Address>>` and the doc explains the key constraint.

thibault-martinez: I'm confused, why can't objects be keys? Seems to be working fine locally

crates/iota-sdk-ffi/src/move_types.rs
```diff
@@ -8,16 +8,31 @@
...
+//!   [`PurchaseCap`].
 //! - **Stardust types** (`0x107a`): [`Nft`], [`Irc27Metadata`],
 //!   [`BasicOutput`], [`NftOutput`], [`AliasOutput`], [`Alias`], plus the
 //!   unlock-condition records [`TimelockUnlockCondition`],
 //!   [`ExpirationUnlockCondition`], [`StorageDepositReturnUnlockCondition`].
 //!
+//! Shims for `key`-object mirrors are macro-generated: `ffi_move_object!`
+//! for mirrors decoded at a fixed Move type tag (including the `<IOTA>`
+//! instantiations), `ffi_move_object_generic!` for mirrors whose type
+//! parameter is validated at runtime against a caller-provided `TypeTag`.
```
Thoralf-M: Is it really interesting for users to know which macro was used for which types?

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: No, I am confused. It doesn't have to be String at all.

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: I don't think it is important tbh.

Thoralf-M: Then let's remove it

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: You were right and my claim was wrong: object keys work because `Address` exports `Eq` and `Hash`, which is what uniffi requires for map keys. Keyed both maps by `Address` in https://github.com/iotaledger/iota-rust-sdk/commit/48ed43c2d: `validator_report_records` now returns `HashMap<Address, Vec<Address>>` and `at_risk_validators` returns `HashMap<Address, u64>`, and the incorrect doc notes are gone.

crates/iota-sdk-ffi/src/move_types.rs
Alex6323: Removed in https://github.com/iotaledger/iota-rust-sdk/commit/e608fc5e5. The per-type docs carry each constructor's contract, so nothing callers need is lost.

thibault-martinez (approved):

PR #1248, opened 2026-07-03
refactor: use fastcrypto for ed25519, secp256r1 and secp256k1 crypto

Thoralf-M:
## Why

Consolidate the SDK's crypto on `fastcrypto` (pinned `=0.1.11`, the same implementation the node uses) instead of a separate `ed25519-dalek` / `p256` / `k256` backend.

## What

- ed25519, secp256r1 and secp256k1 signing/verification now go through `fastcrypto`.
- `ed25519`, `p256` and `k256` are kept only as PKCS#8/PEM codecs behind the `pem` feature (fastcrypto has no PKCS#8 support).
- `fastcrypto`'s `wasm` feature is target-gated so `wasm32` still builds.

### Behavioral changes (now matching the node)

- **secp256r1**: low-S on sign, rejects high-S on verify (WebAuthn/passkey high-S signatures are now rejected).
- **secp256k1**: rejects malleable high-S signatures on verify (previously accepted by `k256`); signing output is unchanged (SHA-256, low-S).
- **ed25519**: ed25519-consensus verification instead of dalek `verify_strict`.

## Test plan

- `cargo test --all-features` → 30/30 (incl. ed25519/secp256r1/secp256k1 & passkey signing fixtures, the secp256k1 high-S rejection test, PEM round-trips)
- clippy `-D warnings`, `wasm32` check (ed25519 + secp256r1 + secp256k1), all feature combos, downstream crates build

Companion monorepo change: iotaledger/iota#12133

### DIFF
--- a/crates/iota-sdk-crypto/Cargo.toml
+++ b/crates/iota-sdk-crypto/Cargo.toml
@@ -29,13 +29,14 @@ default = []
 ed25519 = ["dep:fastcrypto", "dep:rand_core"]
 secp256r1 = ["dep:fastcrypto", "dep:rand_core"]
 passkey = ["secp256r1", "dep:sha2"]
-secp256k1 = ["dep:k256", "dep:rand_core", "signature/std"]
+secp256k1 = ["dep:fastcrypto", "dep:rand_core"]
 bech32 = ["dep:bech32", "signature/std"]
-# PKCS#8/PEM support. fastcrypto has no PKCS#8 codec, so ed25519-dalek and p256
-# are pulled in purely as encoders/decoders here; the actual signing and
+# PKCS#8/PEM support. fastcrypto has no PKCS#8 codec, so ed25519-dalek, p256 and
+# k256 are pulled in purely as encoders/decoders here; the actual signing and
 # verification always go through fastcrypto.
 pem = [
   "dep:ed25519-dalek",
+  "dep:k256",
   "dep:p256",
   "dep:pem-rfc7468",
   "dep:pkcs8",
@@ -60,9 +61,9 @@ bip39 = { version = "2.0.0", features = ["rand"], optional = true }
 blst = { version = "0.3.13", optional = true }
 # ed25519 PKCS#8/PEM codec (see the `pem` feature)
 ed25519-dalek = { version = "2.1.1", optional = true }
-# ed25519 and secp256r1 signing/verification backend
+# ed25519, secp256r1 and secp256k1 signing/verification backend
 fastcrypto = { version = "=0.1.11", optional = true }
-# secp256k1 support
+# secp256k1 PKCS#8/PEM codec (see the `pem` feature)
 k256 = { version = "0.13.4", default-features = false, features = ["ecdsa"], optional = true }
 # secp256r1 PKCS#8/PEM codec (see the `pem` feature)
 p256 = { version = "0.13.2", default-features = false, features = ["ecdsa", "std"], optional = true }
--- a/crates/iota-sdk-crypto/Cargo.toml
+++ b/crates/iota-sdk-crypto/Cargo.toml
@@ -26,10 +26,10 @@ rustdoc-args = [
 
 [features]
 default = []
-ed25519 = ["dep:fastcrypto", "dep:rand_core"]
-secp256r1 = ["dep:fastcrypto", "dep:rand_core"]
+ed25519 = ["dep:fastcrypto", "dep:rand_core", "dep:zeroize"]
+secp256r1 = ["dep:fastcrypto", "dep:rand_core", "dep:zeroize"]
 passkey = ["secp256r1", "dep:sha2"]
-secp256k1 = ["dep:fastcrypto", "dep:rand_core"]
+secp256k1 = ["dep:fastcrypto", "dep:rand_core", "dep:zeroize"]
 bech32 = ["dep:bech32", "signature/std"]
 # PKCS#8/PEM support. fastcrypto has no PKCS#8 codec, so ed25519-dalek, p256 and
 # k256 are pulled in purely as encoders/decoders here; the actual signing and
@@ -78,6 +78,7 @@ sha2 = { version = "0.10.8", optional = true }
 signature = "2.2"
 slip10_ed25519 = { version = "0.1.3", optional = true }
 thiserror.workspace = true
+zeroize = { version = "1.8", features = ["derive"], optional = true }
 
 # internal dependencies
 iota-types = { workspace = true, features = ["hash", "serde"] }
--- a/crates/iota-sdk-crypto/Cargo.toml
+++ b/crates/iota-sdk-crypto/Cargo.toml
@@ -31,17 +31,18 @@ secp256r1 = ["dep:fastcrypto", "dep:rand_core", "dep:zeroize"]
 passkey = ["secp256r1", "dep:sha2"]
 secp256k1 = ["dep:fastcrypto", "dep:rand_core", "dep:zeroize"]
 bech32 = ["dep:bech32", "signature/std"]
-# PKCS#8/PEM support. fastcrypto has no PKCS#8 codec, so ed25519-dalek, p256 and
+# PKCS#8/PEM support. fastcrypto has no PKCS#8 codec, so ed25519, p256 and
 # k256 are pulled in purely as encoders/decoders here; the actual signing and
 # verification always go through fastcrypto.
 pem = [
-  "dep:ed25519-dalek",
+  "dep:ed25519",
   "dep:k256",
   "dep:p256",
   "dep:pem-rfc7468",
   "dep:pkcs8",
-  "ed25519-dalek?/pem",
-  "ed25519-dalek?/pkcs8",
+  "ed25519?/pem",
+  "ed25519?/pkcs8",
+  "ed25519?/zeroize",
   "k256?/pem",
   "k256?/pkcs8",
   "p256?/pem",
@@ -60,7 +61,7 @@ bip39 = { version = "2.0.0", features = ["rand"], optional = true }
 # bls12381 support
 blst = { version = "0.3.13", optional = true }
 # ed25519 PKCS#8/PEM codec (see the `pem` feature)
-ed25519-dalek = { version = "2.1.1", optional = true }
+ed25519 = { version = "2.2", optional = true }
 # ed25519, secp256r1 and secp256k1 signing/verification backend
 fastcrypto = { version = "=0.1.11", optional = true }
 # secp256k1 PKCS#8/PEM codec (see the `pem` feature)
--- a/crates/iota-sdk-crypto/Cargo.toml
+++ b/crates/iota-sdk-crypto/Cargo.toml
@@ -26,12 +26,17 @@ rustdoc-args = [
 
 [features]
 default = []
-ed25519 = ["dep:ed25519-dalek", "dep:rand_core"]
-secp256r1 = ["dep:p256", "dep:rand_core"]
+ed25519 = ["dep:fastcrypto", "dep:rand_core"]
+secp256r1 = ["dep:fastcrypto", "dep:rand_core"]
 passkey = ["secp256r1", "dep:sha2"]
 secp256k1 = ["dep:k256", "dep:rand_core", "signature/std"]
 bech32 = ["dep:bech32", "signature/std"]
+# PKCS#8/PEM support. fastcrypto has no PKCS#8 codec, so ed25519-dalek and p256
+# are pulled in purely as encoders/decoders here; the actual signing and
+# verification always go through fastcrypto.
 pem = [
+  "dep:ed25519-dalek",
+  "dep:p256",
   "dep:pem-rfc7468",
   "dep:pkcs8",
   "ed25519-dalek?/pem",
@@ -53,11 +58,13 @@ bip32 = { version = "0.5.3", optional = true }
 bip39 = { version = "2.0.0", features = ["rand"], optional = true }
 # bls12381 support
 blst = { version = "0.3.13", optional = true }
-# ed25519 support
+# ed25519 PKCS#8/PEM codec (see the `pem` feature)
 ed25519-dalek = { version = "2.1.1", optional = true }
+# ed25519 and secp256r1 signing/verification backend
+fastcrypto = { version = "0.1.10", optional = true }
 # secp256k1 support
 k256 = { version = "0.13.4", default-features = false, features = ["ecdsa"], optional = true }
-# secp256r1 support
+# secp256r1 PKCS#8/PEM codec (see the `pem` feature)
 p256 = { version = "0.13.2", default-features = false, features = ["ecdsa", "std"], optional = true }
 # pkcs8 der and pem support
 pem-rfc7468 = { version = "0.7", optional = true, features = ["std"] }
@@ -73,6 +80,10 @@ thiserror.workspace = true
 
 # internal dependencies
 iota-types = { workspace = true, features = ["hash", "serde"] }
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+# On wasm32 fastcrypto needs getrandom's "js" backend, enabled by its "wasm" feature.
+fastcrypto = { version = "0.1.10", optional = true, features = ["wasm"] }
 
 [dev-dependencies]
 # external dependencies
--- a/crates/iota-sdk-crypto/src/ed25519.rs
+++ b/crates/iota-sdk-crypto/src/ed25519.rs
@@ -89,7 +89,7 @@ impl Ed25519PrivateKey {
     pub fn to_der(&self) -> Result<Vec<u8>, SignatureError> {
         use ed25519_dalek::pkcs8::EncodePrivateKey;
 
-        self.as_dalek()
+        self.to_dalek()
             .to_pkcs8_der()
             .map_err(SignatureError::from_source)
             .map(|der| der.as_bytes().to_owned())
@@ -110,7 +110,7 @@ impl Ed25519PrivateKey {
     pub fn to_pem(&self) -> Result<String, SignatureError> {
         use pkcs8::EncodePrivateKey;
 
-        self.as_dalek()
+        self.to_dalek()
             .to_pkcs8_pem(pkcs8::LineEnding::default())
             .map_err(SignatureError::from_source)
             .map(|pem| (*pem).to_owned())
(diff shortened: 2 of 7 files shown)


DaughterOfMars:
I don't like this because it's mysten maintained and they are slow to update :c it has dragged down so many of our deps in the past

crates/iota-sdk-crypto/src/ed25519.rs
```diff
@@ -99,15 +110,23 @@ impl Ed25519PrivateKey {
...
     pub(crate) fn from_dalek(private_key: ed25519_dalek::SigningKey) -> Self {
-        Self(private_key)
+        Self::new(private_key.to_bytes())
+    }
+
+    /// Re-expose the raw private key through ed25519-dalek, used only as a
+    /// PKCS#8/PEM codec. Don't sign with the returned key — use this type's
+    /// `Signer` impl (`try_sign`) instead, so signing goes through fastcrypto.
+    #[cfg(feature = "pem")]
+    fn as_dalek(&self) -> ed25519_dalek::SigningKey {
```
thibault-martinez: Convention is that `as_` returns a reference and `to_` returns an owned

Thoralf-M: Fixed in https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/55957b498ab539709f0eb49dcc5f91c100ce8b2b

Thoralf-M:
> I don't like this because it's mysten maintained and they are slow to update :c it has dragged down so many of our deps in the past

Problem is that there are some slight differences between the verification rules the crates have and this should not be different than in the node

crates/iota-sdk-crypto/src/ed25519.rs
```diff
@@ -2,14 +2,21 @@
...
+};
 use iota_types::{
     Ed25519PublicKey, Ed25519Signature, SignatureScheme, SimpleSignature, UserSignature,
 };
 
 use crate::{SignatureError, Signer, Verifier};
 
 #[derive(Clone, Eq, PartialEq)]
-pub struct Ed25519PrivateKey(ed25519_dalek::SigningKey);
+pub struct Ed25519PrivateKey([u8; Self::LENGTH]);
```
thibault-martinez, line 19: Just curious, why not wrap the fastcrypto type?

Thoralf-M: Would result in more conversions to bytes first for the methods on it

thibault-martinez: Not asking you to change, but the fastcrypto type is AsRef so I don't think any conversions would be needed

crates/iota-sdk-crypto/src/ed25519.rs
```diff
@@ -37,16 +44,20 @@ impl Ed25519PrivateKey {
...
+        Self(bytes)
     }
 
     pub fn scheme(&self) -> SignatureScheme {
         SignatureScheme::Ed25519
     }
 
+    /// Reconstruct the fastcrypto keypair, which performs all signing.
+    fn keypair(&self) -> Ed25519KeyPair {
+        Ed25519KeyPair::from_bytes(&self.0).expect("any 32 bytes are a valid ed25519 seed")
```
thibault-martinez: The expect message feels a bit off, it's not "any 32 bytes" here, it's an already constructed/validated private key

crates/iota-sdk-crypto/Cargo.toml
```diff
@@ -53,11 +59,13 @@ bip32 = { version = "0.5.3", optional = true }
 bip39 = { version = "2.0.0", features = ["rand"], optional = true }
 # bls12381 support
 blst = { version = "0.3.13", optional = true }
-# ed25519 support
+# ed25519 PKCS#8/PEM codec (see the `pem` feature)
 ed25519-dalek = { version = "2.1.1", optional = true }
```
semenov-vladyslav: `ed25519-dalek` crate is only used for `pkcs8` PEM format support. But it is just a reexport of `ed25519::pkcs8`. Thus I'd recommend to use `ed25519` crate instead.

crates/iota-sdk-crypto/src/ed25519.rs
```diff
@@ -206,24 +229,39 @@ impl Signer<UserSignature> for Ed25519PrivateKey {
     }
 }
 
-#[derive(Clone, Debug, Default, Eq, PartialEq)]
-pub struct Ed25519VerifyingKey(ed25519_dalek::VerifyingKey);
+#[derive(Clone, Eq, PartialEq)]
+pub struct Ed25519VerifyingKey(FcEd25519PublicKey);
```
semenov-vladyslav, line 247: Why do you need a verifying key when you already have a public key? These are the same and the types Ed25519VerifyingKey and Ed25519PublicKey should be the same.

crates/iota-sdk-crypto/src/ed25519.rs
thibault-martinez: Also looks like we need conversion now through the `keypair` methods for signing and stuff?

thibault-martinez: But the fastcrypto type is a bit more heavy, I guess it's pros and cons

semenov-vladyslav: It makes more sense to me to wrap the impl type rather than raw bytes. At least you don't have to construct the impl type value when you need to use the key.

crates/iota-sdk-crypto/src/secp256k1.rs
```diff
@@ -256,21 +302,38 @@ impl Secp256k1VerifyingKey {
...
+        )
+    }
+
+    /// Re-expose the public key through k256, used only as a PKCS#8/PEM codec.
+    /// Don't verify with the returned key — use this type's `Verifier` impl
+    /// instead, so verification goes through fastcrypto.
+    #[cfg(feature = "pem")]
+    fn to_k256(&self) -> Result<k256::ecdsa::VerifyingKey, SignatureError> {
+        k256::ecdsa::VerifyingKey::from_sec1_bytes(self.0.as_ref())
+            .map_err(SignatureError::from_source)
```
semenov-vladyslav: Shouldn't it be `expect` here too? The self value should represent a valid key anyway.

crates/iota-sdk-crypto/src/secp256r1.rs
```diff
@@ -2,19 +2,22 @@
...
-    elliptic_curve::group::GroupEncoding,
-};
 use signature::{Signer, Verifier};
 
 use crate::SignatureError;
 
-#[derive(Clone, Eq, PartialEq)]
-pub struct Secp256r1PrivateKey(SigningKey);
+#[derive(Clone, Eq, PartialEq, zeroize::Zeroize, zeroize::ZeroizeOnDrop)]
+pub struct Secp256r1PrivateKey([u8; Self::LENGTH]);
```
semenov-vladyslav, line 20: I'd keep the old impl. But it's a matter of taste I guess.

semenov-vladyslav (changes requested):
I'd try to use ed25519 crate for pkcs8 operations instead of ed25519-dalek. Other comments are not mandatory.

crates/iota-sdk-crypto/src/ed25519.rs
Thoralf-M: The fastcrypto type also doesn't implement Clone/PartialEq/Eq

thibault-martinez: Maybe it's on purpose, you want the least amount of interactions possible with Private materials

crates/iota-sdk-crypto/Cargo.toml
Thoralf-M: Done in https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/ac3abe1af81bc1acb13e78daade8a039d68fc34c

crates/iota-sdk-crypto/src/secp256k1.rs
Thoralf-M: Done in https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/ac3abe1af81bc1acb13e78daade8a039d68fc34c

crates/iota-sdk-crypto/src/secp256r1.rs
Thoralf-M: But the old impl was using the other crate you didn't want?

crates/iota-sdk-crypto/src/ed25519.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/ac3abe1af81bc1acb13e78daade8a039d68fc34c better?

crates/iota-sdk-crypto/src/ed25519.rs
Thoralf-M: One is in iota-types without crypto dependencies and therefore without verification method, the other one is with crypto dependencies

crates/iota-sdk-crypto/src/ed25519.rs
Thoralf-M: Sure, but here in the SDK they added it, probably also on purpose :trollface:

semenov-vladyslav (approved):

crates/iota-sdk-crypto/src/secp256r1.rs
```diff
@@ -42,20 +45,32 @@ impl Secp256r1PrivateKey {
     pub const LENGTH: usize = 32;
 
     pub fn new(bytes: [u8; Self::LENGTH]) -> Self {
-        Self(SigningKey::from_bytes(&bytes.into()).unwrap())
+        // Validate that the bytes form a well-formed secp256r1 private key.
```
thibault-martinez:
```suggestion
        // Validate that the bytes represent a well-formed secp256r1 private key.
```
Reads a bit weird otherwise

crates/iota-sdk-crypto/src/secp256r1.rs
thibault-martinez: Vlad probably meant with a Struct instead of bytes, not that specific struct

crates/iota-sdk-crypto/src/secp256r1.rs
```diff
@@ -135,7 +158,13 @@ impl crate::ToFromBytes for Secp256r1PrivateKey {
...
         let mut arr = [0u8; Self::LENGTH];
         arr.copy_from_slice(bytes);
-        Ok(Self::new(arr))
+
+        // Reject scalars that aren't a well-formed secp256r1 private key (e.g.
+        // all-zeros or >= the curve order) rather than panicking in `new`.
+        Secp256r1KeyPair::from_bytes(&arr).map_err(|e| {
+            crate::PrivateKeyError::InvalidScheme(format!("invalid secp256r1 private key: {e}"))
+        })?;
+        Ok(Self(arr))
```
thibault-martinez: Why do we not just call `Self::new` anymore?

crates/iota-sdk-crypto/src/secp256r1.rs
```diff
@@ -209,23 +244,30 @@ impl Signer<UserSignature> for Secp256r1PrivateKey {
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
-pub struct Secp256r1VerifyingKey(VerifyingKey);
+pub struct Secp256r1VerifyingKey(FcSecp256r1PublicKey);
```
thibault-martinez, line 250: So why bytes for private key and struct for verifying key?

crates/iota-sdk-crypto/src/secp256r1.rs
```diff
@@ -256,21 +298,38 @@ impl Secp256r1VerifyingKey {
...
+        p256::ecdsa::VerifyingKey::from_sec1_bytes(self.0.as_ref())
+            .expect("validated on construction")
     }
 }
 
 impl Verifier<Secp256r1Signature> for Secp256r1VerifyingKey {
     fn verify(&self, message: &[u8], signature: &Secp256r1Signature) -> Result<(), SignatureError> {
-        let signature = p256::ecdsa::Signature::from_bytes(signature.inner().into())?;
-        self.0.verify(message, &signature)
+        // fastcrypto hashes with SHA-256 and rejects high-S signatures.
```
thibault-martinez: Not sure that's necessary

crates/iota-sdk-crypto/src/secp256r1.rs
Thoralf-M: Done in https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/43dbb15e0017443443dc7a99e626612dee10f25a (applied to secp256k1 too for parity).

crates/iota-sdk-crypto/src/secp256r1.rs
Thoralf-M: Removed in https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/43dbb15e0017443443dc7a99e626612dee10f25a (also dropped the matching comment in the secp256k1 verify impl).

crates/iota-sdk-crypto/src/secp256r1.rs
Thoralf-M: `new` panics on invalid bytes, but `from_bytes` must return an error instead (covered by the `from_bytes_rejects_invalid_scalar` test), so it validates via the fallible `Secp256r1KeyPair::from_bytes` path rather than calling `Self::new`.

crates/iota-sdk-crypto/src/secp256r1.rs
Thoralf-M: So same discussion as https://github.com/iotaledger/iota-rust-sdk/pull/1248/changes#r3530128221 ?

crates/iota-sdk-crypto/src/secp256r1.rs
Thoralf-M: Here the type derives `Clone, Debug, Eq, PartialEq`

crates/iota-sdk-crypto/src/secp256r1.rs
thibault-martinez: `Secp256k1PrivateKey::new` returns an error, why not `Secp256r1PrivateKey::new`?

crates/iota-sdk-crypto/src/ed25519.rs
```diff
@@ -37,16 +44,20 @@ impl Ed25519PrivateKey {
...
+        Self(bytes)
     }
 
     pub fn scheme(&self) -> SignatureScheme {
         SignatureScheme::Ed25519
     }
 
+    /// Reconstruct the fastcrypto keypair, which performs all signing.
+    fn keypair(&self) -> Ed25519KeyPair {
+        Ed25519KeyPair::from_bytes(&self.0).expect("validated Ed25519KeyPair on construction")
```
thibault-martinez:
```suggestion
        Ed25519KeyPair::from_bytes(&self.0).expect("validated on construction")
```
Consistency with the 2 others (or add the name in the 2 others if you prefer)

crates/iota-sdk-crypto/src/secp256k1.rs
```diff
@@ -44,27 +47,48 @@ impl Secp256k1PrivateKey {
...
     }
 
     pub fn generate<R>(mut rng: R) -> Self
     where
         R: rand_core::RngCore + rand_core::CryptoRng,
     {
-        Self(SigningKey::random(&mut rng))
+        // Almost every 32-byte value is a valid secp256k1 private key, but a
+        // few (zero, or values at/above the curve order) are not. Draw fresh
+        // bytes until we get a valid key; in practice this never repeats.
```
thibault-martinez, line 84: And that's not an issue with 256r1?

crates/iota-sdk-crypto/src/secp256k1.rs
```diff
@@ -186,8 +219,14 @@ impl crate::FromMnemonic for Secp256k1PrivateKey {
 
 impl Signer<Secp256k1Signature> for Secp256k1PrivateKey {
     fn try_sign(&self, message: &[u8]) -> Result<Secp256k1Signature, SignatureError> {
-        let signature: k256::ecdsa::Signature = self.0.try_sign(message)?;
-        Ok(Secp256k1Signature::new(signature.to_bytes().into()))
+        // fastcrypto signs with SHA-256 and normalizes the signature to low-S.
```
thibault-martinez: not needed

crates/iota-sdk-crypto/src/secp256r1.rs
Thoralf-M: Maybe because with the other implementation it couldn't fail, but I can change it now

Thoralf-M: Done in https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/731b7c9773172ae61a4162eb6a1de26a88856689 — `new` now returns `Result` like secp256k1 and `from_bytes` calls it.

crates/iota-sdk-crypto/src/secp256k1.rs
Thoralf-M: Good catch — it was. Fixed in https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/731b7c9773172ae61a4162eb6a1de26a88856689: secp256r1 `generate` now retries until it draws a valid key, same as secp256k1.

crates/iota-sdk-crypto/src/ed25519.rs
Thoralf-M: Done in https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/731b7c9773172ae61a4162eb6a1de26a88856689.

crates/iota-sdk-crypto/src/secp256k1.rs
Thoralf-M: Removed in https://github.com/iotaledger/iota-rust-sdk/pull/1248/commits/07fbe0b7f41d2f3b347d76c44646f5e885e47006 (also dropped it from secp256r1).

thibault-martinez (approved):
I am not completely sold on the bytes instead of structs but I guess we can change it later without breaking change if we needed

crates/iota-sdk-crypto/src/secp256r1.rs
thibault-martinez: Yes

Thoralf-M: Here we have at least ZeroizeOnDrop, if we go without Clone and people then do `.as_ref()` to get the bytes it's a bit more annoying to use and when they clone them they won't get zeroize automatically
That's my argument against changing it, but not a strong one, I can also change

thibault-martinez: It's fine, let's go ahead and merge this

DaughterOfMars (approved):

PR #1252, opened 2026-07-06
chore: fix flaky coins_stream test

Thoralf-M:
## Why

`api::coins::tests::test_coins_stream` fails intermittently in CI on `assert!(num_coins >= NUM_COINS_FROM_FAUCET)`.

The root cause is test synchronization, not an indexer/client bug:

- The test funded an address with `request_and_wait`, which only confirms the faucet transfer landed **on the node** — not that the coins were **indexed** for GraphQL queries. `coins_stream` queries the indexer, which ingests checkpoints asynchronously, so it could return a *successful but incomplete* page (often 0 coins).
- The retry loop only retried on a stream **error**, never on a low count, and the delay lived inside the error branch. So under indexer lag all 10 attempts ran back-to-back with no delay, and `num_coins` was never reset per attempt.

## What

- Fund via `request_and_wait_for_finalized`, which waits for the funding transactions to be finalized (in an indexed checkpoint) before streaming.
- Retry on an insufficient count as well as on error, re-counting from scratch each attempt, with the delay applied between attempts.
- Include observed vs. expected count in the assert message.

## Test plan

Requires a running localnet (`make test-with-localnet`). Verified the crate + test target compile (`cargo check --tests -p iota-sdk-graphql-client`) and formatting is clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SQiHjiMiVinJBWLDiNtDjx)_

### DIFF
--- a/crates/iota-sdk-graphql-client/src/api/coins.rs
+++ b/crates/iota-sdk-graphql-client/src/api/coins.rs
@@ -155,10 +155,6 @@ mod tests {
         };
         let key = Ed25519PublicKey::generate(rand::thread_rng());
         let address = key.derive_address();
-        // Wait until the funding transactions are finalized so the coins are
-        // available in the indexer before we stream them. `request_and_wait`
-        // alone only confirms the faucet transfer on the node, not that the
-        // coins have been indexed for GraphQL queries.
         faucet
             .request_and_wait_for_finalized(address, &client)
             .await
--- a/crates/iota-sdk-graphql-client/src/api/coins.rs
+++ b/crates/iota-sdk-graphql-client/src/api/coins.rs
@@ -155,30 +155,48 @@ mod tests {
         };
         let key = Ed25519PublicKey::generate(rand::thread_rng());
         let address = key.derive_address();
-        faucet.request_and_wait(address).await.unwrap();
+        // Wait until the funding transactions are finalized so the coins are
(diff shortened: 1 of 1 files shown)


crates/iota-sdk-graphql-client/src/api/coins.rs
```diff
@@ -155,30 +155,48 @@ mod tests {
         };
         let key = Ed25519PublicKey::generate(rand::thread_rng());
         let address = key.derive_address();
-        faucet.request_and_wait(address).await.unwrap();
+        // Wait until the funding transactions are finalized so the coins are
+        // available in the indexer before we stream them. `request_and_wait`
+        // alone only confirms the faucet transfer on the node, not that the
+        // coins have been indexed for GraphQL queries.
```
thibault-martinez: I think this comment is not really necessary

thibault-martinez (approved):

crates/iota-sdk-graphql-client/src/api/coins.rs
Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1252/commits/0c942700cc8c9f9be241d7c667b51acc01da34c7

DaughterOfMars (approved):

PR #1253, opened 2026-07-07
feat(grpc): improve client API with typed per-endpoint read masks

### LINKED ISSUE #1142: Add additional grpc helper fns

DaughterOfMars:
Split out of #1147. This is the first of two PRs; the `_masked` methods land separately in a follow-up stacked on top of this one.

## Why
Reworks the gRPC client's read-mask handling and general ergonomics, keeping a **single unified method per endpoint** that takes an optional typed mask.

## What
- Typed, per-endpoint read mask types (`ObjectReadMask`, `TransactionReadMask`, `EpochReadMask`, …) — mixing a mask with the wrong endpoint is a compile error. Each query method takes `impl Into<Option<XxxReadMask>>`; pass `None` for the endpoint default.
- `impl Into<Option<..>>` for optional scalars and `impl IntoIterator` for id lists, dropping most `Some(..)`/slice boilerplate.
- New `get_coins` query + a `map_item` variant of `define_list_query!` for per-item proto→SDK conversion.
- Runnable gRPC examples, gated behind the `grpc` feature.
- Dependency housekeeping (tokio bump / feature narrowing, tls features).

## Test plan
- `cargo check`/`clippy -Dwarnings` clean for `iota-sdk-grpc-client` (+ tests) and `iota-sdk --features grpc --examples`.
- `cargo test --doc` passes.

### DIFF
--- a/crates/iota-sdk-grpc-client/src/api/execution/execute.rs
+++ b/crates/iota-sdk-grpc-client/src/api/execution/execute.rs
@@ -3,10 +3,13 @@
 
 //! High-level API for transaction execution.
 
-use iota_grpc_types::v1::{
-    signatures::{UserSignature as ProtoUserSignature, UserSignatures},
-    transaction::ExecutedTransaction,
-    transaction_execution_service::{ExecuteTransactionItem, ExecuteTransactionsRequest},
+use iota_grpc_types::{
+    read_mask_fields::TransactionReadMask,
+    v1::{
+        signatures::{UserSignature as ProtoUserSignature, UserSignatures},
+        transaction::ExecutedTransaction,
+        transaction_execution_service::{ExecuteTransactionItem, ExecuteTransactionsRequest},
+    },
 };
 use iota_types::SignedTransaction;
 
@@ -14,7 +17,7 @@ use crate::{
     Client,
     api::{
         EXECUTE_TRANSACTIONS_READ_MASK, Error, MetadataEnvelope, ProtoResult, ProtocolError,
-        ReadMask, Result, build_proto_transaction, field_mask_with_default,
+        Result, build_proto_transaction, field_mask_with_default,
     },
 };
 
@@ -31,14 +34,11 @@ impl Client {
     /// - `result.input_objects()` - Get input objects (if requested)
     /// - `result.output_objects()` - Get output objects (if requested)
     ///
-    /// # Read Mask
-    ///
-    /// The optional `read_mask` parameter controls which fields the server
-    /// returns. If `None`, uses [`EXECUTE_TRANSACTIONS_READ_MASK`] which
-    /// includes effects, events, and input/output objects.
-    ///
-    /// Use [`TransactionField`](iota_grpc_types::read_mask_fields::TransactionField)
-    /// constants with [`ReadMask::from`] for field selection.
+    /// The optional `read_mask` controls which fields the server returns; pass
+    /// `None` for the default mask [`EXECUTE_TRANSACTIONS_READ_MASK`] (effects,
+    /// events, and input/output objects). Pass a
+    /// [`TransactionField`](iota_grpc_types::read_mask_fields::TransactionField)
+    /// or any slice/array/vec of fields — conversion is automatic.
     ///
     /// # Checkpoint Inclusion
     ///
@@ -53,14 +53,11 @@ impl Client {
     /// # use iota_sdk_grpc_client::Client;
     /// # use iota_types::SignedTransaction;
     /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
-    /// let client = Client::new("http://localhost:9000")?;
+    /// let client = Client::new_localnet()?;
     ///
     /// let signed_tx: SignedTransaction = todo!();
-    ///
-    /// // Execute transaction - returns proto type
     /// let result = client.execute_transaction(signed_tx, None, None).await?;
     ///
-    /// // Lazy conversion - only deserialize what you need
     /// let effects = result.body().effects()?.effects()?;
     /// println!("Status: {:?}", effects.as_v1().status);
     ///
@@ -74,8 +71,8 @@ impl Client {
     pub async fn execute_transaction(
         &self,
         signed_transaction: SignedTransaction,
-        read_mask: Option<ReadMask<'_>>,
-        checkpoint_inclusion_timeout_ms: Option<u64>,
+        read_mask: impl Into<Option<TransactionReadMask>>,
+        checkpoint_inclusion_timeout_ms: impl Into<Option<u64>>,
     ) -> Result<MetadataEnvelope<ExecutedTransaction>> {
         self.execute_transactions(
             vec![signed_transaction],
@@ -83,11 +80,7 @@ impl Client {
             checkpoint_inclusion_timeout_ms,
         )
         .await?
-        .try_map(|results| {
-            results.into_iter().next().ok_or_else(|| {
-                Error::Protocol(ProtocolError::EmptyResponseField("transaction_results"))
-            })?
-        })
+        .try_map(extract_single_execution_result)
     }
 
     /// Execute a batch of signed transactions.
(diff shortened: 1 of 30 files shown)


crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
```diff
@@ -372,24 +400,25 @@ impl Client {
     /// # Ok(())
     /// # }
     /// ```
+    #[allow(clippy::too_many_arguments)]
```
Thoralf-M:
```suggestion
    #[expect(clippy::too_many_arguments)]
```

crates/iota-sdk/Cargo.toml
```diff
@@ -45,7 +45,12 @@ move-types = ["dep:iota-move-types"]
 # GraphQL
 graphql = ["dep:iota-graphql-client"]
 # gRPC
-grpc = ["dep:iota-grpc-client", "dep:iota-grpc-types"]
+grpc = [
+  "dep:iota-grpc-client",
+  "dep:iota-grpc-types",
+  "iota-grpc-client/tls-native-roots",
+  "iota-grpc-client/tls-ring",
```
Thoralf-M: Why these tls changes?  If someone wants webpki roots or a different crypto provider they can no longer opt out while using grpc

crates/iota-sdk-grpc-client/src/api/ledger/checkpoints.rs
```diff
@@ -272,10 +279,29 @@ impl Client {
...
+            start_sequence_number.into(),
+            end_sequence_number.into(),
+            read_mask.into(),
+            transactions_filter.into(),
+            events_filter.into(),
+        )
+        .await
+    }
+
+    async fn stream_checkpoints_unfiltered(
```
Thoralf-M: The other methods that aren't public have `_internal`, should this also be renamed here? `_unfiltered` also sounds a bit wrong since you can provide filters

DaughterOfMars: yeah definitely, I'll rename them

crates/iota-sdk/Cargo.toml
Thoralf-M: Btw https://github.com/iotaledger/iota-rust-sdk/pull/1201 adds 
```
# TLS is enabled so that the HTTPS constructors (`new_mainnet`,
# `new_testnet`, ...) work out of the box. The bundled webpki roots are used
# instead of the platform's native certificate store because the latter is
# unavailable on mobile targets (Android/iOS bindings); this also matches the
# behavior of the GraphQL client (reqwest with `rustls-tls`).
grpc = [
  "dep:iota-grpc-client",
  "dep:iota-grpc-types",
  "iota-grpc-client/tls-ring",
  "iota-grpc-client/tls-webpki-roots",
]
```
instead 😂

thibault-martinez: FWIW, I haven't completely reviewed my own PR yet...

crates/iota-sdk-grpc-client/src/api/execution/execute.rs
```diff
@@ -53,14 +53,11 @@ impl Client {
     /// # use iota_sdk_grpc_client::Client;
     /// # use iota_types::SignedTransaction;
     /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
-    /// let client = Client::new("http://localhost:9000")?;
+    /// let client = Client::new_localnet()?;
```
bingyanglin, line 57: Our original url is actually a placeholder, and the goal is that it can be easier to be replaced by the user, so maybe we don't need the `new_localnet()`?

crates/iota-sdk-grpc-types/src/read_mask_fields.rs
```diff
@@ -1,305 +1,489 @@
...
-//! assert_eq!(TransactionField::EFFECTS, "effects");
-//! assert_eq!(TransactionField::EFFECTS_BCS, "effects.bcs");
-//! assert_eq!(TransactionField::CHECKPOINT, "checkpoint");
+//! // Multiple fields.
+//! client
+//!     .get_objects([id], ObjectReadMask::from([ObjectField::REFERENCE, ObjectField::BCS]))
+//!     .await?;
 //! ```
+//!
+//! Mixing the wrong field type with the wrong endpoint is a compile error.
```
bingyanglin, line 28: One thing I dug into: I fired a few wrong masks at testnet, and the server already guards them for us:
- wrong/unknown field (e.g. `owner` on `get_objects`) → `INVALID_ARGUMENT: "invalid read_mask path: owner"`
- valid mask missing a needed field → `FIELD_MISSING` on conversion (e.g. `bcs`)

So a bad mask can't quietly return wrong data today, it fails loudly either way. Makes me wonder (genuinely asking) if the typed layer mostly overlaps with that, and whether it's worth the extra surface. If there's a case it saves that runtime doesn't, I'd love to hear it. Thanks!

crates/iota-sdk-grpc-client/src/api/execution/execute.rs
DaughterOfMars: I still think it's good to have as a default

crates/iota-sdk-grpc-types/src/read_mask_fields.rs
DaughterOfMars: It's for easier UX, not to protect against bad data

crates/iota-sdk-grpc-client/src/api/ledger/transactions.rs
```diff
@@ -33,84 +33,77 @@ impl Client {
...
-    /// Use [`TransactionField`](iota_grpc_types::read_mask_fields::TransactionField)
-    /// constants with [`ReadMask::from`] for field selection.
-    ///
-    /// The `input_objects`, `output_objects`, `balance_changes` and
-    /// `object_changes` fields (also included by wildcard masks) require the
-    /// serving node to still have the transaction's objects. If one has been
-    /// pruned, the transaction's result is a `FAILED_PRECONDITION` error
-    /// instead of a silently incomplete answer — narrow the read mask, or
-    /// fetch objects individually via
-    /// [`get_objects`](Client::get_objects) for best-effort retrieval.
```
Thoralf-M, line 73: Probably should keep this part here as it was just introduced in https://github.com/iotaledger/iota-rust-sdk/pull/1255 and this PR is not changing this behaviour

crates/iota-sdk-grpc-client/src/api/state/coins.rs
```diff
@@ -0,0 +1,168 @@
...
+        owner: Address,
+        coin_type: Option<StructTag>,
+        page_size: Option<u32>,
+        page_token: Option<prost::bytes::Bytes>,
+        read_mask: OwnedObjectReadMask,
+    ) -> GetCoinsQuery {
+        // When no specific coin type is provided, query for the generic
+        // `0x2::coin::Coin` (no type params); the server returns all
+        // `Coin<T>` objects regardless of `T`.
+        let coin_type = coin_type.map(StructTag::new_coin).unwrap_or_else(|| {
```
Thoralf-M, line 147: Passing the full `0x2::coin::Coin<0x2::iota::IOTA>` to `coin_type` in `get_coins()` produces `Coin<Coin<0x2::iota::IOTA>>` (nested coin) and then returns 0 coins
The docs for `get_coins()` should document that one only has to pass the inner `T` of `Coin<T>` or this place here should be changed so it doesn't created nested coins

crates/iota-sdk-grpc-client/src/api/ledger/transactions.rs
DaughterOfMars: Probably was a bad merge

Thoralf-M (approved):

Alex6323 (approved):

PR #1254, opened 2026-07-07
feat(grpc): add dedicated `_masked` query methods

### LINKED ISSUE #1142: Add additional grpc helper fns

DaughterOfMars:
Split out of #1147. **Stacked on #1253** — review/merge that one first; this PR's diff is only the `_masked` split.

## Why
Adds ergonomic, mask-explicit variants of the gRPC query methods so callers can pass a bare field / slice / array / vec of fields without wrapping in a `XxxReadMask`.

## What
Splits each query endpoint into two methods:
- the plain method uses the endpoint's default read mask (its read-mask parameter is dropped), and
- a new `_masked` variant takes an explicit `impl Into<XxxReadMask>`.

Internal callers (`transaction_builder_client`, examples) use the `_masked` variants where a custom mask is required.

## Test plan
- `cargo check`/`clippy -Dwarnings` clean for `iota-sdk-grpc-client` (+ tests) and `iota-sdk --features grpc --examples`.
- `cargo test --doc` passes.


Alex6323 (approved):

Alex6323:
looks good except for the conflicts of course :)

PR #1262, opened 2026-07-14
feat(iota-sdk-bcs-schema): add variant-level skip to BcsSchema derive

Thoralf-M:
## Why

Enums with reserved or deprecated wire slots (e.g. signature/public-key schemes) currently can't use the `BcsSchema` derive — they hand-write a single-line `definition` string just to omit the rejected slot, which is visually inconsistent with every derived rule and drifts out of sync with the Rust type.

## What

- Add `#[bcs_schema(skip)]` on an enum variant: it's omitted from the generated grammar but still consumes its discriminant, so the following variants keep their correct `%dNN` tags.
- Unit test covering omission + discriminant preservation.

Inert on its own — no generated `bcs-schema.abnf` changes until a type opts in. A follow-up switches the crypto scheme enums over to it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Uavu5pByHv9y9ZjrLBamya)_

### DIFF
--- a/crates/iota-sdk-bcs-schema/src/lib.rs
+++ b/crates/iota-sdk-bcs-schema/src/lib.rs
@@ -57,6 +57,7 @@ struct FieldAttrs {
 }
 
 struct VariantAttrs {
+    skip: bool,
     as_type: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


crates/iota-sdk-bcs-schema/src/lib.rs
```diff
@@ -619,4 +633,33 @@ mod tests {
...
+            r#"
+            enum Scheme {
+                Ed25519(Ed25519Signature),
+                #[bcs_schema(skip)]
+                Bls12381Reserved,
+                Passkey(PasskeyAuthenticator),
+            }
+            "#,
+        );
+        assert!(!schema.contains("Bls12381Reserved"), "{schema}");
```
thibault-martinez: It would not appear in PascalCase anyway?

thibault-martinez (approved):

crates/iota-sdk-bcs-schema/src/lib.rs
Thoralf-M: It actually would — `gen_enum` emits the variant ident verbatim as the trailing comment, so unskipped variants show up in PascalCase (e.g. existing rules render `%d00 move-call   ; MoveCall`). So an unskipped `Bls12381Reserved` would appear as `; Bls12381Reserved`, and the assertion was meaningful.

That said, leaning on the comment text made the intent fuzzy. I've replaced it with an exact-match assertion on the full generated rule, which shows the PascalCase comments and that the skipped variant contributes no line while `Passkey` keeps `%d02`.

---

thibault-martinez: Oh I forgot about the comment, but yeah exact string is better anyway!

Alex6323 (approved):

PR #1266, opened 2026-07-15
feat: add UserSignature address derivation and Hash impl

thibault-martinez:
- Add `UserSignature::derive_address()` to derive the signer's `Address` for all signature variants (simple, multisig, passkey, Move authenticator).
- Add `MoveAuthenticator::address()` exposing the authenticated object's address.
- Implement `std::hash::Hash` for `UserSignature` (via its BCS bytes) so it can be used in hash-based collections.

### DIFF
--- a/crates/iota-sdk-types/src/crypto/move_authenticator.rs
+++ b/crates/iota-sdk-types/src/crypto/move_authenticator.rs
@@ -30,6 +30,14 @@ impl MoveAuthenticator {
             Self::V1(_) => 1,
         }
     }
+
+    /// Returns the address of the object being authenticated, which acts as
+    /// the sender of the transaction.
+    pub fn address(&self) -> Address {
+        match self {
+            Self::V1(v1) => v1.address(),
+        }
+    }
 }
 
 #[cfg(feature = "serde")]
--- a/crates/iota-sdk-types/src/crypto/signature.rs
+++ b/crates/iota-sdk-types/src/crypto/signature.rs
@@ -395,6 +395,13 @@ impl UserSignature {
 }
 
 #[cfg(feature = "serde")]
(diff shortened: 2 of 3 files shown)


crates/iota-sdk-types/src/crypto/move_authenticator.rs
```diff
@@ -30,6 +30,14 @@ impl MoveAuthenticator {
             Self::V1(_) => 1,
         }
     }
+
+    /// Returns the address of the object being authenticated, which acts as
+    /// the sender of the transaction.
+    pub fn address(&self) -> Address {
```
Thoralf-M, line 36: Shouldn't we keep the enum clean in case a future authenticator version won't have an address?

thibault-martinez: Is this what vm-lang said when we introduced this?
Pretty hard to imagine an authenticator not backed by an object

Thoralf-M: No they didn't say that, just mentioning for future proofness, feel free to resolve if you think it's fine

thibault-martinez: @iotaledger/vm-language what do you think?

miker83z: Since it is a UserSignature we must always assume there's a user behind it => and thus an address. A move authenticator without an address could potentially exist but I don't think we would put it under UserSignature, i.e., if needed, we'd create a new type (if not a completely new protocol 😅)

Thoralf-M (approved):

semenov-vladyslav (approved):

PR #1269, opened 2026-07-16
chore(ci): point the drift check at the upstream commits behind the drift

Alex6323:
On a drift failure, the nightly check now also prints linked upstream commits between the pinned rev and develop HEAD that touched `published_api.txt`. Best effort: the list can include commits that only changed the function surface, which the type diff ignores.

Verified the API pipeline locally against today's drift; it points at iotaledger/iota#11616.

### DIFF
--- a/.github/workflows/move_types_drift_nightly.yml
+++ b/.github/workflows/move_types_drift_nightly.yml
@@ -63,8 +63,9 @@ jobs:
             # Best-effort pointer at the drift sources: upstream commits in
             # pinned..develop that touched published_api.txt. May include
             # commits that only changed the function surface (which the type
-            # diff above ignores).
-            pinned="$(grep --only-matching --perl-regexp 'iotaledger/iota\.git", rev = "\K[0-9a-f]{40}' crates/iota-sdk-move-types/Cargo.toml || true)"
+            # diff above ignores). The fetch step above resolved the pin via
+            # `cargo metadata` and recorded it in `.fetched-ref`.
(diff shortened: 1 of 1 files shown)


DaughterOfMars (approved):

Thoralf-M (approved):

.github/workflows/move_types_drift_nightly.yml
```diff
@@ -53,11 +53,27 @@ jobs:
...
         run: |
           if ! diff -u \
             crates/iota-sdk-move-types/src/packages_compiled/published_api.txt \
             head_published_api.txt; then
             echo
+            # Best-effort pointer at the drift sources: upstream commits in
+            # pinned..develop that touched published_api.txt. May include
+            # commits that only changed the function surface (which the type
+            # diff above ignores).
+            pinned="$(grep --only-matching --perl-regexp 'iotaledger/iota\.git", rev = "\K[0-9a-f]{40}' crates/iota-sdk-move-types/Cargo.toml || true)"
```
thibault-martinez: This is a bit brittle, changes to the Cargo.toml could break it. You're already getting the pinned rev in a stronger way (cargo metadata) in a previous step and storing it to `.fetched-ref`, can we not use it?

Alex6323: Good point, done in https://github.com/iotaledger/iota-rust-sdk/commit/a9a50f4f7: the step now reads the rev from `.fetched-ref`.

thibault-martinez (approved):

PR #1274, opened 2026-07-20
feat(gRPC): add `epoch_close_proof` field to `Epoch`

muXxer:
## Summary
- Adds an `epoch_close_proof` field to `Epoch`, exposing the certified closing checkpoint, the epoch-change transaction's effects/events, and the next epoch's start-of-epoch system-state objects.
- New `EpochCloseProof` message reuses the existing `Checkpoint`, `TransactionEffects`, `TransactionEvents`, and `Object` message shapes rather than introducing a new wire format.
- Each sub-field (`checkpoint`, `end_of_epoch_transaction_effects`, `end_of_epoch_transaction_events`, `next_epoch_system_state_objects`) has independent read-mask subtree granularity, mirroring the existing `protocol_config` pattern.
- Excluded from the default `GET_EPOCH_READ_MASK` (opt-in only, like `committee`/`bcs_system_state`).
- `epoch_close_proof` is absent (not an error) for the current, still-open epoch, or any epoch whose closing boundary hasn't been indexed yet.

## Context
The node's `epoch_info` storage keeps this proof material available forever (unlike checkpoints/transactions/objects, which get pruned), so it's the only reliable way to fetch epoch-closing proof material for arbitrarily old epochs. This unblocks a proof-of-inclusion client built purely on gRPC + these SDK types. Server-side population logic lands in a follow-up iota monorepo PR once this merges.

## Test plan
- [x] `cargo check -p iota-sdk-grpc-types`
- [x] `cargo test -p iota-sdk-grpc-types` (unit + doc tests)
- [x] Verified end-to-end against the iota monorepo's `get_epoch` gRPC endpoint via a temporary local path override (e2e tests covering absence for the open epoch, presence for a closed epoch, and independent subtree gating)

### DIFF
--- a/crates/iota-sdk-grpc-types/proto/iota/grpc/v1/epoch.proto
+++ b/crates/iota-sdk-grpc-types/proto/iota/grpc/v1/epoch.proto
@@ -10,6 +10,8 @@ import "google/protobuf/timestamp.proto";
 
 import "iota/grpc/options.proto";
 import "iota/grpc/v1/bcs.proto";
+import "iota/grpc/v1/checkpoint.proto";
+import "iota/grpc/v1/transaction.proto";
 
 // A member of a validator committee.
 message ValidatorCommitteeMember {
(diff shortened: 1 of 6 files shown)


Thoralf-M (changes requested):
fmt

Thoralf-M (approved):

thibault-martinez (approved):

PR #1278, opened 2026-07-22
feat: add a multi-recipient pay helper to the transaction builder

thibault-martinez:
Adds `pay()` and `pay_iota()` methods to `TransactionBuilder` for sending coins to multiple recipients in one transaction, mirroring the classic pay flow: merge the input coins into the first one, split all amounts off it in a single command, and transfer each split coin to its recipient (grouped so repeated recipients share one transfer command).

- Payments are passed as (recipient, amount) pairs, so recipients and amounts can't get out of sync.
- `pay_iota()` splits directly off the gas coin; `pay()` also accepts `Argument::Gas` as the sole coin for the same effect.
- Both are exposed on the FFI `TransactionBuilder` and `ClientTransactionBuilder` via a `Payment` record for the language bindings.

Includes unit tests for merging/recipient grouping and paying off the gas coin via both `pay_iota` and `pay` with `Argument::Gas`.


crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -644,6 +644,116 @@ impl<C, L> TransactionBuilder<C, L> {
...
+        } else if let [coin] = coin_args[..] {
+            coin
+        } else {
+            let primary_coin = coin_args.remove(0);
+            self.command(Command::MergeCoins(MergeCoins {
+                coin: primary_coin,
+                coins_to_merge: coin_args,
+            }));
+            primary_coin
+        };
```
DaughterOfMars: Feels like this should be a match

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -644,6 +644,116 @@ impl<C, L> TransactionBuilder<C, L> {
...
+        let amount_args = self.apply_arguments(amounts);
+        assert_eq!(
+            recipients.len(),
+            amount_args.len(),
+            "recipients and amounts length mismatch"
+        );
```
DaughterOfMars: I realize this is because of the trait impl, but it would be much more rusty to have this be a list of tuples to enforce this at compile time.

DaughterOfMars: It could probably be done with another small trait to cover this specific case

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -644,6 +644,116 @@ impl<C, L> TransactionBuilder<C, L> {
...
+        let Argument::Result(split) = self.command(Command::SplitCoins(SplitCoins {
+            coin,
+            amounts: amount_args,
+        })) else {
+            unreachable!("command() always returns a result argument");
+        };
+        // Group repeated recipients to minimize the number of transfer
+        // commands.
+        let mut transfers: Vec<(Address, Vec<Argument>)> = Vec::new();
```
DaughterOfMars, line 829: HashMap?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1278/commits/754fb8928cfaf707c4ef04dd479395aa3a4f73a8

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: Not possible, order matters. BTreeMap could theoretically work but the commands would not match the user inputs order anymore. Not sure that's actually an issue but for debugability I would keep the Vector.

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: Do you want to open an issue to address it later?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -644,6 +644,116 @@ impl<C, L> TransactionBuilder<C, L> {
...
+    ///     vec![to_address_0, to_address_1],
+    ///     [50000000000u64, 25000000000],
+    /// );
+    /// let txn = builder.finish().await?;
+    /// #   Ok(())
+    /// # }
+    /// ```
+    pub fn pay<T: PTBArgumentList, U: PTBArgumentList>(
+        &mut self,
+        coins: T,
```
Thoralf-M, line 795: How about making this optional for the IOTAs case? Or just accept the coin type optional so also for other coins it could automatically select the input coins

thibault-martinez: Would you be fine with a `pay_iota` method that calls `pay` with a Gas Argument?

Thoralf-M: So either Gas Argument or custom coins there or what?

Thoralf-M: Because then it's the same behaviour as this one here?

crates/iota-sdk-ffi/src/transaction_builder/builder.rs
```diff
@@ -239,6 +239,38 @@ impl TransactionBuilder {
...
+    /// separate input coins are needed.
+    ///
+    /// For a single recipient, consider using
+    /// `TransactionBuilder::send_coins()` or `TransactionBuilder::send_iota()`
+    /// instead.
+    pub fn pay(
+        self: Arc<Self>,
+        coins: Vec<Arc<PTBArgument>>,
+        recipients: Vec<Arc<Address>>,
+        amounts: Vec<Arc<PTBArgument>>,
```
Thoralf-M: I know the pay command had this design, but I don't like that we have two separated Vecs, any reason not to merge them so address and amount are next to each other in the same index?

thibault-martinez: Seems related to Chloé's comment? I could also temporarily accept a vector of tuple just in the FFI if you want

Thoralf-M: Yes related

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: I'm doing it actually

crates/iota-sdk-transaction-builder/src/builder/mod.rs
Thoralf-M: I thought you would still require the user to provide the gas arg, but you want to move it inside, all good

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1278/commits/c07cac3e9a9ff41598b75fb1ad518e84fcd33fa3

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -644,6 +644,149 @@ impl<C, L> TransactionBuilder<C, L> {
...
+    ///
+    /// // This is a coin of type
+    /// // 0xfce9c14e5f0c2b65787debb8145a33a4a2fc83152e8939000b862e174bc86bb8::cert::CERT
+    /// let coin =
+    ///     ObjectId::from_hex("0xe0e45ecb12ddca5f0d5192d2ee9e7f711959aa98614f9905e1e25c612ffd99a2")?;
+    ///
+    /// let mut builder = TransactionBuilder::new(from_address).with_client(client);
+    /// builder.pay(
+    ///     [coin],
+    ///     [(to_address_0, 50000000000u64), (to_address_1, 25000000000)],
```
Thoralf-M: I would change this example now so you also define the address and amount together, not addresses before tx builder and amount in the tx builder

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -644,6 +644,149 @@ impl<C, L> TransactionBuilder<C, L> {
...
+    /// # let client = TestClient;
+    /// let from_address =
+    ///     Address::from_hex("0xda1820edf693ee32b5729907b9b2ec8e64980ee8c008c17e89cfb4e5ecd72151")?;
+    /// let to_address_0 =
+    ///     Address::from_hex("0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900")?;
+    /// let to_address_1 =
+    ///     Address::from_hex("0x111ff11c96e2c9b19d2c47ab973012483b136dfbfee55f79b32ed4980e6ef11c")?;
+    ///
+    /// let mut builder = TransactionBuilder::new(from_address).with_client(client);
+    /// builder.pay_iota([(to_address_0, 50000000000u64), (to_address_1, 25000000000)]);
```
Thoralf-M: Same here

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1278/commits/59b1295320ada2a3f5641d98f6ff2676b6c6387c

crates/iota-sdk-transaction-builder/src/builder/mod.rs
thibault-martinez: https://github.com/iotaledger/iota-rust-sdk/pull/1278/commits/59b1295320ada2a3f5641d98f6ff2676b6c6387c

Thoralf-M (approved):

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: It could be a HashMap<Address, usize> which indexes into the vec

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: Much better, thank you

DaughterOfMars (approved):

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: I approved anyway but I do think this could be slightly optimized

thibault-martinez: Honestly don't think it matters for so few items but feel free to follow up with a PR

Alex6323 (approved):

DaughterOfMars:
Taking this over. Reviewed and pushed one fix; the implementation and docs otherwise look good to me.

**`pay` was not symmetric about empty inputs.** It returned early for an empty coin list, but with an empty *payment* list it still merged the coins and emitted a `SplitCoins` with no amounts — two commands and a set of registered inputs that do nothing. `ProgrammableTransactionBuilder::pay_impl` returns early in that case. Collecting the payments before applying the coins makes both empty cases add nothing; documented, and covered by `pay_without_payments_adds_nothing` (which fails without the fix).

Also added `pay` and `pay_iota` to the crate-level command overview in `lib.rs`, which enumerates every other command method.

Two things I looked at and deliberately left alone:

- Duplicate coins in `coins` collapse via `apply_arguments` dedup, so `pay([c, c], …)` skips the merge entirely. `ProgrammableTransactionBuilder::pay` would emit a degenerate `MergeCoins(Input(0), [Input(0)])` there. Divergence, but yours is the better behaviour.
- The recipient grouping is a linear scan per payment. Fine at realistic list sizes, and it keeps first-appearance ordering, which matches `pay_impl`.

This is consumed by iotaledger/iota#12341, which pins an integration branch (`dev-tools/cli-transfers-pay-deps`) carrying this plus #1289 and #1290 so it builds before the three land.

DaughterOfMars:
Two reverts merged in from #1290 and one note on what this now depends on.

- The object identity check in `fetch_input_objects` is gone (cb1e79bf4, merged as ecad020a7). It moves to the gRPC client in **#1293**, which checks each answered object id and transaction digest against the one requested in that position — one place instead of every caller that pairs by position.
- `pay`'s empty-payment early return (813d6d49a) stays; that one is this PR's own.

**Depends on #1293** for batched-read identity. Merge order is now #1289 → #1290 → #1293 → this, though #1293 is independent of the other three and can land in any order relative to them.

crates/iota-sdk-transaction-builder/src/error.rs
```diff
@@ -17,6 +17,15 @@ pub enum Error {
     Input(String),
     #[error("Gas object should be an immutable or owned object")]
     WrongGasObject,
+    #[error("gas coin {object_id} cannot also be passed to a command")]
+    GasCoinAsArgument { object_id: ObjectId },
+    #[error("transferring gas coin {transferred} would also transfer gas coin {missing}")]
```
Thoralf-M: I think this error is not really helpful this way, it's not really clear why this happens or what to do about it. 
Should be changed a bit, maybe add `Use another gas coin or provide both as params to transfer` or rewrite it in some different way

DaughterOfMars: I changed it, does that work better?

Thoralf-M: Yes, thanks

PR #1280, opened 2026-07-24
feat(transaction-builder): split transaction builder client trait

muXxer:
# Description of change

`TransactionBuilderClient` forced clients to implement dry run, execution, and wait logic even when the builder was only used to build transactions. This splits the trait so a client can implement just what `finish()` needs:

- New `TransactionBuilderResolveClient` with the read-only methods used to resolve and build a transaction (`object`, `objects`, `protocol_config`, `reference_gas_price`, `estimate_tx_budget`).
- `estimate_tx_budget` has a default impl returning `Ok(None)`, since estimation typically requires transaction simulation. A resolve-only client therefore needs nothing that touches the VM; `finish()` then requires an explicit gas budget and fails with `MissingGasBudget` otherwise.
- `TransactionBuilderClient: TransactionBuilderResolveClient` keeps the dry run/execution methods (`transaction`, `transaction_effects`, `dry_run_tx`, `execute_tx`, `wait_for_tx`).
- `TransactionBuilder::finish()` and `MoveAuthenticatorBuilder::finish()` now only require the resolve trait; `dry_run()`/`execute()` still require the full trait. Full clients keep overriding `estimate_tx_budget`, so automatic budget estimation in `finish()` is unchanged for them.
- The GraphQL, gRPC, and FFI client impls are split accordingly (no behavior change).
- Dropped the now-unneeded client bound on `TransactionBuilder<_, Publish>::package_id`, removing the duplicate no-client impl.

## Links to any relevant issues

None.

## How the change has been tested

- [x] Basic tests (linting, compilation, formatting, unit/integration tests)
- [x] Patch-specific tests (correctness, functionality coverage)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have checked that new and existing unit tests pass locally with my changes

New test `finish_requires_only_the_resolve_client` builds a transaction with a client that implements only `TransactionBuilderResolveClient` (relying on the default `estimate_tx_budget`): without an explicit gas budget `finish()` fails with `MissingGasBudget`, with one it succeeds.

### Release Notes

- `TransactionBuilderClient` into `TransactionBuilderResolveClient` (resolution/building) and `TransactionBuilderClient` (dry run/execution) so read-only clients can drive `TransactionBuilder::finish()` without implementing simulation or execution.

#### Breaking Changes Rollout

Affected Crates:

- iota-sdk-transaction-builder

Required User Actions:

- Manual implementers of `TransactionBuilderClient` must split their impl into two blocks: `object`, `objects`, `protocol_config`, `reference_gas_price`, and `type Error` move to `impl TransactionBuilderResolveClient`; the remaining methods and `type DryRunResult` stay in `impl TransactionBuilderClient`. `estimate_tx_budget` also belongs to the resolve trait but is now optional (defaults to `Ok(None)`); clients that keep automatic gas budget estimation should move their existing impl over. Callers are unaffected. The crates are pre-release (`0.0.1-alpha`), so this lands with the next alpha release.


crates/iota-sdk-transaction-builder/src/lib.rs
```diff
@@ -15,9 +15,12 @@
...
 //!
 //! The Transaction Builder can be used with or without a client implementing
-//! [TransactionBuilderClient]. When one is provided via the
+//! [TransactionBuilderResolveClient]. When one is provided via the
 //! [with_client](TransactionBuilder::with_client) method, the resulting builder
-//! will use it to find and validate provided IDs.
+//! will use it to find and validate provided IDs. Clients that also implement
+//! [TransactionBuilderClient] additionally enable
+//! [dry_run](TransactionBuilder::dry_run) and
+//! [execute](TransactionBuilder::execute).
```
Thoralf-M, line 23: These two methods are not all, wouldn't list all here as it gets too much, but maybe just add something like `and more`

DaughterOfMars:
Seems like a weird way to split these traits. I don't disagree with splitting but it seems like there should then be a read-only and a write trait

muXxer:
> Seems like a weird way to split these traits. I don't disagree with splitting but it seems like there should then be a read-only and a write trait

The motivation was that we don't have breaking changes for current users of the clients that implemented the `TransactionBuilderClient` trait. This is done by splitting it via the super trait approach. Also, it doesn't really make sense to have a write only trait, because the user wouldn't be able to build the transaction in the first place without the read trait. But the read trait itself makes sense, because the "write" functions all need to call the VM, which is the most complex thing to do in most of the clients. For example, the graphql API, or the JSON RPC in the indexer. They can all read their own database from the indexer, but they can't execute the VM themselves. They always need to forward the write requests to a fullnode, and that fullnode also needs to call it's executor.

We need the transaction builder now in the node, to implement the view function in the gRPC interface, but I don't want to carry over the executor everywhere, just because the transaction builder client trait is too big right now.

muXxer:
Do not merge before https://github.com/iotaledger/iota/pull/12411 is ready!

DaughterOfMars:
So, I did some investigation, and found some things that you might find interesting:

The `TransactionBuilderResolveClient::estimate_tx_budget` fn is implemented by both clients using a dry run / simulate. Technically, it should be implemented by default using the `TransactionBuilderClient::dry_run_tx`, but because the dry run results are an associated type each client needs to manually implement it. 

I interpret this to mean that `estimate_tx_budget` belongs in the `TransactionBuilderClient` trait, but it is used in transaction building in `finish_internal`. Obviously the purpose of this PR is to divide the concerns of these traits and insist that only the read path (`TransactionBuilderResolveClient`) is needed for building. However, that's not really true because of this.

It is not a stretch to say that a client should implement execute and dry_run in order to build transactions. After all, it MUST be able to simulate in order to estimate the gas cost. IMO, we should just leave the traits as-is.

chriscandless:
@muXxer @DaughterOfMars 

The `estimate_tx_budget` with default dummy `Ok(None)` bothers me: it immediatelly feels like "hack" - runtime check in place of type check, and `MissingGasBudger` error in full client if you do not override that method.
On the other hand, keeping the traits as-is has the opposite problem: a client that is fine with providing an explicit budget still has to implement simulation and execution just to call `finish()`. That is not what you would expect from entity called transaction **BUILDER**. So I want to propose a third option that I think addresses all the concerns raised on both sides:

- `TransactionBuilderResolveClient`: `object`, `objects`, `protocol_config` (default), `reference_gas_price` - reads only, **NO `estimate_tx_budget`**.
- `TransactionBuilderClient: TransactionBuilderResolveClient`: **`estimate_tx_budget` (required, no default)**, `transaction`, `transaction_effects`, `dry_run_tx`, `execute_tx`, `wait_for_tx`.

```rust
impl<C: TransactionBuilderResolveClient> TransactionBuilder<C> {
    /// The budget is a parameter - impossible to forget.
    pub async fn finish_with_budget(self, budget: u64) -> Result<Transaction, Error>;
}

impl<C: TransactionBuilderClient> TransactionBuilder<C> {
    /// Unchanged: auto-estimates when no budget was set.
    pub async fn finish(self) -> Result<Transaction, Error>;
    // dry_run, execute
}
```

`finish()` is just estimate + `finish_with_budget`, so there is one real implementation and a convenience wrapper on top. This gives:

- `estimate_tx_budget` stays required and moves next to `dry_run_tx`, where it belongs.
- For a resolve-only client, a missing budget becomes a compile error instead of a runtime one: there is no `finish()` to call (worth adding a compile-fail test). `Error::MissingGasBudget` then has only one trigger left - a full client returning `Ok(None)` from the estimate - and can be deleted entirely if the signature becomes `Result<u64, _>`: the gRPC impl never returns `Ok(None)` today, and for GraphQL `Ok(None)` means "the dry run produced no effects", which today shows up as this misleading error anyway.
- Existing callers don't change: `finish()` keeps its name, signature and auto-estimation on full clients. Explicit budgets also behave the same: the min-budget clamp only runs when estimating, so `finish_with_budget` does exactly what `finish()` with a preset budget does today.
- The node use case doesn't change either: the node client implements the same four resolve methods and sets the budget explicitly anyway.
- For trait implementors the migration is the same as in the current version of this PR.

semenov-vladyslav:
The split into two traits makes absolute sense: resolve client is used to construct a transaction, "full" client is used to execute it. The gray area is when client needs to build and execute a transaction without knowing certain values: gas coin, gas budget, gas price, - and thus a client first needs to construct a transaction with default/zero/empty values, then simulate the transaction, get gas price and budget, then add a gas coin and build the actual transaction with correct gas values and submit it for execution. In this context it makes sense to me to keep both `reference_gas_price` and `estimate_tx_budget` as part of `TransactionBuilderResolveClient` if a client is only used to build a transaction. Having an explicit `finish_with_budget` exposes some knobs that transaction builder is trying to hide: gas budget estimation and the fact that it should be set to zero if user just wants to simulate transaction and estimate gas budget.

chriscandless:
@semenov-vladyslav Flow you describe (build with zero gas values -> simulate -> get price/budget -> rebuild -> execute) needs simulation and execution - so that is a **full** client, and it keeps the auto-estimating `finish()` unchanged. The question is only about a client that **cannot** simulate: `estimate_tx_budget` in the resolve trait doesn't help it - estimation needs a simulation ("Estimation typically requires simulating the transaction", as the trait doc itself puts it), so it degenerates into the `Ok(None)` default and a runtime error. That is the "hack" I was talking about.

As for hiding the knobs - the budget knob is already public: `gas_budget()` is a setter anyone can call today, and there is the offline builder (`TransactionBuilder<()>`), where nothing is resolved or estimated for the user at all. The builder doesn't hide the budget (why would it?), it offers estimation as a convenience when the client is able to simulate. `finish_with_budget` adds no new knob - it moves an existing one into the signature for clients that cannot estimate.

The zero-budget simulation trick is also untouched: it lives on the `dry_run()` path, which never estimates and doesn't involve `finish_with_budget` at all.

DaughterOfMars:
What client can real but not simulate? I don't know if this is a legitimate use-case. 

Also, @chriscandless, I don't necessarily have a problem with your suggestion, but I do think it makes the usage for read-only a bit confusing, since the user still has access to the `gas_budget` fn. However, I suppose we could gate that under the execution trait so it is only settable in the finish_with_budget fn.

chriscandless:
@DaughterOfMars I agree the reason this particular boundary was picked looks suspicious - it reads like we are bending the SDK to a server need, which is dirty. E.g., we didn't split Simulate from Execute either, though some "hypothetical" client might want that - so yes, this cut is driven by one concrete consumer, fair.

But a real need usually means the concept exists - it just has to be spelled out. If we look at it from the consumer's side - a service that has to build transactions from its own knowledge, with no VM - its options today are:

- Write its own builder - fine, but now there are two implementations that can drift apart.
- Use `TransactionBuilder<()>` and fill in everything by hand - fine, but the resolving logic already exists in the SDK, and it feels sad that we can't reuse it purely because of the interface.
- Implement the full trait for real - shipping and maintaining an execution path which nobody calls.
- Implement the full trait with stubs - although ugly, still workable, I might even go this way myself.

Of these options, **the stub** option is actually the main indicator that our split is worth doing.
A stub is code the interface forces you to write, but without which everything still works perfectly for you. That is empirical proof that, for our use-case, the builder never needed those methods - the trait asks for more than the code uses. Conceptually, the builder could have been implemented as these functions: `resolve_objects(tx, db)`, `estimate_budget(tx, vm)`, `simulate(tx, vm)`, `execute(tx, chain)`, `wait_for_effects(digest, chain)`. They form three layers (db, vm, chain), which are fully independent - the single trait just welds them together. That's convenient for clients that have all three, but limiting for everyone else. The current helper just helps with too many things at once.

So I'd flip the conclusion: the server need didn't invent this boundary, it discovered it. By the same logic Simulate|Execute is probably a real seam too - we're just not cutting it until someone actually knocks on it, and the supertrait approach leaves room to do that later without breaking anyone.

crates/iota-sdk-transaction-builder/src/builder/client.rs
```diff
@@ -126,18 +152,6 @@ pub trait TransactionBuilderClient {
         digest: TransactionDigest,
     ) -> impl std::future::Future<Output = Result<Option<TransactionEffects>, Self::Error>>;
```
thibault-martinez: Should that not be part of the resolver trait?

DaughterOfMars (approved):

crates/iota-sdk-graphql-client/src/transaction_builder_client.rs
```diff
@@ -123,6 +107,33 @@ impl TransactionBuilderClient for Client {
         (*self).dry_run_tx(tx, skip_checks).await
     }
 
+    async fn estimate_tx_budget(&self, tx: &Transaction) -> Result<u64, Self::Error> {
+        let result = (*self).dry_run_tx(tx, true).await?;
+        // Effects carry the gas cost even when the dry run execution itself
+        // failed. Without effects there is nothing to read, so the
+        // server-reported error (if any) is surfaced.
+        let Some(effects) = &result.effects else {
```
muXxer: But does it make sense to return the gas costs of a failure?

crates/iota-sdk-transaction-builder/src/builder/client.rs
```diff
@@ -114,29 +117,19 @@ pub trait TransactionBuilderClient {
         std::future::ready(Ok(ProtocolConfig::default()))
     }
 
-    /// Fetch a transaction
-    fn transaction(
```
muXxer, line 118: Why did we remove that function from the API? wasn't it needed? I mean, yeah maybe it doesn't belong to the builder client. Just wanted to double check.

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -1760,31 +1761,12 @@ impl<C: TransactionBuilderClient, L> TransactionBuilder<C, L> {
...
-
-    /// Convert this builder into a transaction.
-    pub async fn finish(mut self) -> Result<Transaction, Error> {
-        self.finish_internal().await
+    /// Convert this builder into a transaction with the given gas budget,
+    /// used as-is (no estimation or minimum clamp) and overriding any budget
+    /// set via [`gas_budget`](Self::gas_budget).
+    pub async fn finish_with_budget(mut self, gas_budget: u64) -> Result<Transaction, Error> {
+        self.data.gas_budget = Some(gas_budget);
+        self.resolve_ptb(true).await
```
muXxer, line 1771: Shouldn't the clamp we apply in the finish_internal path also be here?

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -1941,19 +1962,7 @@ impl<C> TransactionBuilder<C, MoveCall> {
     }
 }
 
-impl TransactionBuilder<(), Publish> {
-    /// Get the package ID from the UpgradeCap so that it can be used for future
-    /// commands.
-    pub fn package_id(&mut self, name: impl AssignedResult) -> &mut TransactionBuilder {
```
muXxer, line 1949: Where did that came from? Why was it removed? Was it never used?

crates/iota-sdk-transaction-builder/src/builder/client.rs
```diff
@@ -56,14 +56,17 @@ pub struct ProtocolConfig {
     pub attributes: BTreeMap<String, String>,
 }
 
-/// A trait which defines methods needed from the client for the Transaction
-/// Builder.
-pub trait TransactionBuilderClient {
+/// Base trait shared by the transaction builder client traits, carrying the
+/// client's error type.
+pub trait TransactionBuilderClientBase {
```
thibault-martinez, line 61: Why do we have this exactly? To avoid having the 3 traits depending on the lower ones?

crates/iota-sdk-transaction-builder/src/builder/move_authenticator.rs
```diff
@@ -59,7 +59,7 @@ impl MoveAuthenticatorBuilder {
     /// which can be used to execute the given transaction.
     pub async fn finish(
         self,
-        client: impl TransactionBuilderClient,
+        client: impl TransactionBuilderLedgerClient,
```
muXxer, line 62: For the other "finish" we need the SimulationClient, so I don't understand why we only need the LedgerClient

thibault-martinez: A smaller set of requirements is always better?

crates/iota-sdk-transaction-builder/src/builder/client.rs
DaughterOfMars: wasn't actually used

crates/iota-sdk-transaction-builder/src/builder/move_authenticator.rs
muXxer: Yeah sure, I also agree, I just wonder how it is possible to finish without having the possibility to estimate?

thibault-martinez: We're just building a MoveAuthenticator here, not a Transaction

crates/iota-sdk-graphql-client/src/transaction_builder_client.rs
chriscandless: Removed the entire change. We might need separate PR for "errors & effects" stuff. Will post questsion in the channel.

crates/iota-sdk-transaction-builder/src/builder/mod.rs
chriscandless: I don't think it should be there. `finish_internal` does auto-estimation, whose main idea is to "set minimum possible gas which is going to work for this tx".
`finish_with_budget` does not do estimation - it executes tx with the gas gives by the user. If it fails - it fails. It's like with `gas_budget()` setter.

crates/iota-sdk-transaction-builder/src/builder/move_authenticator.rs
chriscandless: MoveAuthenticator does not have gas budget or estimation

crates/iota-sdk-transaction-builder/src/builder/mod.rs
chriscandless: It was removed by you. In the first commit of the branch.
Let me check it what could be the reason

chriscandless: Ah, I see now. Look at the last line of this change:
```
impl<C: TransactionBuilderClient> TransactionBuilder<C, Publish> {
```
was replaced with:
```
impl<C> TransactionBuilder<C, Publish> {
```

There were two byte-identical version of `package_id()` method: one for offlino-builder, one for builder with client. Your change just removed specializations and left one generic for everyone. And that totally makes sense, because `package_id()` does not even use client at all.

crates/iota-sdk-transaction-builder/src/builder/client.rs
chriscandless: Three traits implemented by same client have same error type. Without supertrait, formally there would be three separate error types, one for each trait.
And in `finish()` we use methods from two different traits - Ledger and Simulation. And we get in trouble because we need to return some error 😄

crates/iota-sdk-transaction-builder/src/lib.rs
chriscandless: The entire paragraph was rewritten - instead of listing methods of builder we now explain what methods are enabled by each trait. Please check it out

crates/iota-sdk-grpc-client/src/transaction_builder_client.rs
```diff
@@ -232,6 +202,25 @@ impl TransactionBuilderClient for Client {
             .await?
             .into_inner())
     }
+}
+
+impl TransactionBuilderExecutionClient for Client {
+    async fn transaction_effects(
```
muXxer: Can we keep the same order for the implementations?

chriscandless (approved):

crates/iota-sdk-transaction-builder/src/builder/client.rs
DaughterOfMars: But the supertrait could be the Ledger trait, since it's necessary for the others to work

thibault-martinez: > But the supertrait could be the Ledger trait, since it's necessary for the others to work

The 3 traits are independent, you don't need each other to implement them

DaughterOfMars: You don't, but the two dry run and execute could still be dependent on the ledger one, just for simplicity. You can't really use the txn builder with the others but without the ledger one.

crates/iota-sdk-grpc-client/src/transaction_builder_client.rs
```diff
@@ -202,12 +164,20 @@ impl TransactionBuilderClient for Client {
...
 
     async fn estimate_tx_budget(&self, tx: &Transaction) -> Result<Option<u64>, Self::Error> {
         // Simulate with relaxed checks and read the gas used from the resulting
         // effects.
         let simulated = self
-            .simulate_transaction(tx.clone(), true, SimulateReadMask::default())
+            .simulate_transaction(
+                tx.clone(),
+                true,
+                SimulateField::EXECUTED_TRANSACTION_EFFECTS_BCS,
```
thibault-martinez, line 179: Why did this change?

chriscandless: This is an improvement: for gas estimation not all of the simulation result fields are needed, so we request only what we need.

crates/iota-sdk-transaction-builder/src/builder/client.rs
chriscandless: Makes sense - we never use just Simulation or just Execution, always with Ledger. But is it an argument that we must join them? I don't know. I personally like both options and have no strong feeling about any of them. I can change it to supertraits, but I would leave that decision to the quorum of SDK team @iotaledger/iota-sdk 😄

thibault-martinez (approved):

crates/iota-sdk-transaction-builder/src/builder/client.rs
thibault-martinez: I think it's fine the way it is tbh, more flexible just in case

PR #1281, opened 2026-07-24
chore: make transaction data/effects tests localnet-stable

Thoralf-M:
## Why

`test_transaction_data_effects` and `test_transactions_data_effects` pinned a hardcoded testnet transaction digest. Testnet prunes transactions after 30 days, so the tests started failing once that digest aged out.

## What

- Run both tests against localnet via `test_client()`, matching the other network-backed tests in the crate.
- Fetch a live transaction digest from the network first (`transactions(...)`) instead of hardcoding one, so the tests stay stable regardless of pruning.

## Test plan

- `make test-with-localnet` (CI `tests-with-network` job) runs these against a fresh localnet.

---
_Generated by [Claude Code](https://claude.ai/code/session_014RN4txUu4NxCkwDC8uT9Rx)_

### DIFF
--- a/crates/iota-sdk/examples/get_transaction.rs
+++ b/crates/iota-sdk/examples/get_transaction.rs
@@ -7,8 +7,6 @@ use iota_sdk::graphql_client::{Client, error::Result, pagination::PaginationFilt
 async fn main() -> Result<()> {
     let client = Client::new_localnet();
 
-    // Fetch a recent transaction from the network so the example does not
-    // depend on a specific digest that may have been pruned.
     let transactions = client
         .transactions(None, PaginationFilter::default())
         .await?;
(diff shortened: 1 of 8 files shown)


crates/iota-sdk/examples/get_transaction.rs
thibault-martinez, line 1: What about the bindings example?

Thoralf-M: On the way

Thoralf-M: https://github.com/iotaledger/iota-rust-sdk/pull/1281/changes/fe8fd52f3730fd1f2c827510c05de42bb93ec49d

thibault-martinez (approved):

Alex6323 (approved):

PR #1289, opened 2026-07-28
feat(txn-builder): add `finish_kind` and `gas_refs`

DaughterOfMars:
## Why

`finish` always resolves the gas payment — it picks gas coins when none are set, estimates a budget, and fetches the gas price. Callers that only want the `TransactionKind` had no way to get one, and couldn't just throw away the gas `finish` produced: automatic selection may claim a coin the caller named, and since a claimed coin leaves `inputs`, `Argument::Input` then resolves to `Argument::Gas`. A `SplitCoins` on a specific coin silently becomes a split of the gas coin.

That blocks the IOTA CLI from building its transactions through this builder (iotaledger/iota#12099): it needs the kind first, because selecting gas needs a budget, which needs a dry run of the kind.

## What

- **`finish_kind`** — returns just the kind. Inputs are resolved with the client; no gas is selected, no budget estimated, no price fetched.
- **`gas_refs`** — sets gas coins that are already resolved to references, the counterpart of the offline builder's `gas`. As with `gas`, setting it stops automatic selection.

`resolve_ptb` is split into `select_default_gas` + `resolve_kind` so both paths share the resolution. No behaviour change there, and both additions are additive.

## Test plan

Three tests in `builder::tests::finish_kind`, plus doctests for both methods:

- `keeps_the_coin_that_gas_selection_would_claim` — pins the bug above: the same builder yields `Argument::Input(0)` via `finish_kind` and `Argument::Gas` via `finish`.
- `ignores_gas_and_transaction_metadata` — gas, sponsor, price, budget and expiration make no difference to the kind.
- `gas_refs_are_used_as_given` — the reference reaches the gas payment unmodified (no lookup) and default selection doesn't run.

`make clippy`, `cargo +nightly fmt`, `dprint fmt`, `cargo test --all-features --doc`, and `cargo nextest run --all-features` for `iota-sdk-types` / `-crypto` / `-transaction-builder` (382 passing) all clean. The two `iota-sdk-move-types` `move_shape_compare` failures on my machine reproduce on a clean `develop` — they need `make fetch-compiled-packages`.

### DIFF
--- a/crates/iota-sdk-transaction-builder/src/builder/mod.rs
+++ b/crates/iota-sdk-transaction-builder/src/builder/mod.rs
@@ -1056,12 +1056,70 @@ impl<C: TransactionBuilderClient, L> TransactionBuilder<C, L> {
         self
     }
 
-    async fn resolve_ptb(&mut self, default_gas: bool) -> Result<Transaction, Error> {
-        let mut inputs = Vec::new();
-        let mut gas = Vec::new();
-        let mut input_map = HashMap::new();
+    /// Add gas coins that will be consumed, by reference. Optional.
+    ///
+    /// Same as [`gas`](Self::gas), but for coins the caller has already
+    /// resolved: the builder takes the references as given instead of looking
+    /// the objects up. Setting either one stops the builder from picking gas
+    /// coins on its own.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use iota_sdk_transaction_builder::TestClient;
+    /// use std::str::FromStr;
+    ///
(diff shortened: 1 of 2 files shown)


Thoralf-M (approved):

DaughterOfMars:
Folded into #1278 so the CLI migration has a single PR to review and a single rev to pin. Content unchanged — `finish_kind`, `gas_refs`, and the `resolve_ptb` split into `select_default_gas` + `resolve_kind`.

DaughterOfMars:
Reopened — closing this was my mistake. It stays a separate PR; only its content rides along on #1278's branch so there is a single rev to pin.

Alex6323 (approved):

PR #1290, opened 2026-07-28
feat(txn-builder): resolve input objects in one request

DaughterOfMars:
> Stacked on #1289 — base is `dev-tools/txn-builder-finish-kind`, so review that one first. GitHub will retarget this to `develop` when #1289 merges. Only the second commit belongs to this PR.

## Why

Input resolution fetched objects one at a time, awaiting each in a loop, so a `pay` with ten input coins cost ten sequential round trips.

## What

- **`TransactionBuilderClient::objects_by_id`** — takes a slice of `(ObjectId, Option<Version>)` and answers in request order, `None` for anything missing. The **default impl loops over `object`**, so this is additive: existing implementors (GraphQL, FFI, `TestClient`) keep compiling and behave exactly as before.
- **gRPC override** using `GetObjects`, which turns input resolution into a single request.
- The builder collects the ids it needs up front and resolves them in one call.

A batch request fails as a whole when an object is missing and doesn't say which, so the gRPC override falls back to one request per object on a not-found error. That keeps the `missing object {id}` error intact, and only costs anything on the error path.

## Test plan

Three tests in `builder::tests::input_resolution`, driven by a `RecordingClient` that logs how the builder asks for objects:

- `all_inputs_are_fetched_in_one_request` — three owned inputs plus a shared one produce exactly one batch call holding all four ids, and zero single-object calls.
- `batched_objects_are_matched_to_their_inputs` — the shared object still resolves as shared and the coin as owned, i.e. results are matched back to the right input rather than by position alone.
- `a_missing_object_is_named_in_the_error` — a missing object is still reported by its own id.

The default impl keeps its coverage through every `TestClient`-based test, since `TestClient` doesn't override the new method.

`make clippy`, `cargo +nightly fmt`, `dprint fmt`, doctests, and `cargo nextest run --all-features` for `iota-sdk-types` / `-crypto` / `-transaction-builder` (385 passing) plus `--no-default-features -p iota-sdk-grpc-client` are all clean. Not covered by tests: the gRPC not-found fallback, which needs a live server.


crates/iota-sdk-transaction-builder/src/builder/client.rs
```diff
@@ -71,6 +71,25 @@ pub trait TransactionBuilderClient {
...
+    /// gRPC's `GetObjects`) should override it.
+    fn objects_by_id(
+        &self,
+        object_ids: &[(ObjectId, Option<Version>)],
+    ) -> impl std::future::Future<Output = Result<Vec<Option<Object>>, Self::Error>> {
+        async move {
+            let mut objects = Vec::with_capacity(object_ids.len());
+            for (object_id, version) in object_ids {
+                objects.push(self.object(*object_id, *version).await?);
+            }
```
Thoralf-M, line 88: Not sure if it's worth it, but we could also send the object requests in parallel here?

DaughterOfMars: I'll take a look at it :)

DaughterOfMars:
Folded into #1278 so the CLI migration has a single PR to review and a single rev to pin. Content unchanged — `objects_by_id` with its sequential default impl, the gRPC `GetObjects` override, and the object identity check.

DaughterOfMars:
Reopened — closing this was my mistake. It stays a separate PR; only its content rides along on #1278's branch so there is a single rev to pin.

DaughterOfMars:
Dropped the object identity check I added in 1955b55bb (reverted in cb1e79bf4).

It belongs in the client that made the request and knows what it asked for, not in every caller that pairs results with requests by position. **#1293** adds it to `get_objects` and `get_transactions`, so `fetch_input_objects` relies on `objects_by_id`'s documented ordering instead of policing it.

The length check went with it, for consistency — if implementors are trusted for ordering they may as well be trusted for count, and a short answer still surfaces as `missing object {id}` from the map lookup, just with a less precise message.

**Depends on #1293** for the guarantee. Worth noting this PR predates #1292, so the `objects_by_id` gRPC override still carries a not-found fallback that per-item results make unnecessary; that comes out when the rev moves past it.

crates/iota-sdk-transaction-builder/src/builder/client.rs
DaughterOfMars: So it's not that easy because this is the default trait impl, and we can't convert the tokio join error that could arise from joining the tasks. We could add it in each impl but not here :c

Thoralf-M: It's fine for now, thanks for checking

crates/iota-sdk-transaction-builder/src/builder/mod.rs
```diff
@@ -1681,6 +1717,222 @@ mod tests {
         assert!(matches!(split.amounts[1], Argument::Input(0)));
     }
 
+    /// A [`TestClient`] that records how the builder asked for objects, and can
+    /// report chosen ids as missing.
+    #[cfg(feature = "test-client")]
+    #[derive(Clone, Default)]
+    struct RecordingClient {
```
Thoralf-M: Can we put this test with the RecordingClient in an own file maybe? Gets quite noisy here

crates/iota-sdk-grpc-client/src/transaction_builder_client.rs
```diff
@@ -60,6 +60,36 @@ impl TransactionBuilderClient for Client {
         }
     }
 
+    async fn objects_by_id(
+        &self,
+        object_ids: &[(ObjectId, Option<Version>)],
+    ) -> Result<Vec<Option<Object>>, Self::Error> {
+        if object_ids.is_empty() {
+            return Ok(Vec::new());
+        }
```
Alex6323, line 79: is there are good argument for implementing it like this? Haven't checked it in detail but the call site could just not call this function if it knows it doesn't need to.
I guess it's for ergonomics but still it's a useless call and the caller knows it already.

DaughterOfMars: Technically the caller could call with an unknown slice, so it's not useless IMO

crates/iota-sdk-transaction-builder/src/builder/mod.rs
DaughterOfMars: I moved the client def, is that enough?

Thoralf-M: Yes a bit better there

Thoralf-M (approved):

Alex6323 (approved):

PR #1291, opened 2026-07-29
feat(grpc): view_function_call

semenov-vladyslav:
# Description

This PR adds support for a new gRPC execution service endpoint to call move functions annotated with `#[view]` attribute.


crates/iota-sdk-grpc-types/proto/iota/grpc/v1/transaction_execution_service.proto
```diff
@@ -143,3 +145,43 @@ message SimulateTransactionsResponse {
   // Results for each transaction in the request, in the same order.
   repeated SimulateTransactionResult transaction_results = 1;
 }
+
+message ViewFunctionCallArgument {
+  option (iota.grpc.message_accessors) = "with";
+
+  optional iota.grpc.v1.bcs.BcsData bcs = 3;
+  optional google.protobuf.Value json = 4;
```
muXxer: Why not start with "1"?

crates/iota-sdk-grpc-types/proto/iota/grpc/v1/transaction_execution_service.proto
```diff
@@ -143,3 +145,43 @@ message SimulateTransactionsResponse {
...
+  repeated ViewFunctionCallItem function_calls = 1;
+
+  // Mask specifying which fields to read, applied to each ViewedFunctionCall in the response.
+  optional google.protobuf.FieldMask read_mask = 2;
+}
+
+message ViewFunctionCallResult {
+  option (iota.grpc.message_accessors) = "with";
+
+  iota.grpc.v1.command.CommandOutputs view_function_call_result = 1;
```
muXxer, line 179: Can the function calls fails? Can they fail one by one? If yes then we report errors back in the same order as the request items.
```suggestion  
  oneof result {
    iota.grpc.v1.command.CommandOutputs view_function_call_result = 1;
    google.rpc.Status error = 2;
  }
```

crates/iota-sdk-grpc-types/proto/iota/grpc/v1/transaction_execution_service.proto
semenov-vladyslav: Sure, just a typo.

crates/iota-sdk-grpc-types/proto/iota/grpc/v1/transaction_execution_service.proto
semenov-vladyslav: No, I don't think so, the results are reported from executor in `SimulateTransactionResult::execution_result: Result<Vec<ExecutionResult>, ExecutionError>`. Do you think we should call each view function in a separate transaction and simulation?

semenov-vladyslav: Changed, now each view function call runs in a separate simulate transaction, thus status error is returned for each failed call.

jkrvivian (approved):

crates/iota-sdk-grpc-client/src/api/execution/view.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub async fn view_function_calls(
+        &self,
+        function_calls: Vec<ViewFunctionCallItem>,
+        read_mask: impl IntoReadMask<ViewFunctionCallReadMask>,
+    ) -> Result<MetadataEnvelope<Vec<Result<ViewFunctionCallOutputs>>>> {
+        if function_calls.is_empty() {
+            return Err(Error::EmptyRequest);
+        }
+
+        let read_mask = read_mask.into_read_mask();
```
bingyanglin: `ViewFunctionCallOutputs::path_builder().return_values().json()` produces `return_values.json`. The node accepts this during read-mask validation because `return_values` is a field of `ViewFunctionCallOutputs`, but response construction only looks under the `execution_result` subtree. This leaves the result variant unset and the client reports `ProtoConversion("missing field: result")`. Maybe we reject masks without an `execution_result` prefix (or `*`) before sending for better UX?

crates/iota-sdk/examples/move_view_call_grpc.rs
```diff
@@ -0,0 +1,78 @@
...
+use serde_json::json;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let client = Client::new_testnet()?;
+
+    // A single call, with arguments as JSON.
+    let outputs = client
+        .view_function_call(
+            "0x2::hash::blake2b256",
```
bingyanglin: `0x2::hash::blake2b256` is not marked `#[view]`, so the node rejects this call on every network. Maybe in the example we use a real `#[view]` function or clearly require a user-published package?

bingyanglin (commented):
Others LGTM.

crates/iota-sdk/examples/move_view_call_grpc.rs
bingyanglin: Fixed at [05addb1](https://github.com/iotaledger/iota-rust-sdk/pull/1291/commits/05addb1eb322ea505379a555d247b302e8e1fe61).

bingyanglin (approved):

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/transaction_execution_service.rs
```diff
@@ -176,3 +176,77 @@ impl SimulateTransactionResult {
         }
     }
 }
+
+// ViewFunctionCallOutputs
```
Thoralf-M, line 180: Can be deleted? Or is this for generated code relevant?

Makefile
```diff
@@ -432,7 +432,8 @@ example:
 examples: ## Run all Rust examples
 	@# NOTE: -maxdepth 1 -type f excludes package-based examples like polling-indexer
 	@# that require external services (e.g. PostgreSQL). Run those separately.
-	@for example in $$(find crates/iota-sdk/examples -maxdepth 1 -type f -name "*.rs" -exec basename {} .rs \;); do \
+	@# TODO(#1363): Re-enable Move View call over gRPC examples
```
Thoralf-M, line 435: It's already enabled on testnet by now

thibault-martinez: over gRPC?

Thoralf-M: Ah no, thought it was just about the move view feature itself

crates/iota-sdk-grpc-types/src/proto/mod.rs
```diff
@@ -196,3 +196,99 @@ pub fn prost_to_json(value: &prost_types::Value) -> serde_json::Value {
...
+///
+/// `google.protobuf.Value` carries every number as an `f64`, which cannot hold
+/// every integer exactly. All numeric values go as strings for the receiver to
+/// parse against the type it expects. This applies at every depth, so a number
+/// nested in an array or object becomes a string too.
+///
+/// Numeric values not representable by `serde_json::Number`, eg.
+/// a Move `u128`/`u256` argument, must be passed as
+/// `serde_json::Value::String` directly.
+pub fn json_to_prost(value: &serde_json::Value) -> prost_types::Value {
```
Thoralf-M: Maybe another name would be better here like `json_to_prost_stringify_numbers` as it's rather specific to that and not directly the reverse of `prost_to_json`

crates/iota-sdk-grpc-client/src/api/execution/view.rs
```diff
@@ -0,0 +1,160 @@
...
+                            .collect(),
+                    ),
+            ],
+            read_mask,
+        )
+        .await?
+        .try_map(|results| {
+            results.into_iter().next().ok_or_else(|| {
+                Error::Protocol(ProtocolError::EmptyResponseField(
+                    "view_function_call_results",
```
Thoralf-M: Why return this in the error if this field doesn't exist? Should it be `transaction_results`?

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/transaction_execution_service.rs
```diff
@@ -176,3 +176,77 @@ impl SimulateTransactionResult {
...
+
+impl ViewFunctionCallOutputs {
+    /// Get the execution result (return values on success, execution error on
+    /// failure).
+    ///
+    /// **Read mask:** `"execution_result"` (see
+    /// [`VIEW_FUNCTION_CALL_OUTPUTS_EXECUTION_RESULT`]).
+    ///
+    /// [`VIEW_FUNCTION_CALL_OUTPUTS_EXECUTION_RESULT`]: crate::read_masks::VIEW_FUNCTION_CALL_OUTPUTS_EXECUTION_RESULT
```
thibault-martinez, line 189: Here and below, why is there a read mask section?

crates/iota-sdk-grpc-client/src/api/execution/view.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        let read_mask = read_mask.into_read_mask();
+        let request = ViewFunctionCallsRequest::default()
+            .with_view_function_calls(function_calls)
+            .with_read_mask(read_mask);
+
+        let response = self
+            .execution_service_client()
+            .view_function_calls(request)
+            .await?;
```
Thoralf-M, line 161: Should we also check that the length of responses match the request?
`check_result_count()` is already defined and used in other places

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/transaction_execution_service.rs
bingyanglin: Not by codegen, it's just a section marker that the same as others in the files, for consistency.

crates/iota-sdk-grpc-types/src/proto/mod.rs
bingyanglin: Renamed to json_to_prost_stringify_numbers in 925f381ae. Thanks.

crates/iota-sdk-grpc-client/src/api/execution/view.rs
bingyanglin: The response field is `call_results`. Fixed it in 6b02af500. Thanks.

crates/iota-sdk-grpc-types/src/proto/iota/grpc/v1/transaction_execution_service.rs
bingyanglin: Good questions, The accessor returns nothing unless the mask covered the field, so each accessor documents the path that makes it readable. 
These mirror the existing `SimulatedTransaction` accessors above them for consistency.

crates/iota-sdk-grpc-client/src/api/execution/view.rs
bingyanglin: Agreed. Added in c8f6d5ba4.

thibault-martinez (approved):

Thoralf-M (approved):

PR #1292, opened 2026-07-30
fix(grpc-client)!: surface per-item errors in `get_objects`/`get_transactions`

muXxer:
## Why

`get_objects` and `get_transactions` collapsed the server's per-item results with `collect::<Result<Vec<_>>>()`, so the first per-item error discarded every item already collected — and the caller could not tell which request failed.

A missing item is routine (coins get consumed or smashed, objects get pruned by the serving node), so one of them made the whole batch unusable. Callers were working around this by refetching one item at a time whenever a batch failed.

The node was always right here: `ObjectResult`/`TransactionResult` carry a per-item `google.rpc.Status` and the stream does not abort. Only the client threw that away.

## What

- `get_objects`/`get_transactions` now return `Vec<Result<T>>` — one result per request, in request order. This matches `execute_transactions`/`simulate_transactions`, which were already per-item.
- Adds `Error::is_not_found()`, replacing four hand-rolled copies of the same status-code check.

Callers that need every item can `collect::<Result<Vec<_>>>()`; callers that tolerate gaps inspect each slot.

**Breaking, deliberately.** The old signature cannot express "this item is missing, the rest are fine" — that *is* the bug — so a deprecation step would mean shipping a parallel method whose only purpose is to be correct.

## Test plan

- New unit tests for the per-item mapping and `Error::is_not_found()`.
- `cargo test -p iota-sdk-grpc-client` (7 lib + 22 doctests), clippy `-D warnings`, nightly fmt — all clean.
- Verified against a real node in the monorepo: 105/105 gRPC e2e tests pass. Their previous assertions encoded the old behaviour (`"Mixed valid/invalid should return an error"`).

Downstream: the monorepo needs a follow-up PR once the pinned `rev` is bumped — five call sites, mostly mechanical, plus removal of the one-at-a-time fallback in `iota-vm-sdk`.

### DIFF
--- a/crates/iota-sdk-grpc-client/src/api/common.rs
+++ b/crates/iota-sdk-grpc-client/src/api/common.rs
@@ -133,6 +133,11 @@ pub enum ProtocolError {
     /// Error during checkpoint data stream reassembly.
     #[error("checkpoint stream error: {0}")]
     CheckpointStream(#[from] CheckpointStreamError),
+
+    /// A batched read returned a number of results that does not match the
+    /// number of requested items.
+    #[error("expected {expected} results, got {actual}")]
+    UnexpectedResultCount { expected: usize, actual: usize },
 }
 
 /// Errors during checkpoint data stream reassembly.
@@ -265,6 +270,21 @@ pub trait ProtoResult {
 /// while one that tolerates gaps can inspect each slot.
 pub fn into_item_results<T: ProtoResult>(batch: Vec<T>) -> Vec<Result<T::Value>> {
     batch.into_iter().map(ProtoResult::into_result).collect()
+}
+
+/// Check that a batched read answered every requested item.
+///
+/// Callers pair results with requests by position, so a count that does not
+/// match the request leaves no way to tell which item each result belongs to.
+pub fn check_result_count<T>(results: &[T], expected: usize) -> Result<()> {
+    if results.len() == expected {
+        Ok(())
+    } else {
+        Err(Error::Protocol(ProtocolError::UnexpectedResultCount {
+            expected,
+            actual: results.len(),
+        }))
+    }
 }
 
 impl ProtoResult for ObjectResult {
--- a/crates/iota-sdk-grpc-client/src/api/common.rs
+++ b/crates/iota-sdk-grpc-client/src/api/common.rs
@@ -63,6 +63,27 @@ pub enum Error {
     Grpc(Box<tonic::Status>),
 }
 
+impl Error {
+    /// Returns `true` if the error carries a `NOT_FOUND` status, whether the
+    /// server reported it for the call or for a single item of a batched
+    /// request.
+    ///
+    /// **Warning:** do not test the outer error of a batched read to decide
+    /// that an item is absent. Those calls report an absent item against the
+    /// request that asked for it, so a `NOT_FOUND` for the call means the
+    /// endpoint is wrong; treating it as a missing item turns a misconfigured
+    /// endpoint into a normal "not there" answer. Test the per-item result
+    /// instead. Calls returning a single response, such as fetching a
+    /// checkpoint, do report absence at the call level.
+    pub fn is_not_found(&self) -> bool {
+        match self {
+            Error::Server(status) => status.code == i32::from(tonic::Code::NotFound),
+            Error::Grpc(status) => status.code() == tonic::Code::NotFound,
+            _ => false,
+        }
+    }
+}
+
 impl From<TryFromProtoError> for Error {
     fn from(err: TryFromProtoError) -> Self {
         Error::ProtoConversion(Box::new(err))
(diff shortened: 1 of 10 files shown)


crates/iota-sdk-grpc-client/src/api/common.rs
```diff
@@ -63,6 +63,27 @@ pub enum Error {
...
+    /// server reported it for the call or for a single item of a batched
+    /// request.
+    ///
+    /// **Warning:** do not test the outer error of a batched read to decide
+    /// that an item is absent. Those calls report an absent item against the
+    /// request that asked for it, so a `NOT_FOUND` for the call means the
+    /// endpoint is wrong; treating it as a missing item turns a misconfigured
+    /// endpoint into a normal "not there" answer. Test the per-item result
+    /// instead. Calls returning a single response, such as fetching a
+    /// checkpoint, do report absence at the call level.
```
Thoralf-M, line 77: Do you think it would be useful to add a small helper fn wich iterates over the returned items and checks for each if it exists or not and return the missing IDs?
```
let results = self
            .client
            .get_objects(refs, None)
            .await?;
let missing_object = missing_object_refs(results);
if !missing_object.is_empty() {
 panic!("rekt");
}
// can assume everything exists

```

crates/iota-sdk-grpc-client/src/api/ledger/objects.rs
```diff
@@ -61,6 +66,9 @@ impl Client {
     ///     .await?;
     ///
     /// for obj in objs.body() {
+    ///     // Skip the objects the node could not serve
+    ///     let Ok(obj) = obj else { continue };
```
Thoralf-M: Probably better to at least print a warning in the doc examples

bingyanglin (approved):

crates/iota-sdk-grpc-client/src/api/ledger/objects.rs
muXxer: Changed the examples to print the error instead of silently skipping — also fixed the same pattern in the crate-level example in `lib.rs`, which had `if let Some(Ok(..))` and would have taught the same habit.

On the helper idea from the other thread: added a snippet to `get_objects` showing how to pair results with the requested refs to get the missing IDs, which covers the use case without new public API. It filters on `NOT_FOUND` specifically, so a `FAILED_PRECONDITION` (pruned object behind a read-mask field) or an internal error isn't miscounted as a missing object.

Since that snippet tells users to rely on positional pairing, I've also documented the request-order guarantee in `ledger_service.proto` — it was only an implementation detail of the server before.

Thoralf-M (approved):

crates/iota-sdk-grpc-client/src/transaction_builder_client.rs
```diff
@@ -28,14 +28,20 @@ const WAIT_FOR_TX_TIMEOUT: Duration = Duration::from_secs(60);
...
+/// Extract the result for the single item of a one-item batch request.
+///
+/// A `NOT_FOUND` for the item maps to `None`, since these callers treat a
+/// missing object or transaction as an absence rather than a failure. A failure
+/// of the call itself still propagates — including a `NOT_FOUND`, which says
+/// the endpoint is wrong rather than that the item is absent.
+fn single_item<T>(
+    response: Result<MetadataEnvelope<Vec<Result<T, Error>>>, Error>,
+) -> Result<Option<T>, Error> {
+    match response?.into_inner().into_iter().next() {
```
semenov-vladyslav: Should it check if there are more than one item? How should it be iterpreted?

crates/iota-sdk-grpc-client/src/transaction_builder_client.rs
```diff
@@ -28,14 +28,20 @@ const WAIT_FOR_TX_TIMEOUT: Duration = Duration::from_secs(60);
...
+        Some(Ok(item)) => Ok(Some(item)),
+        Some(Err(e)) if e.is_not_found() => Ok(None),
+        Some(Err(e)) => Err(e),
+        None => Ok(None),
```
semenov-vladyslav: No elements in the result list is not the same as single NOT_FOUND entry. To me it would indicate an error since server is obliged to return the vector of expected length. Maybe add a warn message in both cases?

crates/iota-sdk-grpc-client/src/transaction_builder_client.rs
muXxer: It should, yes — it was silently dropping anything past the first result. Fixed together with your other comment.

`single_item` now treats anything other than exactly one result as `ProtocolError::UnexpectedResultCount`. And since the count is only meaningful against the request, I put the check where the request length is actually known: `get_objects` and `get_transactions` now verify the collected results match `refs.len()` / `digests.len()`. That matters beyond this helper — the docs tell callers to `zip` results with requests to find the missing IDs, and a short vector would have mis-paired IDs with results rather than failing.

crates/iota-sdk-grpc-client/src/transaction_builder_client.rs
muXxer: Agreed, and this was a real bug rather than just a naming imprecision: `object()` would have reported an empty batch as "the object does not exist", and `wait_for_tx` would have read it as "not indexed yet" and polled to the 60s timeout instead of failing.

An empty batch is now `ProtocolError::UnexpectedResultCount { expected: 1, actual: 0 }`, and the same check runs in `get_objects`/`get_transactions` against the request length. I went with an error rather than a warning — the crate has no `tracing`/`log` dependency, and a broken response contract is not something the caller can act on by reading a log line. Tests cover the empty batch and the over-long batch.

semenov-vladyslav (approved):

PR #1293, opened 2026-07-31
fix(grpc-client): check batched reads answered the items requested

DaughterOfMars:
## Why

#1292 added `check_result_count`, which confirms a batched read returned as many results as items were requested. That says how many came back, not that position `i` holds item `i`. Callers pair the two by position — and go on to build and sign transactions over the result — so an out-of-order or substituted answer had nothing standing in its way.

Downstream this was being handled per caller: iotaledger/iota#12341 checked ids in the CLI, and the transaction builder checked them again in `fetch_input_objects`. It belongs in the client that made the request and knows what it asked for.

## What

`get_objects` and `get_transactions` compare each answered object id / transaction digest against the one requested in that position.

- Mismatch → `ProtocolError::UnexpectedObject` / `UnexpectedTransaction`, carrying the position and both values as `ObjectId` / `TransactionDigest` rather than strings, so the message reads `requested object 0x… at position 1, but got 0x…`.
- Missing id or digest → the existing `EmptyResponseField`, since that is a different failure from a mismatch.
- Positions the server reported an error for are skipped: they carry a status, not an item, so there is no identity to compare.

**One thing worth a look:** the id and digest are now appended to whatever read mask the caller passes, because a narrow mask (say `ObjectField::BCS` alone) would otherwise leave nothing to check against and turn a working call into `EmptyResponseField`. That means a caller can get back one field it did not ask for. The alternative was a best-effort check that silently skips when the id is absent, which seemed like the wrong property for something guarding what gets signed.

## Test plan

Four unit tests on the object check — in-order accepted, substitution rejected with both ids and the position, an error position skipped, and a missing id reported as `EmptyResponseField`.

`cargo nextest run -p iota-sdk-grpc-client --all-features` (15 tests), 22 doctests, `make clippy`, nightly fmt and `dprint fmt` all clean.

Not covered by a unit test: the transaction check, which needs a proto `ExecutedTransaction` fixture; it is the same shape as the object one.

## Follow-up

Once this lands, the duplicate checks downstream come out — the builder's `fetch_input_objects` and the CLI's `grpc_input_refs` in iotaledger/iota#12341.

### DIFF
--- a/crates/iota-sdk-grpc-client/src/api/common.rs
+++ b/crates/iota-sdk-grpc-client/src/api/common.rs
@@ -24,7 +24,7 @@ use iota_grpc_types::{
         types::ObjectId as ProtoObjectId,
     },
 };
-use iota_types::{ObjectId, TransactionDigest};
+use iota_types::{ObjectId, TransactionDigest, Version};
 use serde::Serialize;
 
 use super::MetadataEnvelope;
@@ -138,6 +138,24 @@ pub enum ProtocolError {
     /// number of requested items.
     #[error("expected {expected} results, got {actual}")]
     UnexpectedResultCount { expected: usize, actual: usize },
+
+    /// `get_objects` answered a position with a different object than the one
+    /// requested there.
+    #[error("requested object {expected} at position {position}, but got {actual}")]
+    UnexpectedObject {
+        position: usize,
+        expected: ObjectId,
+        actual: ObjectId,
+    },
+
+    /// `get_transactions` answered a position with a different transaction than
+    /// the one requested there.
+    #[error("requested transaction {expected} at position {position}, but got {actual}")]
+    UnexpectedTransaction {
+        position: usize,
+        expected: TransactionDigest,
+        actual: TransactionDigest,
+    },
 }
 
 /// Errors during checkpoint data stream reassembly.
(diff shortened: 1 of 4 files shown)


crates/iota-sdk-grpc-client/src/api/common.rs
```diff
@@ -287,6 +305,76 @@ pub fn check_result_count<T>(results: &[T], expected: usize) -> Result<()> {
...
+///
+/// Check that each answered object is the one requested in that position.
+///
+/// A matching count only says how many results came back, not that position `i`
+/// holds object `i`. Callers pair the two by position and go on to build and
+/// sign transactions over the result, so an out-of-order or substituted answer
+/// has to be caught rather than trusted.
+///
+/// Positions the server reported an error for are skipped: they carry a status,
+/// not an object, so there is no id to compare.
```
Thoralf-M: Comment should be more concise, it's not that complicated what it does

crates/iota-sdk-grpc-client/src/api/ledger/transactions.rs
```diff
@@ -128,12 +129,15 @@ impl Client {
                 .collect(),
         );
 
+        // The digests are needed to check that each slot answers the
+        // transaction asked for in that position, so they are requested
+        // whatever the caller's mask says.
+        let mut mask = field_mask_with_default(read_mask, GET_TRANSACTIONS_READ_MASK);
+        mask.paths.push(TRANSACTION_DIGEST.to_owned());
```
Thoralf-M: This feels wrong, we shouldn't add network overhead in the requests for such checks and the reasoning in the PR description is wrong, if we have the BCS we can compute the digest from that

DaughterOfMars: It's a tradeoff, I guess. In the other case we have to do a bunch of extra computation. But you're probably still right that it would be better to just compute it.

DaughterOfMars: The real problem is if the user doesn't request either BCS or the digest, in which case we can't check at all

Thoralf-M: Actually I'd say if the field doesn't exist, don't check it
Even if the field exists, a node could also return wrong data in the other fields, so we have to trust it in some parts anyways, for txs it's also less important since it's not used in other txs for signing as the objects

Alex6323 (approved):

crates/iota-sdk-grpc-client/src/api/ledger/transactions.rs
DaughterOfMars: Okay that's fine I guess

crates/iota-sdk-grpc-client/src/api/common.rs
DaughterOfMars: Done

crates/iota-sdk-grpc-client/src/api/ledger/transactions.rs
DaughterOfMars: I made some changes

Thoralf-M (approved):

PR #1296, opened 2026-08-03
fix(examples): avoid type-only object queries that time out

Thoralf-M:
Three examples filtered objects by type alone. Testnet holds too many `StakedIota` objects for that, so the GraphQL server answers with `Query request timed out. Limit: 8s` and the examples jobs fail. Reported upstream as iotaledger/iota#12536.

- `decode_staked_iota` and `unstake`: also filter by owner, using the address the other examples already use — it owns a `StakedIota`.
- `objects_by_type`: query IOTA coins instead, so a type-only filter stays the point of the example without hitting the timeout.

All three changed in every language: Rust, Go, Kotlin, Python, C#, Swift, WASM.

Verified locally against testnet — the Rust and Python variants of all three return in about a second.

[run-ci]

### DIFF
--- a/bindings/csharp/examples/DecodeStakedIota/Program.cs
+++ b/bindings/csharp/examples/DecodeStakedIota/Program.cs
@@ -15,29 +15,16 @@ class Program
     {
         var client = GraphQlClient.NewTestnet();
 
-        // Filtering objects by type alone scans every object on the network, which
-        // the GraphQL server rejects with a timeout. Pick a recent staker and filter
-        // by owner as well, so only that address' objects are looked at.
-        var stakers = await client.Transactions(
-            new TransactionsFilter(Function: "0x3::iota_system::request_add_stake"),
-            new PaginationFilter(Direction.Backward, Limit: 1));
-
-        if (stakers.Data.Length == 0)
-        {
-            Console.WriteLine("No staking transactions on testnet right now.");
-            return;
-        }
-
-        var staker = stakers.Data[^1].Transaction.Sender();
-        Console.WriteLine($"Latest staker: {staker.ToHex()}\n");
-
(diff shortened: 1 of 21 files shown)


bindings/csharp/examples/DecodeStakedIota/Program.cs
```diff
@@ -14,13 +14,17 @@ class Program
     static async Task Main(string[] args)
     {
         var client = GraphQlClient.NewTestnet();
-        var filter = new ObjectFilter(TypeTag: "0x3::staking_pool::StakedIota");
+
+        // Filtering by type alone scans every object on the network, which the
+        // GraphQL server rejects with a timeout, so filter by owner as well.
```
DaughterOfMars: Is this a bug?

Thoralf-M: Yes — filed as iotaledger/iota#12536. A type-only filter should not need an owner to complete, so this is a workaround, not the expected shape of the query.

The timeout only happens for types with a lot of objects; measured against testnet, `StakedIota` times out even with `first: 3`, while `Coin<IOTA>`, `CoinMetadata` and `StakingPool` all come back in under 2s with the same type-only filter. Adding `owner` brings the `StakedIota` query down to ~1s.


---

DaughterOfMars: Okay, cool. Do you think we actually need these comments? I feel they are unnecesary AI bloat

Thoralf-M: Dropped them in `3035b3b` — the reason lives in this PR and in iotaledger/iota#12536, which is where it belongs.


---

DaughterOfMars (approved):

PR #1297, opened 2026-08-03
feat: add on-chain deny-rule transaction kinds and DenyRuleSet

filipdulic:
## Why

Implements iotaledger/iota#12499 (epic iotaledger/iota#12498): validators will mirror the stake-aggregated deny-rule set into a shared on-chain object at `0xDE9` via a new system transaction, and `TransactionKind` lives in this SDK. The monorepo follow-up (iotaledger/iota#12501) pins the release cut from this PR.

## What

Adds `DenyRuleSet`, the `TransactionDenyRulesUpdate` payload, the two appended transaction kinds, the `0xDE9` `TRANSACTION_DENY_RULES` constants, and their FFI mirrors. `DenyRuleSet` is copied byte-compatible from the monorepo — it is the payload of a live consensus message, so field order and encoding are fixed, and the new enum tags are permanent and pinned in tests immediately.


crates/iota-sdk-types/src/address.rs
```diff
@@ -87,6 +87,7 @@ impl Address {
     pub const AUTHENTICATOR_STATE: Self = Self::from_u16(7);
     pub const RANDOMNESS_STATE: Self = Self::from_u16(8);
     pub const GENESIS_IOTA_BRIDGE: Self = Self::from_u16(9);
+    pub const TRANSACTION_DENY_RULES: Self = Self::from_u16(0xa);
```
Thoralf-M: Why not 
```suggestion
    pub const TRANSACTION_DENY_RULES: Self = Self::from_u16(10);
```
?

crates/iota-sdk-types/src/transaction/mod.rs
```diff
@@ -273,6 +273,88 @@ pub struct RandomnessStateUpdate {
...
+#[cfg_attr(feature = "bcs-schema", derive(iota_bcs_schema::BcsSchema))]
+pub struct DenyRuleSet {
+    /// Addresses denied as transaction sender or gas sponsor. A denied
+    /// address can still receive objects.
+    pub denied_addresses: Vec<Address>,
+    /// Objects denied as transaction inputs or receiving objects.
+    pub denied_objects: Vec<ObjectId>,
+    /// Packages denied as a (transitive) dependency of any command; upgrading
+    /// a denied package is denied too.
+    pub denied_packages: Vec<ObjectId>,
```
Thoralf-M: Should these be `BTreeSet` instead?

filipdulic: This is deliberate.

The wire encoding is the same — a sorted Vec and a BTreeSet produce identical BCS bytes. The issue is that BTreeSet breaks the grammar fuzzer’s byte-exact round-trip, since the ABNF grammar cannot express its canonical ordering. Supporting it would require weakening the transaction-envelope tests to fixed-point checks and adding BTreeSet support to the schema generator.

Using a Vec lets us keep the grammar exact without changing the test harness. Internally, the node still uses a BTreeSet and only converts to a Vec at the serialization boundary.

This is similar to Sui’s AuthenticatorStateUpdate, which uses Vec<ActiveJwk> on the wire while the producer is set-backed and the consumer validates the list.

crates/iota-sdk-types/src/address.rs
filipdulic: ammeded, thanks

crates/iota-sdk-types/src/transaction/mod.rs
Thoralf-M: Isn't the node supposed to make use of this type internally at the end? Sounds a bit wrong to use a Vec just for some fuzzer round trip (if that's the only reason)

filipdulic: I thought about it some more, and will implement it as a set.

crates/iota-sdk-bcs-schema/src/lib.rs
```diff
@@ -193,6 +193,15 @@ fn type_to_schema(ty: &Type) -> String {
                     }
                     None => "(size *(unknown unknown))".into(),
                 },
+                // A BCS set has the same wire shape as a vector; the grammar
+                // cannot express the canonical (sorted, unique) element order,
+                // which is a BCS-level semantic like map key order.
+                "BTreeSet" | "HashSet" => match extract_single_generic(seg) {
```
Thoralf-M: `HashSet` doesn't belong in this arm — its BCS encoding isn't canonical, so the comment above it is false for half the types it covers. 64 `HashSet<u32>` values each populated with `0..8`, in one process, produce 64 distinct encodings (`RandomState` gives every instance its own hash key); `BTreeSet` produces the one ascending encoding. A `HashSet` field would give the same logical value a different transaction digest on every construction.

Nothing in the tree uses `HashSet` in a BCS type today, so dropping it here costs nothing and makes the next one fail loudly on an unknown `hash-set` rule instead of getting a schema that looks correct.


---

Thoralf-M (approved):

DaughterOfMars (approved):

PR #1298, opened 2026-08-03
feat(crypto): add CommitteeChainVerifier

### LINKED ISSUE #1287: Add `CommitteeChainVerifier` to the SDK
## Description

Add an SDK-native [`CommitteeChainVerifier`](https://github.com/iotaledger/iota/blob/ac3c0d2402047bc7c7c4f286cbd95191a0e492d1/crates/iota-types/src/committee.rs#L477-L554) to `iota-sdk-crypto::validator`.

## Motivation

Proof of Inclusion needs to authenticate historical committees from a trusted genesis committee using signed end-of-epoch checkpoint summaries.

The verifier currently exists only in the IOTA node repository. SDK consumers must either depend on node-specific crates or duplicate security-sensitive committee-transition logic.

DaughterOfMars:
## Why

Proof of Inclusion needs to authenticate historical committees from a trusted genesis committee using signed end-of-epoch checkpoint summaries. The verifier only existed in the node repo, forcing SDK consumers to depend on node-internal crates or reimplement security-sensitive committee-transition logic.

## What

- **`iota-sdk-crypto::validator::CommitteeChainVerifier`** — start at a trusted committee and feed each epoch's closing `SignedCheckpointSummary` in order; each one is verified against the current committee, then advances the walk to the committee it elects. Transport-agnostic: the caller drives the loop. Structural checks run before signature verification, and errors leave the verifier's state unchanged.
- Built on the existing `ValidatorCommitteeSignatureVerifier`; new `CommitteeChainError` (thiserror, `#[non_exhaustive]`).
- **FFI**: exposed the verifier, plus a `SignedCheckpointSummary` binding it consumes (this core type had no FFI wrapper), with BCS/JSON round-trips.

## Test plan

- Proptests for the happy-path multi-epoch walk and each rejection path (wrong epoch, non-closing checkpoint, bad signature). Full crypto suite passes.
- `clippy`, `cargo +nightly fmt --check`, and `wasm32` all clean.
- Generated the Python bindings via the in-repo `uniffi-bindgen` to confirm the uniffi metadata is valid end-to-end.

Closes #1287


crates/iota-sdk-crypto/src/validator.rs
```diff
@@ -209,6 +209,118 @@ impl Verifier<ValidatorAggregatedSignature> for ValidatorCommitteeSignatureVerif
...
+    /// now committee-verified summary.
+    ///
+    /// Errors leave the verifier unchanged, e.g. if the summary is for a
+    /// different epoch, its signatures don't verify under the current
+    /// committee, or it is not a close-of-epoch checkpoint (no end-of-epoch
+    /// data).
+    pub fn verify_epoch_close(
+        &mut self,
+        summary: SignedCheckpointSummary,
+    ) -> Result<CheckpointSummary, CommitteeChainError> {
```
Thoralf-M: Why consume it and return a subset instead of only taking a reference and not returning anything?

crates/iota-sdk-crypto/src/validator.rs
```diff
@@ -209,6 +209,118 @@ impl Verifier<ValidatorAggregatedSignature> for ValidatorCommitteeSignatureVerif
...
+            .verify_checkpoint_summary(&checkpoint, &signature)?;
+
+        // Signatures verified; the elected committee is now trusted.
+        let next_committee = ValidatorCommittee {
+            epoch: expected_epoch
+                .checked_add(1)
+                .ok_or(CommitteeChainError::EpochOverflow)?,
+            members: end_of_epoch_data.next_epoch_committee.clone(),
+        };
+        self.verifier = ValidatorCommitteeSignatureVerifier::new(next_committee)?;
```
Thoralf-M, line 318: `ValidatorCommitteeSignatureVerifier::new` computes `((total_weight - 1) / 3) * 2 + 1`, so a committee with zero total weight underflows: panic in debug, and in release a `quorum_threshold` near `u64::MAX` that rejects every later epoch. `total_weight += member.stake` can overflow-panic too.

That's pre-existing, but this is the first path that builds a committee out of checkpoint payload instead of caller-supplied data, so an end-of-epoch summary electing an empty (or all-zero-stake) committee now reaches it. `new` already returns `Result` — rejecting those committees there turns this into a `CommitteeChainError` and covers the other callers as well.

---

DaughterOfMars: I added some protections

Thoralf-M (approved):

crates/iota-sdk-ffi/src/crypto/validator.rs
```diff
@@ -105,3 +105,46 @@ impl ValidatorCommitteeSignatureAggregator {
             .into())
     }
 }
+
+#[derive(Debug, derive_more::From, uniffi::Object)]
+#[uniffi::export(Debug)]
+pub struct CommitteeChainVerifier(pub RwLock<iota_sdk::crypto::validator::CommitteeChainVerifier>);
```
thibault-martinez, line 124: Why an RwLock?

thibault-martinez: Oh, `verify_epoch_close` takes mut

thibault-martinez, line 124: Put the same doc as the main type?

crates/iota-sdk-types/src/validator.rs
```diff
@@ -26,6 +26,75 @@ pub struct ValidatorCommittee {
...
+    ///
+    /// A valid committee is one that:
+    ///  - Has at least one member
+    ///  - Has a nonzero total stake that fits in a [`StakeUnit`]
+    ///  - Contains no duplicate public keys
+    ///
+    /// Deserialization and the public fields both bypass this check, so a
+    /// committee that came off the wire or out of a checkpoint payload has to
+    /// be validated before its stake is used to derive a quorum threshold.
+    pub fn validate(&self) -> Result<(), ValidatorCommitteeError> {
```
thibault-martinez, line 64: Where does this come from?

crates/iota-sdk-crypto/src/validator.rs
```diff
@@ -209,6 +213,123 @@ impl Verifier<ValidatorAggregatedSignature> for ValidatorCommitteeSignatureVerif
...
+    }
+
+    /// The committee of [`Self::epoch`] (trusted root or chain-verified).
+    pub fn committee(&self) -> &ValidatorCommittee {
+        self.verifier.committee()
+    }
+
+    /// Verify `summary` as the certified closing checkpoint of [`Self::epoch`]
+    /// and advance to the committee it elects for the next epoch, returning the
+    /// now committee-verified summary.
```
thibault-martinez: It does not return anything

crates/iota-sdk-crypto/src/validator.rs
```diff
@@ -30,17 +30,21 @@ struct MemberInfo<'a> {
 
 impl ExtendedValidatorCommittee {
     fn new(committee: ValidatorCommittee) -> Result<Self, SignatureError> {
+        let total_weight = committee
+            .validate()
+            .and_then(|()| committee.total_stake())
```
thibault-martinez, line 35: bit sad that we compute total stake twice, would it be weird if validate returns it?

crates/iota-sdk-ffi/src/types/validator.rs
```diff
@@ -168,7 +183,15 @@ impl ValidatorAggregatedSignature {
...
+    ValidatorCommittee,
+    ValidatorSignature,
+    ValidatorAggregatedSignature
+);
+crate::export_iota_types_json_conversion!(ValidatorCommitteeMember);
+crate::export_iota_types_objects_json_conversion!(
+    ValidatorCommittee,
+    ValidatorSignature,
+    ValidatorAggregatedSignature
+);
```
thibault-martinez, line 197: This is bypassing validation that exists in `ValidatorCommittee::new`. Maybe we should handroll these two? Or do we not care?

crates/iota-sdk-types/src/validator.rs
DaughterOfMars: The method? It is just useful

crates/iota-sdk-ffi/src/types/validator.rs
DaughterOfMars: This is why the check still exists. I don't think we can add serialization validation because the node still needs to deserialize invalid committees.

crates/iota-sdk-crypto/src/validator.rs
DaughterOfMars: I think it would be a bit weird

crates/iota-sdk-ffi/src/crypto/validator.rs
DaughterOfMars: yep

DaughterOfMars: done

crates/iota-sdk-crypto/src/validator.rs
DaughterOfMars: fixed

crates/iota-sdk-crypto/src/validator.rs
Alex6323: we could cache the result, no?

DaughterOfMars: not really

Alex6323:
Mark this PR as breaking? (edit: because `uniffi::Record` became `uniffi::Object`)

crates/iota-sdk-crypto/src/validator.rs
```diff
@@ -382,4 +502,227 @@ mod tests {
...
+        // Certified by a committee the verifier doesn't trust.
+        let signed = certify(
+            &wrong_keys,
+            committee(&wrong_keys, 0),
+            close_epoch(summary, 0, &next),
+        );
+        let err = verifier.verify_epoch_close(&signed).unwrap_err();
+        assert!(matches!(err, CommitteeChainError::Signature(_)));
+        assert_eq!(verifier.committee(), &committee0);
+    }
```
Alex6323, line 727: Claude suggested to add 2 more tests here that cover 2 very specific cases (one failure, one success case):
(a) insufficient signing weight (currently it's checked that one cannot build an under-quorum certificate, but that the verifier rejects it is NOT checked)
(b) minimum weight is sufficient (just tests the boundary)

```
/// A closing checkpoint certified by a stake minority fails on weight, not
/// on the signature: one member's signature verifies, the quorum does not.
/// `certify` cannot build this case, since `finish` refuses it.
#[proptest]
fn under_quorum_certification_is_rejected(
    keys: [Bls12381PrivateKey; 4],
    summary: CheckpointSummary,
) {
    let committee0 = committee(&keys, 0);
    let next = committee(&keys, 1);
    let mut verifier = CommitteeChainVerifier::new(committee0.clone()).unwrap();

    let checkpoint = close_epoch(summary, 0, &next);
    // One of four equal stakes: weight 1 against a threshold of 3.
    let lone = keys[0].sign_checkpoint_summary(&checkpoint);
    let mut bitmap = roaring::RoaringBitmap::new();
    bitmap.insert(0);
    let signed = SignedCheckpointSummary {
        checkpoint,
        signature: ValidatorAggregatedSignature {
            epoch: lone.epoch,
            signature: lone.signature,
            bitmap,
        },
    };

    let err = verifier.verify_epoch_close(&signed).unwrap_err();
    assert!(matches!(err, CommitteeChainError::Signature(_)), "{err}");
    assert!(err.to_string().contains("insufficient signing weight"), "{err}");
    assert_eq!(verifier.committee(), &committee0);
}

/// Exactly the threshold is enough: three of four equal stakes close the
/// epoch.
#[proptest]
fn certification_at_the_quorum_threshold_is_accepted(
    keys: [Bls12381PrivateKey; 4],
    summary: CheckpointSummary,
) {
    let committee0 = committee(&keys, 0);
    let next = committee(&keys, 1);
    let mut verifier = CommitteeChainVerifier::new(committee0.clone()).unwrap();

    let signed = certify(&keys[..3], committee0, close_epoch(summary, 0, &next));
    verifier.verify_epoch_close(&signed).unwrap();
    assert_eq!(verifier.epoch(), 1);
}
```

DaughterOfMars: My claude double checked and found only the first one necessary

crates/iota-sdk-types/src/validator.rs
thibault-martinez: No but I mean, did you take this from the monorepo?

DaughterOfMars: Sort of? All the various checks are done in some way in the mono repo, but not exactly like this

crates/iota-sdk-ffi/src/types/checkpoint.rs
```diff
@@ -416,6 +441,7 @@ crate::export_iota_types_objects_bcs_conversion!(
 crate::export_iota_types_json_conversion!(EndOfEpochData);
 crate::export_iota_types_objects_json_conversion!(
     CheckpointSummary,
+    SignedCheckpointSummary,
```
thibault-martinez, line 444: Also add to display macro?

crates/iota-sdk-ffi/src/types/validator.rs
```diff
@@ -168,9 +183,15 @@ impl ValidatorAggregatedSignature {
     }
 }
 
-crate::export_iota_types_bcs_conversion!(ValidatorCommittee, ValidatorCommitteeMember);
-crate::export_iota_types_objects_bcs_conversion!(ValidatorSignature, ValidatorAggregatedSignature);
-crate::export_iota_types_json_conversion!(ValidatorCommittee, ValidatorCommitteeMember);
-crate::export_iota_types_objects_json_conversion!(ValidatorSignature, ValidatorAggregatedSignature);
-crate::export_iota_types_display!(ValidatorCommittee, ValidatorCommitteeMember);
-crate::export_iota_types_objects_display!(ValidatorSignature, ValidatorAggregatedSignature);
```
thibault-martinez, line 176: You are removing these two, a failed conflict resolution?

DaughterOfMars: yeah, good catch

crates/iota-sdk-ffi/src/types/checkpoint.rs
DaughterOfMars: fixed

thibault-martinez (approved):

Alex6323 (approved):

PR #1301, opened 2026-08-04
feat(grpc): make field path `as_str` const

DaughterOfMars:
Makes the generated `as_str` on read mask field path types a `const fn`.

The typed constants can't be mixed across namespaces, so a user wanting one list of paths spanning several endpoints had no way to build it at compile time. Going through `as_str` in a `const` gives them a plain `&[&str]`.


crates/iota-sdk-grpc-types/src/read_mask_fields.rs
```diff
@@ -961,6 +964,18 @@ mod tests {
         assert_eq!(DynamicFieldField::CHILD_OBJECT, "child_object");
     }
 
+    #[test]
+    fn field_paths_usable_in_consts() {
```
Thoralf-M: Do you really see a value in this test? Looks like we're testing Rust itself here

DaughterOfMars: I'll remove it. AI junk

Thoralf-M (approved):

thibault-martinez (approved):

PR #1302, opened 2026-08-04
feat(grpc)!: fold nested read mask field namespaces into their endpoints

DaughterOfMars:
Removes `CheckpointTransactionField`, `CheckpointEventField` and `SimulateExecutedTransactionField`, moving their paths onto `CheckpointResponseField` and `SimulateField`.

The three types had no `From`/`IntoReadMask` impl into any mask, so no value of theirs could reach a client method — `define_scoped_read_mask!` only wires up the single field type it's given, while their doc comments told callers to use them with the endpoint's mask. Nothing in this repo referenced them.

Folding rather than adding the missing impls is what makes a mask spanning both levels expressible: two field types can't share one array, so `&[SimulateField]` was the only shape that lets a caller keep e.g. a dry-run mask of `executed_transaction.*` paths plus `suggested_gas_price` in a single constant.

Breaking in name only — the types are unreachable today and unreleased (no tag contains #1253).

### DIFF
--- a/crates/iota-sdk-grpc-types/src/read_mask_fields.rs
+++ b/crates/iota-sdk-grpc-types/src/read_mask_fields.rs
@@ -591,6 +591,8 @@ define_field_paths! {
         TRANSACTIONS_EVENTS = "transactions.events",
         /// The events digest.
         TRANSACTIONS_EVENTS_DIGEST = "transactions.events.digest",
+        /// Individual events — full BCS-encoded.
+        TRANSACTIONS_EVENTS_EVENTS_BCS = "transactions.events.events.bcs",
         /// Checkpoint sequence number of the transaction.
         TRANSACTIONS_CHECKPOINT = "transactions.checkpoint",
         /// Timestamp of the transaction.
--- a/crates/iota-sdk-grpc-types/src/read_mask_fields.rs
+++ b/crates/iota-sdk-grpc-types/src/read_mask_fields.rs
@@ -891,34 +891,6 @@ mod tests {
     }
 
     #[test]
-    fn response_and_nested_fields_share_one_mask() {
-        // Paths into a nested response message and paths at the top level are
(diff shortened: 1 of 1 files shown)


crates/iota-sdk-grpc-types/src/read_mask_fields.rs
```diff
@@ -675,42 +571,67 @@ define_field_paths! {
...
+        /// Transaction effects (all sub-fields).
+        TRANSACTIONS_EFFECTS = "transactions.effects",
+        /// The effects digest.
+        TRANSACTIONS_EFFECTS_DIGEST = "transactions.effects.digest",
+        /// The full BCS-encoded effects.
+        TRANSACTIONS_EFFECTS_BCS = "transactions.effects.bcs",
+        /// Transaction events (all sub-fields).
+        TRANSACTIONS_EVENTS = "transactions.events",
+        /// The events digest.
+        TRANSACTIONS_EVENTS_DIGEST = "transactions.events.digest",
```
Thoralf-M, line 593: @muXxer should we also have `TRANSACTIONS_EVENTS_EVENTS_BCS` here? Event bcs exists for `SimulateField`, but not here

bingyanglin: Yes I think we should, because `transactions.events.events.bcs` is used at `CHECKPOINT_RESPONSE_CHECKPOINT_DATA`
https://github.com/iotaledger/iota-rust-sdk/blob/2ef46c27107348379d0b056af9b04e36728a61e8/crates/iota-sdk-grpc-types/src/read_masks.rs#L165
anyway.

DaughterOfMars: Added!

Thoralf-M (approved):

muXxer (approved):

PR #1303, opened 2026-08-04
feat(bindings): expose GraphQL subscriptions over the FFI

DaughterOfMars:
Follow-up to #1194: exposes `events_stream` / `transactions_stream` to the language bindings.

uniffi has no `Stream` equivalent, so a subscription is a handle object that the caller pulls one update at a time with `next`. Two things that shaped the API:

`cancel` lives on the handle instead of being left to the caller's runtime, because the generated Go and C# async wrappers are blocking with no cancellation hook — without it, a `next` waiting on a quiet subscription is unstoppable in those two languages. It is not called `close`: that name overrides the disposal method uniffi generates for Kotlin objects, so the handle would stop being freed by `use { }`.

A lost connection or a client that falls behind the server arrives as an `Interrupted` update rather than an error, because the subscription recovers from both on its own and an exception reads as terminal in every idiomatic consumer. Only errors the reconnect loop cannot handle — a rejected filter, for instance — are raised.

The TS package gets no working subscriptions: the transport is `tokio-tungstenite`, which is not built for `wasm32`. The methods are still exported there and `next` raises, because `ubrn` generates the wasm glue from the host library — an API missing on `wasm32` leaves that glue referencing symbols the wasm library never exports, and the link fails. Making them work needs a browser `WebSocket` transport in the client crate.

Verified locally: all five bindgens accept the API; the Go and Kotlin examples compile and the Python and Swift ones pass their format checks. Nothing is run end-to-end locally — localnet needs Docker here — so CI's `start-local-network` job is the first real execution of the examples.

[run-ci]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ unused_async = "warn"
 
 [workspace.dependencies]
 # external dependencies
+async-stream = "0.3"
 base64ct = "1.6.0"
 bcs = "0.1.6"
 cynic = "3.7.3"
@@ -37,6 +38,7 @@ eyre = "0.6"
 futures = "0.3.29"
 getrandom_2 = { version = "0.2", package = "getrandom", features = ["js"] }
 getrandom_3 = { version = "0.3", package = "getrandom", features = ["wasm_js"] }
+graphql-ws-client = { version = "0.13", default-features = false, features = ["client-cynic", "tungstenite-0.28"] }
 hex = "0.4.3"
 js-sys = "0.3"
 paste = "1.0.15"
@@ -57,6 +59,7 @@ syn = { version = "2", features = ["full"] }
 test-strategy = "0.4.0"
 thiserror = "2.0"
 tokio = "1.52.3"
+tokio-tungstenite = { version = "0.28", default-features = false, features = ["connect", "rustls-tls-webpki-roots"] }
 tonic = "0.14"
 tracing = "0.1"
 variadics_please = "1.1"
--- a/crates/integration-tests/Cargo.toml
+++ b/crates/integration-tests/Cargo.toml
@@ -8,6 +8,7 @@ publish = false
 [dev-dependencies]
 # external dependencies
 eyre.workspace = true
+futures.workspace = true
 rand.workspace = true
 serde_json.workspace = true
 tokio.workspace = true
--- a/crates/integration-tests/tests/graphql_ptbs.rs
+++ b/crates/integration-tests/tests/graphql_ptbs.rs
@@ -464,9 +464,9 @@ async fn test_auto_gas_pins_full_first_page_for_consolidation() {
 
 /// Open a live transactions subscription and assert it delivers a transaction.
 /// A transfer is executed on a background task once the subscription has had a
-/// moment to connect, so the live stream observes it (localnet also produces
-/// system transactions continuously). This exercises the WebSocket transport,
-/// payload decoding, and resume-cursor tracking shared by `events_stream`.
+/// moment to connect, so the live stream observes it. This exercises the
+/// WebSocket transport, payload decoding, and resume-cursor tracking shared by
+/// `events_stream`.
 #[tokio::test]
 async fn test_transactions_subscription() {
     use futures::StreamExt;
--- a/crates/integration-tests/tests/graphql_ptbs.rs
+++ b/crates/integration-tests/tests/graphql_ptbs.rs
@@ -461,3 +461,36 @@ async fn test_auto_gas_pins_full_first_page_for_consolidation() {
         resolved.gas_payment.objects.len(),
     );
 }
+
+/// Open a live transactions subscription and assert it delivers a transaction.
+/// A transfer is executed on a background task once the subscription has had a
+/// moment to connect, so the live stream observes it (localnet also produces
+/// system transactions continuously). This exercises the WebSocket transport,
+/// payload decoding, and resume-cursor tracking shared by `events_stream`.
+#[tokio::test]
+async fn test_transactions_subscription() {
+    use futures::StreamExt;
+
+    let client = Client::new_localnet();
+    let stream = client.transactions_stream(None, None);
+    futures::pin_mut!(stream);
+
+    tokio::spawn(async move {
+        // Give the subscription time to connect before generating activity.
+        tokio::time::sleep(Duration::from_secs(2)).await;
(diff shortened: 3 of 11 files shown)


crates/iota-sdk-ffi/src/graphql/subscription.rs
```diff
@@ -0,0 +1,382 @@
...
+};
+use tokio::sync::{Mutex, Notify};
+
+use crate::{
+    error::Result,
+    graphql::{client::GraphQLClient, query_types::GraphQlEvent},
+    types::{address::Address, transaction::SignedTransaction},
+};
+
+/// Filter incoming events in a subscription. Exactly one field may be set.
```
Thoralf-M: There is only one field?

DaughterOfMars: Yes but it could be added to later

bindings/csharp/examples/SubscribeTransactionsGraphQL/Program.cs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2026 IOTA Stiftung
+// SPDX-License-Identifier: Apache-2.0
+
+// Tail transactions as they are executed, over a GraphQL subscription.
```
Thoralf-M, line 4: Reminds me of bundles from the ternary era 😂

bindings/python/examples/subscribe_transactions.py
```diff
@@ -0,0 +1,60 @@
...
+    # otherwise the transaction lands before anyone is listening.
+    await asyncio.sleep(2)
+    address = Address.from_hex(
+        "0x0000a4984bd495d4346fa208ddff4f5d5e5ad48c21dec631ddebc99809f16900")
+    await FaucetClient.new_localnet().request_and_wait(address)
+
+
+async def main():
+    client = GraphQlClient.new_localnet()
+    subscription = await client.transactions_subscription()
```
Thoralf-M: We should also filter for programmable tx kind in the examples otherwise one usually gets a system tx returned

Thoralf-M (approved):

thibault-martinez (changes requested):
Please can you rename the examples to add a graphql prefix like we have for grpc

thibault-martinez:
C# failing

crates/iota-sdk-ffi/src/graphql/subscription.rs
```diff
@@ -0,0 +1,379 @@
...
+        self.cancelled.store(true, Ordering::Release);
+        self.notify.notify_waiters();
+    }
+
+    fn is_cancelled(&self) -> bool {
+        self.cancelled.load(Ordering::Acquire)
+    }
+
+    /// Resolve once [`Cancel::cancel`] has been called.
+    #[cfg(not(target_arch = "wasm32"))]
```
thibault-martinez, line 109: Just curious, if this whole subscription thing is not available for WASM, why not gate the whole module instead of sprinkling attributes here and there?

crates/iota-sdk-ffi/src/graphql/subscription.rs
```diff
@@ -0,0 +1,379 @@
...
+
+/// A cancellation flag that a pending `next` can wait on.
+///
+/// Foreign async support is uneven — Kotlin, Swift and Python can cancel a
+/// pending call, Go and C# cannot — so cancellation has to be something the
+/// subscription itself understands rather than something the caller's runtime
+/// does to it.
+#[derive(Default)]
+struct Cancel {
+    cancelled: AtomicBool,
```
thibault-martinez: Here and everywhere, `canceled` with a single `l`

crates/iota-sdk-ffi/src/graphql/subscription.rs
```diff
@@ -0,0 +1,379 @@
...
+/// Stand-in for the wasm build, where `next` raises instead of reading a
+/// stream.
+#[cfg(target_arch = "wasm32")]
+fn open_transactions(
+    _client: iota_sdk::graphql_client::Client,
+    _filter: Option<SubscriptionTransactionFilter>,
+    _start_after: Option<String>,
+) -> BoxStream<'static, GqlResult<iota_sdk::types::SignedTransaction>> {
+    futures::stream::empty().boxed()
+}
```
thibault-martinez, line 337: Can this not be in the macro somehow?

crates/iota-sdk-ffi/src/graphql/subscription.rs
```diff
@@ -0,0 +1,379 @@
...
+    #[uniffi::method(default(filter = None, start_after = None))]
+    pub async fn transactions_subscription(
+        &self,
+        filter: Option<SubscriptionTransactionFilter>,
+        start_after: Option<String>,
+    ) -> TransactionSubscription {
+        let client = self.0.read().await.clone();
+        TransactionSubscription::new(open_transactions(client, filter, start_after))
+    }
+}
```
thibault-martinez, line 379: same question I guess

crates/iota-sdk-ffi/src/graphql/subscription.rs
```diff
@@ -0,0 +1,379 @@
...
+        ///
+        /// Call `next` in a loop to receive updates; it only returns `None` once
+        /// `cancel` has been called, since the subscription itself never ends.
+        #[derive(uniffi::Object)]
+        pub struct $name {
+            stream: Mutex<BoxStream<'static, GqlResult<$item>>>,
+            cancel: Cancel,
+        }
+
+        #[uniffi::export(async_runtime = "tokio")]
```
thibault-martinez, line 154: Depending whether you gate the whole module or not, this may need to be gated like we usually do?

crates/iota-sdk-ffi/src/graphql/subscription.rs
DaughterOfMars: Some things cannot be gated because they are needed by external types

crates/iota-sdk-ffi/src/graphql/subscription.rs
DaughterOfMars: I mean yes, but IMO in this case we should consider un-macroing instead of macroing more. It's fiddly stream code and I'd rather not have all that in a macro where it's hard to debug.

thibault-martinez: Fair enough

thibault-martinez (approved):

PR #1321, opened 2026-08-07
chore: use US English spellings in comments

thibault-martinez:
## Why

The codebase is US English throughout, with a handful of UK spellings scattered in comments.

## What changes

Comment-only spelling fixes, plus one internal test rename and a proto doc comment (`iota.grpc.v1.filter.rs` is the regenerated output of that proto change). Protocol identifiers that inherit the `Cancelled` spelling are handled separately in a follow-up PR.

## Test plan

`cargo check --workspace --all-features`; `make grpc` reproduces the generated file.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -16,7 +16,7 @@ the diff, delete it. Repeat until a pass deletes nothing.
 
 ## What changes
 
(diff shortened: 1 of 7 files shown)


thibault-martinez:
Folded into #1322.

PR #1324, opened 2026-08-10
feat: add transaction_deny_rules Move type mirrors

filipdulic:
## Why

Client-visible mirrors for the on-chain `TransactionDenyRules` object, needed before the governance flags flip. Tracked by iotaledger/iota#12613.

## What changes

Mirrors of `0x2::transaction_deny_rules`: the object, its inner state, and the update event. Shape-compare entries and the matching pin bump follow separately once the monorepo stack merges.

## Test plan

Crate tests pass. `shapes_match` verified manually with blobs fetched at iotaledger/iota#12517's tip and the entries temporarily registered.


filipdulic:
Parking until the monorepo stack (iotaledger/iota#12541) merges — this will return as one complete PR: mirrors + shape-compare entries + `move-binary-format` pin at a develop rev. Tracked in iotaledger/iota#12613.

PR #1329, opened 2026-08-11
feat: impl From<T> for String for Address, Digest, and TypeTag

### LINKED ISSUE #1320: iota-sdk-types: impl From<T> for String for Addres, Digest, and TypeTag
https://github.com/MystenLabs/sui-rust-sdk/pull/150

thibault-martinez:
Replicates MystenLabs/sui-rust-sdk#150. Also covers `ObjectId`, which upstream merged into `Address` but is a separate type here, and `Identifier`/`StructTag` as in the upstream PR.

Closes #1320

### DIFF
--- a/crates/iota-sdk-types/src/address.rs
+++ b/crates/iota-sdk-types/src/address.rs
@@ -326,6 +326,24 @@ impl From<Address> for Vec<u8> {
     }
 }
 
+impl From<&Address> for Vec<u8> {
+    fn from(value: &Address) -> Self {
+        value.0.to_vec()
+    }
+}
+
+impl From<Address> for String {
+    fn from(value: Address) -> Self {
+        value.to_string()
+    }
+}
+
+impl From<&Address> for String {
+    fn from(value: &Address) -> Self {
+        value.to_string()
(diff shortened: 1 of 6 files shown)


DaughterOfMars (commented):
But why

thibault-martinez:
> But why

Why not? This is decent convenience

Thoralf-M (approved):

DaughterOfMars:
> > But why
> 
> Why not? This is decent convenience

Is it? I don't see any value. We already have ToString which is always used to do this. And this hides the inner working and implies that these types can be converted by value into Strings which isn't true.

thibault-martinez:
> Is it? 

I think so, yeah.

> implies that these types can be converted by value into Strings which isn't true

What do you mean?

DaughterOfMars:
They take an owned `T` but need only `&T`

thibault-martinez:
> They take an owned `T` but need only `&T`

I'm providing both for convenience, what's the actual issue here?

DaughterOfMars (approved):

Issue #1328, opened 2026-08-11
iota-sdk-types: interpret effects here, not in the node (accessors, then change derivation)


Thoralf-M:
`iota-sdk-types` owns the effects wire format but not its interpretation. `TransactionEffects` exposes only `as_v1()` / `into_v1()`, so a client that wants to know what a transaction created, mutated or deleted has to match on the version and walk `changed_objects` with `ObjectIn` / `ObjectOut` / `IdOperation` itself.

That interpretation already exists — in the node monorepo, as the sealed `TransactionEffectsAPI` trait (`iota/crates/iota-types/src/effects/`), implemented for this crate's own `TransactionEffects`. Every method is derived purely from `changed_objects` and `unchanged_shared_objects`; none of it needs node state. Being sealed, a client cannot even implement it.

Two consequences:

- Every client and FFI binding re-implements `created()` / `mutated()` / `deleted()` by hand, and each copy has to be updated when a new effects version lands.
- Code that derives higher-level views from effects cannot live here, because it is written in terms of those accessors.

## Where each piece lives

Accessors on the `TransactionEffects` enum would promise an answer for every future version. The object sets are derived from V1's `ObjectIn` / `ObjectOut` / `IdOperation` model, so a version whose model differs could keep that promise only by returning an empty set (silently telling a caller that a transaction created nothing), by panicking, or by changing signatures and breaking every caller.

So the split is:

- **This crate**: the derivation on the version that defines the model it reads (`TransactionEffectsV1`), plus `as_opt_v1` / `into_opt_v1` so callers can reach it without assuming a version. No cross-version promises.
- **The monorepo**: keeps a small sealed trait mapping over those, for node-internal ergonomics. It has no external consumers, so breaking it when a new version lands is free — which is exactly the promise this crate cannot make.

Two further rules for anything that does get added here:

- **Return `Option` where a concept may not apply.** `gas_object()` is the live example: the node's version returns a dummy reference for a transaction that needs no gas and carries a TODO to make it optional. Shipping the non-optional shape would make fixing it a breaking change.
- **Keep version-specific details as fields, not accessors.** `lamport_version` is how V1 in particular assigns output versions; callers get output versions from the references the accessors return.

## Steps

[...]

PR #1331, opened 2026-08-11
feat(iota-sdk-types): derive the object sets from TransactionEffectsV1

Thoralf-M:
Step 1 of #1328: interpret effects here instead of in the node.

Answering "what did this transaction create, mutate or delete?" means walking `changed_objects` with `ObjectIn` / `ObjectOut` / `IdOperation`. That derivation lives in the node's sealed `TransactionEffectsAPI`, which clients can neither reach nor implement, so every client and binding hand-rolls it.

`TransactionEffectsV1` now provides it: `modified_at_versions`, `old_object_metadata`, `created`, `mutated`, `unwrapped`, `deleted`, `unwrapped_then_deleted`, `wrapped`, `gas_object`. The enum gets its version accessors from `def_is_as_into_opt!`, as `Transaction` does.

## Nothing on the enum

An accessor on `TransactionEffects` promises an answer for every future version. These sets are derived from V1's object model, so a version that models it differently could answer only with an empty set, a panic, or a changed signature. Callers reach the derivation through a version instead, with `as_opt_v1()` for those that do not want to assume one. The node keeps its version-agnostic mapping over these, where breaking it costs nothing.

Two smaller choices: `gas_object()` returns `Option`, since a transaction that needs no gas has none, where the node's returns a dummy reference; and `lamport_version` stays a field, being how V1 in particular assigns output versions, which callers get from the references returned here.

Two places where this reads the effects more loosely than the node's trait does, both for shapes that cannot occur — a package is never unwrapped, and gas is never a package: `unwrapped()` and `gas_object()` accept any written output rather than matching `ObjectWrite` specifically. `gas_object()` also returns `None` for a gas index outside `changed_objects`, where the node panics; these bytes come off the network, so a malformed blob should not be able to panic a caller.

For review: `as_v1()` now panics on a variant that is not V1, where the previous irrefutable `let` would have stopped compiling once a second variant existed. That matches `Transaction`, and `as_opt_v1()` is the non-panicking path, but it trades a compile-time signal for a runtime one.

## Tests

A proptest covers the half that holds for any effects: the six object sets are selected by mutually exclusive shapes, so together they report each changed object at most once.

The rest is against the existing BCS fixtures, which cover the other half — real effects leave none out: the sets partition `changed_objects`; removals carry the right tombstone digest and output version; a sponsored transaction reports its gas object and the genesis transaction none.

The monorepo has no unit tests of this derivation to port — its effects tests cover the digest, and the accessors appear only in node execution tests that need an engine.

`cargo test --all-features` passes (381 tests), clippy is clean, wasm32 builds with default features.

## Not in scope

`input_shared_objects()` and `object_changes()`, which need `InputSharedObject` / `ObjectChange` to move here — step 2 of #1328. Nothing asserts yet that these agree with the node's trait; that belongs in step 4, where both implementations are in scope.


crates/iota-sdk-types/src/lib.rs
```diff
@@ -287,7 +287,7 @@ macro_rules! def_is_as_into_opt {
         paste::paste! {
         #[doc = "Converts this into a " $rename:snake " if it is a " $variant:snake " variant, or returns `None` otherwise."]
         #[inline]
-        pub fn [< into_ $rename _opt >](self) -> Option<$inner> {
+        pub fn [< into_opt_ $rename >](self) -> Option<$inner> {
```
DaughterOfMars, line 290: Why this change?

crates/iota-sdk-types/src/effects/mod.rs
```diff
@@ -28,17 +28,7 @@ pub enum TransactionEffects {
...
-    pub fn as_v1(&self) -> &TransactionEffectsV1 {
-        let Self::V1(effects) = self;
-        effects
-    }
-
-    pub fn into_v1(self) -> TransactionEffectsV1 {
-        let Self::V1(effects) = self;
-        *effects
-    }
+    crate::def_is_as_into_opt!(V1(Box<TransactionEffectsV1>));
```
DaughterOfMars, line 31: Why?

crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -424,3 +424,170 @@ impl crate::TreeDisplay for ObjectOut {
         }
     }
 }
+
+impl TransactionEffectsV1 {
```
DaughterOfMars, line 429: Add these to FFI?

crates/iota-sdk-types/src/effects/mod.rs
Thoralf-M: To align it with the other enums we have, they all use `def_is_as_into_opt`

crates/iota-sdk-types/src/lib.rs
Thoralf-M: `TransactionEffects::V1` is boxed, so it expands the macro's boxed-variant arm — the first thing in the crate to do so, since the other three versioned enums are unboxed and effects previously had hand-written `as_v1`/`into_v1`. That arm has two naming bugs that stayed latent while nothing expanded it: its `as_$name` calls an `as_$name_opt` it never defines (it defines `as_opt_$name`), and it names the optional conversion `into_$name_opt` where the unboxed arm names it `into_opt_$name`. So without this fix the macro line in `effects/mod.rs` fails with `no method named as_v1_opt`.

crates/iota-sdk-types/src/effects/v1.rs
Thoralf-M: Done, in a third commit. They are methods on the `TransactionEffects` object, delegating through `as_v1` the way the `Transaction` object does for `kind` / `sender` / `digest` — the `TransactionEffectsV1` record can't carry methods.

The pairs needed records, since tuples don't cross the boundary: `OwnedObjectReference { reference, owner }` and `ObjectVersion { object_id, version }`. Verified by generating the Python bindings, e.g. `def created(self) -> typing.List[OwnedObjectReference]` and `def gas_object(self) -> typing.Optional[OwnedObjectReference]`.

crates/iota-sdk-types/src/effects/mod.rs
DaughterOfMars: But this returns the box now, no?

crates/iota-sdk-types/src/lib.rs
DaughterOfMars: Oh, I see

crates/iota-sdk-types/src/effects/mod.rs
Thoralf-M: No, Box is unwrapped
```Rust
let _: &TransactionEffectsV1         = effects.as_v1();
let _: Option<&TransactionEffectsV1> = effects.as_opt_v1();
let _: bool                          = effects.is_v1();
let _: Option<TransactionEffectsV1>  = effects.clone().into_opt_v1();
let _: TransactionEffectsV1          = effects.into_v1();
```

DaughterOfMars: ah, nice

crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -424,3 +424,170 @@ impl crate::TreeDisplay for ObjectOut {
...
+                    .input_state
+                    .version_opt()
+                    .map(|version| (changed.object_id, version))
+            })
+            .collect()
+    }
+
+    /// The reference and owner, before this transaction, of every object it
+    /// modified.
+    pub fn old_object_metadata(&self) -> Vec<(ObjectReference, Owner)> {
```
thibault-martinez: That would probably result in a lot of changes in the monorepo but what about actually adding `OwnedObjectReference` that you already added to FFI? Would remove lots of .0 and .1 that are not super nice

Thoralf-M: Also want to have `Vec<(ObjectId, Version)>` replaced by `ObjectVersion` struct?

thibault-martinez: Probably yeah

Thoralf-M: Added

DaughterOfMars (approved):

crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -424,3 +425,170 @@ impl crate::TreeDisplay for ObjectOut {
...
+
+    /// Objects that were wrapped inside another object before this transaction
+    /// and that it promoted back to top-level objects in the store.
+    pub fn unwrapped(&self) -> Vec<OwnedObjectReference> {
+        self.changed_objects
+            .iter()
+            .filter(|changed| {
+                changed.input_state.is_missing() && changed.id_operation == IdOperation::None
+            })
+            .filter_map(|changed| self.output_reference(changed))
```
thibault-martinez, line 498: Is this correct? monorepo would have returned `None` for `PackageWrite`

crates/iota-sdk-types/src/effects/v1.rs
```diff
@@ -424,3 +425,170 @@ impl crate::TreeDisplay for ObjectOut {
...
+            |changed| changed.input_state.is_data() && changed.id_operation == IdOperation::None,
+            ObjectDigest::OBJECT_WRAPPED,
+        )
+    }
+
+    /// The post-transaction reference and owner of the gas object, or `None`
+    /// for a transaction that requires no gas (a system transaction).
+    pub fn gas_object(&self) -> Option<OwnedObjectReference> {
+        let changed = self.changed_objects.get(self.gas_object_index? as usize)?;
+        self.output_reference(changed)
```
thibault-martinez: Monorepo panics if not `ObjectWrite`?

crates/iota-sdk-types/src/effects/v1.rs
Thoralf-M: You're right — fixed in 3b21cd4c0. `unwrapped` now also requires the output to be an object write, so a `(Missing, PackageWrite, IdOperation::None)` entry is reported by neither implementation.

My reasoning for the looser match was that the shape is unreachable (a package is never wrapped, so never unwrapped), but that isn't a good enough reason: step 4 of #1328 is meant to assert that the node's trait and these accessors agree, and that test shouldn't have to except inputs where they differ. No fixture can produce the shape either, so there's a constructed test pinning it.

crates/iota-sdk-types/src/effects/v1.rs
Thoralf-M: It does — `panic!("Gas object must be an ObjectWrite in changed_objects")`. This returns `None` instead, deliberately: the accessor is already optional for a transaction that needs no gas, and these bytes come off the network, so a malformed blob shouldn't be able to panic a caller.

Until 3b21cd4c0 it was worse than either — a package gas object would have been returned as if it were a coin. Now it is `None`, with a constructed test for it.

One consequence worth flagging: `None` now means both "this transaction needs no gas" and "the gas index points at something that isn't an object write". The second is unreachable for valid effects, so I did not think a `Result` was worth the noise, but say if you'd rather have the two distinguished.

thibault-martinez (approved):

PR #1333, opened 2026-08-12
feat: add ServiceConfig::supports_feature

DaughterOfMars:
Checking whether a node has a feature enabled meant iterating `enabled_features` and matching, since `Feature` carried no equality derives. A set-typed field isn't an option — cynic requires GraphQL list positions to be `Vec`.

### DIFF
--- a/crates/iota-sdk-graphql-client/src/query_types/service_config.rs
+++ b/crates/iota-sdk-graphql-client/src/query_types/service_config.rs
@@ -74,7 +74,14 @@ pub struct ServiceConfig {
     pub request_timeout_ms: i32,
 }
 
-#[derive(Clone, Copy, cynic::Enum, Debug)]
+impl ServiceConfig {
+    /// Whether `feature` is among the features this RPC service has enabled.
+    pub fn supports(&self, feature: Feature) -> bool {
+        self.enabled_features.contains(&feature)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


crates/iota-sdk-graphql-client/src/query_types/service_config.rs
```diff
@@ -74,7 +74,14 @@ pub struct ServiceConfig {
     pub request_timeout_ms: i32,
 }
 
-#[derive(Clone, Copy, cynic::Enum, Debug)]
+impl ServiceConfig {
+    /// Whether `feature` is among the features this RPC service has enabled.
+    pub fn supports(&self, feature: Feature) -> bool {
```
thibault-martinez: I feel like the name is maybe a tiny bit too generic

thibault-martinez (approved):

Thoralf-M (approved):

crates/iota-sdk-graphql-client/src/query_types/service_config.rs
DaughterOfMars: would you prefer supports_feature?

thibault-martinez: I would

DaughterOfMars: renamed

thibault-martinez (commented):
Maybe semantics but a node could support a feature while still disabling it...

